Issue #326, opened 2026-04-16
Replace pos_or_panic! in src/strategies/ with checked constructors


joaquinbejar:
## Overview

Remove every `pos_or_panic!` invocation from production code under
`src/strategies/`. Allowed only in `#[cfg(test)]` and doc examples.

## Context

- Hotspots: `src/strategies/mod.rs` (~22), `src/strategies/custom.rs`,
  `src/strategies/collar.rs`, `src/strategies/bull_put_spread.rs`, and
  every concrete spread / butterfly / condor.

## Tasks

- [ ] Replace every production-code `pos_or_panic!` in `src/strategies/`
      with `Positive::new_decimal(...)?` and extend `StrategyError` with
      `From<PositiveError>` (or a dedicated variant)
- [ ] Keep `pos_or_panic!` only under `#[cfg(test)]` and in doc examples

## Acceptance criteria

- [ ] `grep -rn "pos_or_panic!" src/strategies/` outside `#[cfg(test)]`
      and doc comments returns **0** matches
- [ ] `cargo clippy --all-targets --all-features --workspace -- -D warnings` clean
- [ ] `cargo test --all-features --workspace` clean

## References

- `rules/global_rules.md` — §Error Handling

joaquinbejar:
Implemented and merged via PR #361.

PR #361, opened 2026-04-18
Replace pos_or_panic! in strategies with checked constructors (#326)

joaquinbejar:
Replace 24 production sites across spreads, custom, and probabilities modules:
- bull_call_spread, bull_put_spread, bear_put_spread, short_butterfly_spread, long_butterfly_spread: swap throwaway probability arguments from pos_or_panic!(max_profit/loss) to Positive::ZERO (recomputed by calculate_probability).
- probabilities/core.rs: expected_value returns Result, use Positive::new().map_err() for trend-adjusted computation.
- probabilities/utils.rs: default volatility (0.2) via Positive::new_decimal(dec!(0.2)).unwrap_or(Positive::ZERO).
- custom.rs: 12 literal multipliers (0.05, 0.5, 1.5, 0.01, 50.0) replaced with Positive::new_decimal(dec!(X.X)).unwrap_or(Positive::ZERO).
- Gate pos_or_panic! imports behind #[cfg(test)] in all modified files so test modules retain access.

All 3725 lib + 416 integration + 12 property tests green; clippy -D warnings clean; full release build clean.

### DIFF
--- a/src/strategies/bear_put_spread.rs
+++ b/src/strategies/bear_put_spread.rs
@@ -1,4 +1,6 @@
-use positive::{Positive, pos_or_panic};
+use positive::Positive;
+#[cfg(test)]
+use positive::pos_or_panic;
 /*
 Bear Put Spread Strategy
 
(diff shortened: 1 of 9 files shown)


joaquinbejar:
Thanks for the sweep, nothing else to address on this one.

Issue #339, opened 2026-04-16
Apply #[inline]/[always]/[never]/[cold] per hot-path and error-path rules


joaquinbejar:
## Overview

Apply `#[inline]`, `#[inline(always)]`, `#[inline(never)]`, and `#[cold]`
per the rules. Hot-path helpers and numeric kernels get `#[inline]`;
inner-loop kernels (Monte-Carlo step, binomial backward induction,
Black-Scholes closed form) get `#[inline(always)]`; error constructors,
logging, I/O, and large builders get `#[inline(never)]` plus `#[cold]`.

## Context

- `rules/global_rules.md` §Compiler Attributes — §`#[inline]` / §`#[cold]`.
- No premature optimization: for `#[inline(always)]`, require a Criterion
  benchmark showing the win; `#[inline]` is safe by default on small
  accessors / converters.

## Tasks

- [ ] Audit hot-path helpers and apply `#[inline]`
- [ ] Apply `#[inline(always)]` only on benched inner-loop kernels
- [ ] Apply `#[cold] #[inline(never)]` on every error constructor and
      large / rarely-taken branch
- [ ] Benchmark (Criterion) before / after for any `#[inline(always)]`
      addition

## Acceptance criteria

- [ ] Criterion run shows no regression on existing benches
- [ ] Any `#[inline(always)]` addition has a documented benchmark win
- [ ] `cargo clippy --all-targets --all-features --workspace -- -D warnings` clean
- [ ] `cargo test --all-features --workspace` clean
- [ ] `cargo build --release` clean

## References

- `rules/global_rules.md` — §Compiler Attributes, §Performance Discipline

PR #376, opened 2026-04-19
Apply #[inline]/[cold]/[never] per hot-path rules (#339)

joaquinbejar:
- **refactor(attrs): apply #[inline] / #[cold]+#[inline(never)] per rules (#339)**
- **refactor(attrs): #[inline] on AdjustmentConfig builder chain**
- **refactor(attrs): extend #[inline] sweep to chains, metrics, delta_neutral**


joaquinbejar:
@copilot review

PR #412, opened 2026-07-12
Eliminate redundant per-step work in walk generators (#404)

joaquinbejar:
## Summary

Every simulation step in the walk generators paid for work that was either duplicated or immediately discarded: a full greek recomputation after `build_chain` had already computed greeks, per-step re-derivation of build parameters that were immediately overwritten, and several full-chain deep clones per iteration. This PR removes that overhead — Criterion shows **-27% / -29%** on `generator_optionchain` — with no behavioral change for chains whose strikes price successfully.

First PR of the `chains/generators.rs` improvement series (#404 → #411). Stacked: subsequent PRs branch off this one; all target `main`.

## Changes

- `create_chain_from_step` no longer calls `update_greeks()` after `build_chain` — `build_chain` computes delta/gamma per strike already. The price-failure arm of `create_chain_data` now computes greeks too, so greek coverage is identical to the removed post-build pass (greeks do not depend on bid/ask).
- Build parameters are derived **once** from the initial chain/series before the loop and cloned per step, instead of being re-derived from each freshly rebuilt chain (`to_build_params` does several O(S) scans plus an O(S log S) sort per call). Every step's chain shape is now anchored to the input chain instead of feeding back the previous rebuild's lossy parameters.
- Removed the spurious deep clone after `Ystep::next` (it returns an owned value) and replaced the O(n) `y_steps.remove(0)` front-shift with `iter().skip(1)` in all three generators (`generator_optionchain`, `generator_positive`, `generator_optionseries`).
- `create_chain_from_step` takes `new_price: &Positive` instead of `Option<Box<Positive>>` — `None` was an impossible state (guaranteed downstream "missing underlying price" error), and the sibling `create_series_from_step` already used `&Positive`.
- The size ≤ 1 early return now also covers walks whose walker produced only the initial value, avoiding a pointless `to_build_params` call on chains that will never be rebuilt (this also keeps empty synthetic chains working, as exercised by the existing coverage tests).
- Drive-by (separate commit): fixed two pre-existing clippy lints (`question_mark` in `simulation/exit.rs`, `for_kv_map` in `tests/unit/strategies/custom_test.rs`) that fail the `-D warnings` gate on current stable clippy; both pre-date this branch.

## Public API impact

None. `create_chain_from_step` / `create_series_from_step` are private; `generator_*` signatures unchanged.

## Performance

Criterion, group "Chain Generators" (new benchmark in `benches/chains/generators.rs`), Apple Silicon (darwin), default features, 20 samples:

[...]

PR #413, opened 2026-07-12
Fix chain_size units mismatch across to_build_params, build_chain and strike_step (#405)

### LINKED ISSUE #405: Fix chain_size units mismatch across to_build_params, build_chain and strike_step
## Overview

`OptionChain::to_build_params` sets `chain_size = self.len()` (total strike rows) while `OptionChain::build_chain` consumes `chain_size` as a per-side half-width, and `strike_step` consumes it as a total count. The result: every chain rebuilt inside `generator_optionchain` changes shape on the first step and, at longer expiries, grows unboundedly per step. Empirically verified with the SP500 fixture: 45 rows → 77 (30d expiry, saturating), 45 → 91 → 149 → 183 compounding at 180d — with proportional growth in `build_chain`/greeks/clone cost per step and a silent change of output semantics.

## Context

Three code sites disagree about the units of `chain_size`:

- `src/chains/chain.rs:595` correctly computes `strikes_below.max(strikes_above)` (per-side), then `:599-603` overwrites it with `self.len()` (total rows, `Len` impl at `src/chains/chain.rs:3012-3015`) — the careful computation is dead code.
- `src/chains/chain.rs:513-545` (`build_chain`) uses `chain_size` per-side (~2n+1 rows generated); the field is documented as per-side in `src/chains/utils.rs:117-118`.
- `src/chains/utils.rs:730-756` (`strike_step`) treats `chain_size` as a TOTAL count (`raw_step = 2k*sigma/(size-1)`), and is invoked whenever `strike_interval` is `None` — which `OptionSeries::build_series` forces per expiration (`src/series/model.rs:168`).

**Warning — the naive fix breaks a passing test.** Replacing the overwrite with the computed per-side value makes `series::generators::tests_generator_optionseries::test_generator_optionseries_historical` fail with `ChainError::InvalidStrike { strike: 0.0 }`: with a small per-side count, `strike_step` yields a giant interval and `rounder` (`src/chains/utils.rs:647-664`) rounds ATM to `Positive::ZERO`. Both consumers' units must be reconciled in the same change (either make `strike_step` per-side aware, or pass it a separately-derived total).

Also verified: the growth is capped only accidentally, by the `some_price_is_none` break (`src/chains/chain.rs:539-543`) when deep strikes lose bid/ask — with tighter spreads the compounding continues. At very short expiries (2d) the chain actually shrinks (45 → 17), so the bug manifests as a shape discontinuity in both directions, not always inflation.

This bug is also a performance multiplier for walk generators: from step ~3 onward every `build_chain`, greek update, and deep clone in the generator loop pays 3-10x the cost implied by the input chain.

## Tasks

- [ ] Decide and document the canonical unit for `OptionChainBuildParams::chain_size` (per-side half-width, per the existing doc in `src/chains/utils.rs:117-118`)
- [ ] In `to_build_params` (`src/chains/chain.rs:595-603`), use the computed `strikes_below.max(strikes_above)`; keep the `== 0 → 10` fallback
- [ ] Fix `strike_step` (`src/chains/utils.rs:730-756`) to interpret the size consistently with the canonical unit (its `size-1` divisor currently assumes total count)
- [ ] Audit all other `chain_size` consumers for unit agreement (grep `chain_size` across `src/`)
- [ ] Fix `series::generators::tests_generator_optionseries::test_generator_optionseries_historical` root cause (ATM rounding to zero strike via giant interval) rather than adjusting the assertion
- [ ] Add a regression test: `to_build_params` → `build_chain` round-trip is stable in strike count (no compounding across ≥3 round-trips, at 30d and 180d expiries)

## Acceptance criteria

- [ ] Round-tripping a chain through `to_build_params` + `build_chain` N times keeps the strike count stable after the first rebuild (no per-step compounding)
- [ ] The dead `max(strikes_below, strikes_above)` computation is either used or removed
- [ ] `chain_size` unit documented on the field and on both consumers

## References

- `src/chains/chain.rs:570-681` (to_build_params), `:513-545` (build_chain strike loop)
- `src/chains/utils.rs:117-118` (chain_size doc), `:730-756` (strike_step), `:647-664` (rounder)
- `src/series/model.rs:168` (build_series forces strike_interval = None)
- `src/chains/generators.rs:45,54` (per-step feedback loop that amplifies the bug)

joaquinbejar:
## Summary

Three code sites disagreed about the units of `chain_size`: the field is documented and consumed by `build_chain` as a **per-side half-width**, but `to_build_params` fed back `self.len()` (total rows) and `build_chain` passed the per-side count to `strike_step`, which expects a **total** strike count. The feedback loop roughly doubled the strike grid on every `to_build_params` → `build_chain` round-trip (empirically 45 → 91 → 149 → 183 rows at 180d expiry), silently changing chain shape mid-simulation and multiplying per-step build/greeks/clone cost 3-10x from step ~3 onward.

**Stacked PR** (2/8 of the generators series): based on `issue-404-generator-perf` (#412). Targets `main`; contains #412's commits until that merges.

## Changes

- `to_build_params` keeps the computed `max(strikes_below, strikes_above)` per-side value — the `chain_size = self.len()` overwrite (which made the careful computation dead code) is gone. The `== 0 → 10` fallback is preserved.
- `build_chain` now converts the per-side half-width to the `2n+1` total that `strike_step` expects when inferring a strike interval. Inferred intervals stop being ~2x too wide — this is also what keeps the corrected (smaller) per-side counts from producing a giant interval that rounded ATM down to a zero strike in the series historical path (the failure mode that a naive one-site fix triggers in `test_generator_optionseries_historical`).
- `strike_step` documents its TOTAL-count contract explicitly.
- Regression test `test_to_build_params_round_trip_strike_count_stable` pins round-trip stability of the strike count at 30d and 180d expiries (the first rebuild may symmetrize an asymmetric input, but must not compound).

## Public API impact

None. `chain_size` and `strike_step` semantics now match their existing docs; no signatures changed. Note: chains rebuilt via `to_build_params` will keep a stable strike count across round-trips instead of growing — this is the documented intent, but downstream code relying on the accidental growth would observe smaller chains.

## Testing

- [x] Unit tests added (round-trip strike-count stability at 30d/180d)
- [x] `series::generators::tests_generator_optionseries::test_generator_optionseries_historical` still passes (the known trap for partial fixes)
- [x] `cargo test --all-features --workspace` clean (3846 lib + integration; same 6 pre-existing environmental `plotly` WebDriver failures as `main`)

## Checklist

Closes #405

### DIFF
--- a/src/chains/chain.rs
+++ b/src/chains/chain.rs
@@ -400,11 +400,14 @@ impl OptionChain {
                     &format!("failed to get days: {e}"),
                 )
             })?;
+            // `chain_size` is a per-side half-width while `strike_step`
+            // expects the TOTAL number of strikes covering ±kσ, so convert
+            // to the 2n+1 grid the builder below actually generates.
             strike_step(
                 underlying_price,
                 params.implied_volatility,
                 days,
-                params.chain_size,
+                params.chain_size * 2 + 1,
                 None,
             )
         };
@@ -600,11 +603,14 @@ impl OptionChain {
             chain_size = strikes_below.max(strikes_above);
         }
 
-        // Default to a reasonable chain size if calculation fails
+        // Default to a reasonable chain size if calculation fails.
+        // `chain_size` is the per-side half-width consumed by `build_chain`
+        // (which generates up to `chain_size` strikes above AND below ATM),
+        // so keep the computed max(strikes_below, strikes_above) instead of
+        // the total row count — feeding the total back would roughly double
+        // the strike grid on every to_build_params -> build_chain round-trip.
         if chain_size == 0 {
             chain_size = 10;
-        } else {
-            chain_size = self.len();
         }
 
         // Estimate the average bid-ask spread from the available options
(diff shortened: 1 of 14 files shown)


src/chains/chain.rs
```diff
@@ -400,11 +400,14 @@ impl OptionChain {
...
             })?;
+            // `chain_size` is a per-side half-width while `strike_step`
+            // expects the TOTAL number of strikes covering ±kσ, so convert
+            // to the 2n+1 grid the builder below actually generates.
             strike_step(
                 underlying_price,
                 params.implied_volatility,
                 days,
-                params.chain_size,
+                params.chain_size * 2 + 1,
```
joaquinbejar: 🟠 **Warning**: This public `build_chain` path multiplies and increments a caller supplied `usize` without checks. A very large `chain_size` will panic in debug builds or wrap in release and can produce an invalid strike grid. Use `checked_mul(2).and_then(|n| n.checked_add(1))` and map overflow to `ChainError`.

joaquinbejar (commented):
The unit reconciliation and regression coverage are solid, but the new size conversion can overflow before strike_step sees it. REQUEST_CHANGES.

src/chains/chain.rs
joaquinbejar: Good catch, the 2n+1 conversion now goes through checked_mul and checked_add and maps overflow to ChainError::invalid_parameters, with a usize::MAX regression test pinning it.

joaquinbejar:
Thanks for the review, the overflow point was fair. The 2n+1 conversion is checked now and the regression test covers the pathological size.

PR #414, opened 2026-07-12
Unify edge-case contracts of the three walk generators (#406)

### LINKED ISSUE #406: Unify edge-case contracts of the three walk generators (panic on empty walker, Historical divergence, series never age)
## Overview

The three walk generators (`generator_optionchain`, `generator_positive` in `src/chains/generators.rs`; `generator_optionseries` in `src/series/generators.rs`) are supposed to implement the same contract but verifiably diverge on edge cases — including a reachable panic and one generator whose series never age. Each past fix landed in a subset of the three copies (the empty-guard comments at `chains/generators.rs:139-142` and `series/generators.rs:165-168` show the same patch applied twice, missing the third site).

## Context

Verified divergences:

1. **Panic**: `generator_positive` calls `y_steps.remove(0)` unconditionally (`src/chains/generators.rs:216`) — panics on an empty walker result. The other two generators guard with `if y_steps.is_empty() { return Ok(vec![init_step]) }` (`chains:138-143`, `series:164-169`). `WalkTypeAble` is public and its default methods are overridable; nothing documents a non-empty requirement, so a user walker returning `Ok(vec![])` legally reaches the panic.
2. **Historical with `prices.len() < size`**: `generator_optionchain` silently returns `Ok(vec![init_step])` (`chains:121-122` + guard), while `generator_positive` propagates `Err(SimulationError::InsufficientHistoricalData)` from the walker (`chains:213` → `src/simulation/traits.rs:773-780`). Same input, opposite contracts, same file. The silent path is also indistinguishable from a legitimate size-1 walk, so a backtest fed short historical data quietly simulates nothing — `RandomWalk::new` does no length validation (`src/simulation/randomwalk.rs:82-90`).
3. **Series never age**: `create_series_from_step` takes no expiration argument (`src/series/generators.rs:38-51`) and the series loop never passes the decayed `x_step.datetime()` — unlike chains (`chains:156-161`). Option series therefore keep their original expirations across the whole walk.
4. **Size policy**: chains uses `debug_assert!(steps.len() <= size)` (no-op in release, `chains:172-177`, `:234-239`); series truncates at runtime (`series:196-203`).
5. **Volatility default**: series substitutes `0.20` when `None` (`series:176-177`); chains forwards the `Option`.
6. **Silent error swallowing**: `Err(_) => break` on `Xstep::next` (`chains:151-154`, `:223-226`, `series:180-183`) swallows every `SimulationError` — not just `ExpirationReached` but also step-size conversion errors — silently truncating the walk.
7. **Doc drift**: `generator_optionchain` docs claim it "uses a fixed volatility of 0.20" (`chains:62-63`, false) and reference `ChainError::DynError` (`chains:77`), a variant that does not exist — the actual conversion is `From<SimulationError> for ChainError` producing `ChainError::Simulation` (`src/error/chains.rs:693-698`).

This issue fixes the behavioral divergences and documents one explicit contract. The structural deduplication (single generic driver) is a separate follow-up issue; doing the contract work first makes that refactor purely mechanical. Note: unifying items 2/4 is a semver-observable behavior change to `pub` functions — changelog entry required.

## Tasks

- [ ] Add the empty-`y_steps` guard to `generator_positive` before `remove(0)` (`src/chains/generators.rs:216`)
- [ ] Decide ONE Historical-insufficient-data policy (recommended: propagate the typed error — `SimulationError::InsufficientHistoricalData` already flows through `From<SimulationError> for ChainError`; never a silent `Ok`) and apply it to all three generators; update the tests that pin the silent behavior (`chains:728-766`, `series:350-425`)
- [ ] Pass the decayed expiration into `create_series_from_step` so series age along the walk like chains do
- [ ] Decide ONE size-enforcement policy (recommended: runtime truncate) and apply to all three
- [ ] Remove the dead `0.20` volatility fallback in series (dead once the guard/early-return contract is unified)
- [ ] Replace `Err(_) => break` with an explicit match: break on `SimulationError::ExpirationReached`, propagate everything else
- [ ] Fix doc comments: remove the "fixed volatility of 0.20" claim, replace `ChainError::DynError` references with the real error flow, and document the unified contract (init-step invariant, Historical policy, size policy) on all three `pub fn`s
- [ ] Document the non-empty-output expectation (or lack thereof) on `WalkTypeAble`

## Acceptance criteria

- [ ] A walker returning `Ok(vec![])` produces the same, documented, non-panicking result from all three generators
- [ ] `WalkType::Historical` with insufficient prices produces the same result (same variant) from all three generators, covered by tests
- [ ] `generator_optionseries` steps have strictly decreasing time-to-expiry, covered by a test
- [ ] No `debug_assert`-only enforcement of a documented contract remains
- [ ] Changelog notes the contract changes on `pub` generators

## References

- `src/chains/generators.rs:118-143` (Historical pre-check + guard), `:216` (unguarded remove), `:151-154`/`:223-226` (Err break), `:62-63`/`:77` (doc drift)
- `src/series/generators.rs:38-51` (no expiration param), `:147-169`, `:176-177`, `:196-203`
- `src/simulation/traits.rs:766-786` (historical walker), `src/error/chains.rs:693-698` (From<SimulationError>)
- `src/simulation/randomwalk.rs:82-90` (no length validation downstream)

joaquinbejar:
## Summary

The three walk generators (`generator_optionchain`, `generator_positive`, `generator_optionseries`) are supposed to implement the same contract but verifiably diverged on every edge case — including a reachable panic (`generator_positive` on empty walker output), opposite results for the same `Historical` input (silent `Ok` vs typed `Err` in the same file), and option series that never aged along the walk. This PR makes the behavior identical, documents the contract on all three `pub fn`s, and records the semver-observable changes in `CHANGELOG.md`.

**Stacked PR** (3/8 of the generators series): based on `issue-405-chain-size-units` (#413). Targets `main`; contains #412/#413 commits until those merge.

## Changes

- **Historical policy unified (typed error)**: `WalkType::Historical` with fewer prices than `size` now returns `ChainError::Simulation(InsufficientHistoricalData)` from all three generators. The silent init-only fallback was indistinguishable from a legitimate size-1 walk, so a backtest fed short data quietly simulated nothing (`RandomWalk::new` does no length validation). The walker validates history before the vol estimate runs, so empty price vectors surface the same typed error.
- **Empty-walker guard in `generator_positive`**: a user-supplied `WalkTypeAble` returning `Ok(vec![])` (legal — the trait never required non-empty output) no longer panics; it returns the initial step only, like the other two generators. The trait docs now state the output contract.
- **Series age along the walk**: `create_series_from_step` now receives the series expirations reduced by the elapsed walk time, with expired entries dropped; the walk ends early once every expiration has passed. Previously rebuilt series kept their original expirations for the entire walk — time never passed for them.
- **Runtime size enforcement everywhere**: walks longer than `WalkParams::size` are truncated at runtime in all three generators (chains generators previously used `debug_assert!`, a no-op in release).
- **Errors are no longer swallowed**: the x-step advance now breaks only on `SimulationError::ExpirationReached` (the normal end of a walk) and propagates any other failure. Previously `Err(_) => break` silently truncated the walk on every error kind.
- **Dead code / stale docs**: removed the undocumented `0.20` volatility fallback in series (dead under the unified contract), the "uses a fixed volatility of 0.20" claim, and references to the nonexistent `ChainError::DynError` (the real conversion is `From<SimulationError> for ChainError` → `ChainError::Simulation`).

## Public API impact

No signature changes, but **behavior of three `pub fn`s changed** (Historical error, runtime truncation, error propagation, series aging). All changes are listed under `[Unreleased]` in `CHANGELOG.md`. Tests that pinned the old silent behavior were updated to pin the new contract.

## Testing

- [x] Unit tests updated: `test_generator_optionseries_historical_empty_prices` / `_insufficient_prices` now assert the typed `InsufficientHistoricalData` error
- [x] `cargo test --all-features --workspace` clean (3850 lib tests; same 6 pre-existing environmental `plotly` WebDriver failures as `main`)

## Checklist

- [x] Checked arithmetic only
- [x] CHANGELOG updated with the contract changes

Closes #406

### DIFF
--- a/src/series/generators.rs
+++ b/src/series/generators.rs
@@ -1,4 +1,4 @@
-use crate::error::ChainError;
+use crate::error::{ChainError, SimulationError};
 use crate::series::{OptionSeries, OptionSeriesBuildParams};
 use crate::simulation::steps::{Step, Ystep};
 use crate::simulation::{WalkParams, WalkType};
@@ -8,30 +8,30 @@ use crate::volatility::{adjust_volatility, constant_volatility};
 use core::option::Option;
 use positive::Positive;
 use rust_decimal::Decimal;
-use rust_decimal_macros::dec;
 use tracing::debug;
 
 #[cfg(test)]
 use positive::pos_or_panic;
 
-/// Creates a new `OptionSeries` from a given previous `Ystep` series, a new price,
-/// and an optional volatility value.
+/// Creates a new `OptionSeries` from pre-derived build parameters, a new price,
+/// an optional volatility, and the aged series expirations.
 ///
 /// # Parameters
-/// - `previous_y_step`: A reference to the previous `Ystep` series of type `OptionSeries`.
-///   It represents the state of the series prior to this computation.
-/// - `new_price`: A reference to a `Positive` value representing the new price to use as an input
-///   for the computation.
-/// - `volatility`: An optional `Positive` value representing the volatility used in the calculation.
-///   If `None`, a default behavior or calculation is assumed.
+/// - `build_params`: Build parameters derived once from the initial `OptionSeries`.
+/// - `new_price`: A reference to a `Positive` value representing the new underlying price.
+/// - `volatility`: An optional `Positive` value representing the implied volatility to
+///   stamp on the rebuilt chains. If `None`, the volatility carried by `build_params`
+///   is kept.
+/// - `aged_series`: The series' days-to-expiration after subtracting the walk time
+///   elapsed since the initial step (already-expired entries removed), so the option
+///   series ages along the walk.
 ///
 /// # Returns
-/// Returns a `Result` that, on success, contains the newly created `OptionSeries`. On failure,
-/// it contains a boxed dynamic error with details about what went wrong.
+/// Returns a `Result` that, on success, contains the newly created `OptionSeries`.
 ///
 /// # Errors
 /// This function can return an error if:
-/// - The `previous_y_step` contains invalid or inconsistent data for creating the new series.
+/// - The `build_params` contain invalid or inconsistent data for creating the new series.
 /// - The `new_price` or optional `volatility`, if provided, result in an invalid computation.
 /// - Any other unexpected error occurs during the processing.
 ///
@@ -39,12 +39,14 @@ fn create_series_from_step(
     build_params: &OptionSeriesBuildParams,
     new_price: &Positive,
     volatility: Option<Positive>,
+    aged_series: Vec<Positive>,
 ) -> Result<OptionSeries, ChainError> {
     let mut series_params = build_params.clone();
     series_params.set_underlying_price(new_price);
     if let Some(volatility) = volatility {
         series_params.set_implied_volatility(volatility);
     }
+    series_params.series = aged_series;
     let new_chain = OptionSeries::build_series(&series_params)?;
     Ok(new_chain)
 }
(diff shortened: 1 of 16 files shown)


src/series/generators.rs
```diff
@@ -144,50 +158,74 @@ pub fn generator_optionseries(
...
+        // Age the series' expirations by the walk time elapsed since the
+        // initial step, dropping the ones that have already expired, so the
+        // option series ages along the walk like the chain generator does.
+        let elapsed_days =
+            (initial_days_left.to_dec() - previous_x_step.days_left()?.to_dec()).max(Decimal::ZERO);
+        let aged_series: Vec<Positive> = build_params
+            .series
+            .iter()
+            .filter(|days| days.to_dec() > elapsed_days)
+            .map(|days| Positive::new_decimal(days.to_dec() - elapsed_days))
```
joaquinbejar: 🟠 **Warning**: This expiration calculation uses raw `Decimal` subtraction on financial time values, despite `global_rules.md` requiring checked arithmetic. Use `checked_sub(elapsed_days)` and map failure to `ChainError`, and do the same for the `elapsed_days` calculation above.

joaquinbejar (commented):
The edge-case behavior is much more consistent, but the new series-aging path violates the required checked-arithmetic contract. REQUEST_CHANGES.

src/series/generators.rs
joaquinbejar: Fixed, both the elapsed_days computation and the per-entry aging subtraction now use Decimal::checked_sub with a typed ChainError on overflow.

joaquinbejar:
Thanks, the aging arithmetic is fully checked now per the review.

PR #418, opened 2026-07-12
Add multi-step behavioral tests for the walk generators (#410)

joaquinbejar:
## Summary

The generator "coverage" tests all used `size = 1` — which returns before the rebuild loop ever runs — and asserted nothing but `steps.len()`. Expiration decay, price propagation and IV propagation were executed but never checked; exactly the behaviors the preceding refactors (#406–#409) touch. This PR adds a deterministic test walker and multi-step behavioral tests that pin the unified generator contract with exact assertions, no seeds and no wall-clock dependence.

**Stacked PR** (7/8 of the generators series): based on `issue-409-smile-fit` (#417). Targets `main`; contains #412–#417 commits until those merge.

## Changes

- **`simulation::walk_test_support`** (test-only, `#[cfg(test)]`): `RampWalker` produces the deterministic ramp `start, start+δ, start+2δ, …` for every synthetic walk type (Historical keeps the default trait behavior so its data-validation semantics stay testable); `EmptyWalker` pins the empty-output contract.
- **`generator_positive`**: exact multi-step assertions — ramp prices, `Ystep` index increments, time-to-expiry decreasing one day per step; truncation exactly at expiration (3-day expiry + daily steps → init + 3 steps, last at 0 DTE); `size = 0` returns the init-only walk.
- **`generator_optionchain`**: rebuilt chains track the walked price exactly; chain expiration follows the x-step (1-day parse tolerance); ATM IV tracks the constant walk volatility; **Historical multi-step** replays the provided prices and the final step's ATM IV equals the expanding-window estimate computed (in the test, by hand, via `constant_volatility` + `adjust_volatility`) over exactly the walked slice — the price beyond the walk is present in the input and must not affect the estimate; empty-walker contract test.
- **`generator_optionseries`**: series expirations age by exactly the elapsed days at every step (30/60/90 → 29/59/89 → …), and the walk stops once every expiration in the series has passed (single 2-day series entry → init + 1 step).
- **Robustness fix surfaced by the new determinism**: simulated vol paths can touch the zero boundary (Heston CIR truncation), and `build_chain` rejects IV = 0 — the per-step IV clamp at the generator boundary is now `(0.0001, 1.0]`, removing a real flaky failure (the #408 Heston test failed intermittently before this).

## Public API impact

None (test-support module is `#[cfg(test)] pub(crate)`). The IV floor is an internal robustness fix on the same clamp introduced in #408.

## Testing

[...]

Issue #410, opened 2026-07-12
Add multi-step behavioral tests for walk generators


joaquinbejar:
## Overview

The walk generators have coverage-only tests that never exercise the code they nominally cover: all 10 tests in `generators_coverage_tests` (`src/chains/generators.rs:362-767`) use `size = 1`, which triggers the early return before the chain-rebuild loop ever runs, and assert nothing but `steps.len() == 1`. No test anywhere asserts multi-step correctness — expiration decrement, IV propagation into rebuilt chains, or price propagation are executed but never checked. The behaviors most likely to regress during the planned refactors are exactly the unpinned ones.

## Context

Verified gaps:

- All `generators_coverage_tests` set `size: 1` (`src/chains/generators.rs:400,431,468,505,543,581,621,660,701,741`) and assert only `steps.len() == 1`. The dispatch line is covered; the loop body is not.
- `test_create_chain_from_step_with_volatility_change` runs a real 4-step GBM walk but asserts only `random_walk.len() == n_steps` (`:322`); `test_generator_positive` likewise (`:358`).
- The Historical vol-computation path (`:124-134`) executes in `test_generator_optionchain_historical` (3 prices, size 1) but its computed IV is never consumed or observed — `remove(0)` empties `y_steps` and the loop never runs.
- The `Err(_) => break` truncation path (expiration reached mid-walk, `:151-154`, `:223-226`) has zero coverage in either generator.
- `generator_positive`'s panic on empty walker output and the `size = 0` contract are untested (size 0 makes the default `historical` return an empty vec for ANY prices — empirically confirmed to panic today).
- Fixture/time fragility: tests load `examples/Chains/SP500-18-oct-2024-5781.88.json` via relative path (`:260`, `:296`) and depend on wall-clock via `get_x_days_formatted`/`get_tomorrow_formatted` (`:261`, `:392`, `:460`, `:497`; `Local::now()` in `src/utils/time.rs:257-260`, `:279-282`). The fixture has a stale 2024 expiry and percent-scaled IVs, forcing tests to patch it before use.
- No shared contract test pins the (currently divergent) semantics across `generator_optionchain` / `generator_positive` / `generator_optionseries`.

These tests should pin the contract decided in the contracts issue and protect the unification + vol-dynamics refactors. A deterministic `WalkTypeAble` test walker (returning a fixed price ramp) makes every assertion exact — no seeds, no wall-clock.

## Tasks

[...]

Issue #407, opened 2026-07-12
Consolidate WalkType dispatch and walk driver into a single generic implementation


joaquinbejar:
## Overview

Roughly 200 of the ~240 non-test lines across `src/chains/generators.rs` and `src/series/generators.rs` are one algorithm written three times: a 53-line, 10-arm `match` on `WalkType` whose only job is to pick the trait method named after the variant and copy out the `volatility` field, plus a copy-pasted driver loop (x-advance, y-build, push). Adding a `WalkType` variant today requires touching five dispersed sites; the copies have already diverged (see the contracts issue). Collapse dispatch and driver into single implementations so each generator becomes <10 lines and a new variant requires edits in exactly two places (enum + trait).

## Context

- The 10-arm dispatch appears at `src/chains/generators.rs:85-137` (generator_optionchain), `:203-214` (generator_positive), and near-byte-identically at `src/series/generators.rs:111-163`. The Historical vol-estimation block (`calculate_log_returns` → `constant_volatility` → `adjust_volatility`) is likewise duplicated (`chains:124-134`, `series:150-160`).
- `WalkType` (`src/simulation/model.rs`) currently has **no methods** — only a `Display` impl. Every `WalkTypeAble` method already re-matches `params.walk_type` and errors `InvalidWalkType` on mismatch, so the outer match is pure double dispatch.
- The driver loop appears at `chains:145-178`, `chains:216-241`, `series:171-204` with identical structure.
- Feasibility verified: `WalkParams` (`src/simulation/params.rs:37-41`) and `Ystep` (`src/simulation/steps/y.rs:35-37`) bounds are exactly `TryInto<Positive> + Display + Clone`; all `WalkTypeAble` methods return `Result<Vec<Positive>, SimulationError>` independent of `Y`; `From<SimulationError> for ChainError` exists (`src/error/chains.rs:693`); `RandomWalk::new` is generic over `FnOnce(&WalkParams<X,Y>) -> Result<Vec<Step<X,Y>>, E>` (`src/simulation/randomwalk.rs:82-90`) so consumers need no changes; `simulation/` already imports `crate::volatility`, so absorbing the Historical vol block breaks no module boundary and stays pure-numeric.

Proposed shape:

1. `impl WalkType { pub fn volatility(&self) -> Option<Positive> }` — `Some(*volatility)` for the nine synthetic variants; for `Historical` either `None` or absorb the duplicated log-returns/constant-vol/annualize block here.
[...]

PR #419, opened 2026-07-12
Parallelize per-step chain builds in the walk generators with rayon (#411)

### LINKED ISSUE #411: Parallelize per-step chain builds in generator_optionchain with rayon
## Overview

Once build params are hoisted out of the generator loop (prerequisite issue), each simulation step's chain depends only on `(price_i, expiration_i, vol_i)` plus the init-chain params — the serial `previous_y_step` round-trip disappears and all per-step `build_chain` calls become independent. Building them with rayon `par_iter` unlocks ~num_cores speedup on the dominant per-step cost for long walks (the shipped example runs 10,080 steps).

## Context

Feasibility verified:

- The walker already produces the full price vector up front (`Vec<Positive>`, ~16 B/step — negligible memory), then chains are built strictly serially (`src/chains/generators.rs:150-170`). The only loop-carried state is `previous_y_step`, whose params are immediately overwritten each step — i.e. the dependency is artificial once params are hoisted.
- The expiration sequence is deterministic and precomputable: `Xstep::next` (`src/simulation/steps/x.rs:206-233`) has no randomness — collect the x-steps first (respecting the expiration-break contract), then build chains in parallel over `(i, price_i, expiration_i)`.
- `OptionChainBuildParams` derives `Clone` (`src/chains/utils.rs:109`); `build_chain` is deterministic (no RNG on the build path); rayon is already a non-optional dependency used in `volatility/`, `curves/`, `surfaces/`, `utils/` — no module-boundary or feature-flag issue.
- `Ystep::new(index, value)` is `pub` and `next()` only increments the index, so indices can be assigned from the enumeration index after the parallel build.
- Streaming/iterator generation is NOT a useful alternative: `RandomWalk::new` materializes `Vec<Step>` anyway (`src/simulation/randomwalk.rs:51`, `:88`), so peak memory would not drop.

Depends on: the per-step-overhead issue (params hoist) and, if landed, should be implemented inside the unified driver from the generator-unification issue rather than in each generator. Error handling: with per-step vol paths (vol-dynamics issue), zip `vols` into the parallel iterator — the design composes.

Per repo rules, performance work requires Criterion numbers before/after.

## Tasks

- [ ] Precompute the x-step/expiration sequence serially (respecting the documented expiration-truncation contract), producing the definitive step count
- [ ] Build chains with `rayon::par_iter` over `(index, price, expiration[, vol])`, cloning the hoisted `OptionChainBuildParams` per task
- [ ] Collect into `Vec<Step>` preserving order (`collect` on indexed par_iter) and assign `Ystep` indices from enumeration
- [ ] Propagate the first `ChainError` from the parallel section (e.g. `try_fold`/`collect::<Result<...>>`)
- [ ] Criterion benchmark: `generator_optionchain` at 1k and 10k steps, serial vs parallel, on a realistic chain; record numbers in the PR
- [ ] Verify determinism: same inputs produce identical output to the serial path (order and values)

## Acceptance criteria

- [ ] Output is byte-identical to the serial implementation for identical inputs (test-pinned)
- [ ] Criterion shows near-linear scaling on the build-dominated portion for 10k-step walks (numbers in PR)
- [ ] No behavior change to truncation/expiration semantics

## References

- `src/chains/generators.rs:150-170` (serial loop), `src/simulation/steps/x.rs:206-233` (deterministic x-advance)
- `src/simulation/randomwalk.rs:51,88` (Vec materialization)
- `examples/examples_simulation/src/bin/random_walk_build_chain.rs:18` (10,080-step walk)

## Dependencies

- Depends on #404 (params hoist makes steps independent)
- Should be implemented inside the unified driver from #407 if that has landed

joaquinbejar:
## Summary

With build params hoisted (#404/#412) and the driver unified (#407/#415), each simulation step's chain depends only on `(price_i, vol_i, expiration_i)` plus the initial context — the serial rebuild loop was an artifact, not a real dependency. This PR adds a parallel driver variant and runs the chain/series generators on it: **−82% / −87%** on the chain-generator benchmarks (2.59 ms → 468 µs at 10 steps, 7.01 ms → 881 µs at 25 steps), with output identical to the serial driver.

**Stacked PR** (8/8, closes the generators series): based on `issue-410-generator-tests` (#418). Targets `main`; contains #412–#418 commits until those merge.

## Changes

- **`simulation::walk_steps_par`**: parallel variant of `walk_steps` with an identical contract. The walker output, per-step volatilities and the x-step sequence are computed serially — deterministic, and honoring the expiration-truncation contract — then the per-step `next_y` calls (the dominant cost when they rebuild an option chain, or one chain per expiration for series) fan out on the rayon pool with order preserved on collection. Documented semantic notes: `next_y` must be `Fn + Sync`; `Ok(None)` still ends the walk at the previous step (steps past it are computed speculatively and discarded); on failure the first error in step order is returned.
- **`generator_optionchain` / `generator_optionseries`** now run on the parallel driver. The lazily-derived build-params context moved into a `Mutex<Option<…>>` (locked briefly once per step; the params were already cloned per step by design), and the closures capture only `Sync` data — the boxed walker inside `WalkParams` is not `Sync`, so they capture `ystep_ref()` / the `Copy` init x-step instead.
- **`generator_positive`** stays on the serial driver (its per-step work is a copy; nothing to gain).
- rayon was already a non-optional dependency (used in `volatility/`, `curves/`, `surfaces/`, `utils/`) — no manifest change.

## Performance

Criterion, group "Chain Generators", Apple Silicon (darwin), default features, 20 samples, baseline saved on this branch immediately before the change (i.e. isolating the parallelization from the rest of the stack):

| Benchmark | Serial | Parallel | Change |
|---|---|---|---|
| `generator_optionchain` 10 steps | 2.594 ms | 468 µs | **−81.9%** |
| `generator_optionchain` 25 steps | 7.005 ms | 881 µs | **−87.3%** |
| `generator_positive` 1000 steps | — | — | unchanged (serial driver) |

Cumulative across the series (#404 + #405 + #411): 25-step chain walk 12.46 ms → 0.88 ms (~14x).

## Public API impact

New public item: `simulation::walk_steps_par` (prelude-exported). `generator_optionchain` / `generator_optionseries` signatures and outputs unchanged (determinism pinned by test).

## Testing

- [x] Serial-vs-parallel equality test (`test_walk_steps_par_matches_serial`): identical values, indices and dates for identical inputs
- [x] All multi-step behavioral tests from #410 pass unchanged on the parallel driver (price/IV/expiration propagation, truncation, aging, early-stop)
- [x] Criterion numbers above (baseline on same branch)
- [x] `cargo test --all-features --workspace` clean (same 6 pre-existing environmental `plotly` WebDriver failures as `main`)

## Checklist

- [x] No `.unwrap()` / `.expect()` / unchecked indexing in production code (parallel section uses `.get()` + `ok_or_else`)
- [x] CHANGELOG updated (Performance section)

Closes #411

### DIFF
--- a/src/simulation/walk_driver.rs
+++ b/src/simulation/walk_driver.rs
@@ -36,8 +36,11 @@ use tracing::debug;
 /// # Errors
 ///
 /// Propagates errors from the historical helpers (`calculate_log_returns`,
-/// `constant_volatility`, `adjust_volatility`).
-fn walk_volatility<Y>(walk_params: &WalkParams<Positive, Y>) -> Result<Option<Positive>, ChainError>
+/// `constant_volatility`, `adjust_volatility`) as [`SimulationError`], so
+/// the driver stays free of chain-layer error types.
+fn walk_volatility<Y>(
+    walk_params: &WalkParams<Positive, Y>,
+) -> Result<Option<Positive>, SimulationError>
 where
     Y: TryInto<Positive> + Display + Clone,
 {
@@ -86,33 +89,36 @@ where
 /// # Returns
 ///
 /// * `Ok(Vec<Step<Positive, Y>>)` - The simulated walk.
-/// * `Err(ChainError)` - If the walker, the volatility estimation, the
-///   x-step advance, or `next_y` fail.
+/// * `Err(E)` - If the walker, the volatility estimation, the x-step
+///   advance, or `next_y` fail.
 ///
 /// # Errors
 ///
-/// Returns [`ChainError::Simulation`] (via the `From<SimulationError>`
-/// conversion) if the walker returns an error — including
-/// `SimulationError::InsufficientHistoricalData` when a `Historical` walk has
-/// fewer prices than `walk_params.size` — and propagates errors from the
-/// volatility estimation and from `next_y`.
-pub fn walk_steps<Y, F>(
+/// The driver is generic over its error type: every internal failure is a
+/// [`SimulationError`] lifted into `E` via `From`, so the simulation layer
+/// carries no chain-specific error types — the chain and series generators
+/// simply instantiate `E = ChainError`. Walker errors include
+/// `SimulationError::InsufficientHistoricalData` when a `Historical` walk
+/// has fewer prices than `walk_params.size`; `next_y` errors are returned
+/// verbatim.
+pub fn walk_steps<Y, E, F>(
     walk_params: &WalkParams<Positive, Y>,
     mut next_y: F,
-) -> Result<Vec<Step<Positive, Y>>, ChainError>
+) -> Result<Vec<Step<Positive, Y>>, E>
 where
     Y: TryInto<Positive> + Display + Clone,
-    F: FnMut(&Positive, Option<Positive>, &Xstep<Positive>) -> Result<Option<Y>, ChainError>,
+    E: From<SimulationError>,
+    F: FnMut(&Positive, Option<Positive>, &Xstep<Positive>) -> Result<Option<Y>, E>,
 {
     debug!("{}", walk_params);
-    let y_steps = walk_params.walker.generate(walk_params)?;
+    let y_steps = walk_params.walker.generate(walk_params).map_err(E::from)?;
     if y_steps.len() <= 1 {
         // Preserve the init-step invariant when the walker produces no
         // values beyond the initial one; downstream consumers expect at
         // least the initial step to be present.
         return Ok(vec![walk_params.init_step.clone()]);
     }
-    let volatility = walk_volatility(walk_params)?;
+    let volatility = walk_volatility(walk_params).map_err(E::from)?;
 
     let mut steps: Vec<Step<Positive, Y>> = vec![walk_params.init_step.clone()];
     let mut previous_x_step = walk_params.init_step.x;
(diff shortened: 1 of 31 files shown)


src/simulation/walk_driver.rs
```diff
@@ -0,0 +1,847 @@
...
+            let y_step = y_steps.get(price_index).copied().ok_or_else(|| {
+                ChainError::invalid_parameters("prices", "walker path shorter than x-steps")
+            })?;
+            let volatility = step_vols
+                .as_ref()
+                .and_then(|vols| vols.get(price_index).copied())
+                .or(constant_vol);
+            next_y(&y_step, volatility, x_step)
+        })
+        .collect::<Result<Vec<Option<Y>>, ChainError>>()?;
```
joaquinbejar: 🟠 **Warning**: Collecting as `Result` propagates an error from any speculative step before the assembly loop can honor an earlier `Ok(None)`. A serial closure that returns `None` at step 2 and `Err` at step 3 succeeds and truncates, while this parallel driver returns `Err`, contradicting its identical-contract claim. Collect ordered `Vec<Result<...>>`, then stop on `None` and only propagate errors encountered before that stop.

joaquinbejar (commented):
The parallel build path preserves ordering for successful steps, but speculative error handling is not equivalent to the serial contract. REQUEST_CHANGES.

src/simulation/walk_driver.rs
joaquinbejar: Good catch, that did contradict the identical-contract claim. The parallel driver now collects Vec<Result> without short-circuiting and assembles in step order, so an error only propagates when it precedes the stop signal, with a regression test covering both orderings.

joaquinbejar:
Thanks, the parallel driver now matches the serial stop/error ordering exactly and there is a regression test for it.

PR #425, opened 2026-08-17
feat(simulation): expose expanding_window_vols (#423); close out the audit triage (#422)

### LINKED ISSUE #422: cargo audit red on main since 2026-08-05: crossbeam-epoch, quinn-proto, rkyv, rustls-webpki + number_prefix (unmaintained)
The daily audit workflow is red on main since 2026-08-05 (last green 2026-08-04), independent of any PR. Local `cargo audit` today lists:

- RUSTSEC-2026-0204 — `crossbeam-epoch` (invalid pointer dereference in fmt::Pointer)
- RUSTSEC-2026-0185 — `quinn-proto` (remote memory exhaustion)
- RUSTSEC-2026-0235 — `rkyv` (OOB reads; note: in Layer-V repos this one is lockfile-only via a rust_decimal optional feature — same likely here, verify with `cargo tree -i rkyv --all-features`)
- RUSTSEC-2026-0104 — `rustls-webpki` (reachable panic in CRL parsing)
- RUSTSEC-2025-0119 — `number_prefix` (unmaintained warning)

Each needs the usual triage: `cargo update` where a fix is reachable, documented waiver in audit config where it is not. Noted while landing #421 (statrs 0.19 bump), which is unrelated to all of the above.

### LINKED ISSUE #423: Expose the causal volatility estimator used by walk_steps_par
## Request

Make `expanding_window_vols` (`src/simulation/walk_driver.rs:80`) part of the public API, or expose an equivalent entry point, so a consumer that does not drive its walk through `walk_steps` / `walk_steps_par` can price a historical walk the way this crate already does internally.

## Why

`WalkType::volatility()` returns `None` for `Historical` — the volatility is not a parameter of that model, it has to be estimated from the price series. Inside the crate that is solved: `walk_steps_par` computes a rolling estimate with `expanding_window_vols`, whose value at index `i` uses only the log returns of `prices[..=i]`, so it is causal and carries no look-ahead. It is exactly the right behaviour.

The estimator is private to the driver, so a consumer that generates a path with `generate_with_vol` and prices chains itself has no way to reach it. In OptionChain-Simulator that is the v2 rolling simulator: it deliberately stops materialising one chain per step, generates the price path directly, and therefore falls back to a constant volatility for `Historical` walks. The same series and seed priced through the two paths disagree — not on the price path, which matches, but on the volatility used to price it.

The alternatives on our side are both worse than a small upstream change. Duplicating the mathematics gives a second copy to keep in step with this crate on every upgrade, which is a cost we already carry for the seeded walk kernels and would rather not double. Refusing `Historical` removes a model that works.

## Shape

Anything that lets a caller obtain the per-step estimate for a price series would do:

```rust
pub fn expanding_window_vols(
    prices: &[Positive],
    timeframe: TimeFrame,
) -> Result<Vec<Positive>, ...>
```

Making the existing function `pub` is the smallest version. A method on the walk types, or a small `VolatilityEstimator` trait, would work equally well — the property that matters to us is that the estimate at step `i` depends only on observations up to `i`, and that it is the same function `walk_steps_par` uses, so the two paths cannot drift apart.

## Context

Tracked downstream as joaquinbejar/OptionChain-Simulator#63, which documents the current v1/v2 difference until this lands.

joaquinbejar:
Closes #423. Closes #422. Both land before the 0.19.0 tag, so they ship in that release (the changelog entry for 0.19.0 covers them).

## #423 — `expanding_window_vols` is public

`WalkType::Historical` has no volatility parameter — `WalkType::volatility()` returns `None` for it, because the volatility has to be estimated from the price series. Inside the crate `walk_steps` / `walk_steps_par` solve that with a private expanding-window estimator. A consumer that generates a path with `generate_with_vol` and prices its own chains could not reach it, so it had to either fall back to a constant volatility or duplicate the mathematics; the same series and seed then disagreed on the volatility used to price an identical path (downstream: OptionChain-Simulator#63).

The function is now `pub`, exported from `simulation::` and the prelude, with the same signature it already had:

```rust
pub fn expanding_window_vols(
    prices: &[Positive],
    timeframe: TimeFrame,
) -> Result<Option<Vec<Positive>>, SimulationError>
```

`Option` is kept deliberately: below three prices no sample variance exists, and the estimator says so rather than inventing a number. The documented contract is the property the issue asks for:

- exactly `prices.len()` elements, indexing in step with the path;
- element `i` is a function of `prices[..=i]` only — appending prices never moves an earlier estimate;
- indices 0 and 1 are backfilled with the first computable estimate;
- `Ok(None)` below three prices.

It is the *same function* the driver uses, so the two paths cannot drift.

Tests added: prefix stability checked across every prefix length (the causality property), the three-price threshold, and agreement between the last estimate and whole-series `constant_volatility` + `adjust_volatility` (the expanding window at the last index *is* the full window, so the incremental prefix-sum form and the two-pass form must agree). Plus a doctest showing a flat prefix followed by a jump. The pre-existing look-ahead test is untouched.

## #422 — audit triage finished

Of the five advisories that had the daily audit red on `main` since 2026-08-05:

| Advisory | Crate | Status |
|---|---|---|
| RUSTSEC-2026-0204 | crossbeam-epoch | resolved — patched release via the 0.19.0 dependency refresh |
| RUSTSEC-2026-0185 | quinn-proto | resolved — idem |
| RUSTSEC-2026-0104 | rustls-webpki | resolved — idem |
| RUSTSEC-2026-0235 | rkyv 0.7.46 | waived — lockfile-only, optional dep of `rust_decimal` we never enable |
| RUSTSEC-2025-0119 | number_prefix 0.4.0 | waived — lockfile-only, via an `indicatif` feature we never enable |
| RUSTSEC-2025-0134 | rustls-pemfile 1.0.4 | waived — *build*-dependency of `plotly_static` behind `static_export` |

The `rustls-pemfile` path is `optionstratlib -> plotly -> plotly_static -> [build-dependencies] webdriver-downloader -> reqwest 0.11 -> rustls-pemfile`: it downloads a webdriver on the build machine, is absent from the library and from every default-feature build, and never parses attacker-supplied PEM at runtime. Blocked upstream until `plotly_static` moves off `reqwest` 0.11.

Each waiver in `.cargo/audit.toml` carries its reachability rationale (with the `cargo tree` invocation that proves it), an owner, and a 2027-02-15 review date. With the three known ones waived, the workflow now sets `denyWarnings: true` and `ignore: ''` — `.cargo/audit.toml` is the single source of truth, and a *new* unmaintained or unsound dependency fails the gate instead of passing as a warning.

`cargo audit` is now clean: zero findings, exit 0.

## Also

`README.md` and the crate-level docs it is generated from still said `0.18.0` (they were not refreshed for 0.18.1 either); they now say `0.19.0`.

## Verification

- `cargo clippy --all-targets --all-features --workspace -- -D warnings`, `cargo fmt --all --check`, `cargo audit`: clean.
- `make pre-push`: **8987 passed, 5 failed**; `cargo test --all-features --doc`: 203 passed, 4 failed.
- All 9 failures are static-export tests blocked by a missing local binary — `Invalid WebDriver path '/opt/homebrew/bin/chromedriver': WebDriver executable not found`. The 5 unit failures reproduce identically on a pristine `origin/main` worktree; the 4 doctests are the same mechanism in `curves/surfaces/lib.rs` visualization examples, none of which this branch touches.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -31,13 +31,48 @@
 # Owner        : @joaquinbejar
 # Review by    : 2027-02-15
 # ---------------------------------------------------------------------------
-ignore = ["RUSTSEC-2026-0235"]
+# RUSTSEC-2025-0119 — number_prefix 0.4.0, unmaintained.
+#
+# Reachability : NOT REACHABLE. Lockfile-only, exactly like the rkyv entry
+#                above: `cargo tree --all-features --target all -i
+#                number_prefix` reports that the package is not in the graph
+#                at all. It is recorded as an optional dependency of
+#                `indicatif`, whose feature this crate does not enable.
+# Upgrade path : None to take — an unmaintained advisory has no fixed release.
+#                It would disappear on its own once `indicatif` drops the
+#                dependency.
+# Tracking     : https://github.com/console-rs/indicatif
+# Owner        : @joaquinbejar
+# Review by    : 2027-02-15
+# ---------------------------------------------------------------------------
+# RUSTSEC-2025-0134 — rustls-pemfile 1.0.4, unmaintained.
+#
+# Reachability : NOT LINKED INTO THIS CRATE. It arrives through
+#                `reqwest 0.11`, which is a *build*-dependency of
+#                `plotly_static` (via `webdriver-downloader`), itself pulled in
+#                only by the `static_export` feature:
+#                    optionstratlib -> plotly -> plotly_static
+#                      -> [build-dependencies] webdriver-downloader
+#                      -> reqwest 0.11 -> rustls-pemfile 1.0.4
+#                It runs on the build machine while downloading a webdriver, is
+#                absent from the library and from every default-feature build,
+#                and never parses attacker-supplied PEM at runtime.
+# Upgrade path : Blocked upstream: it needs `plotly_static` to move off
+#                `reqwest` 0.11, which is the crate that pins the old
+#                `rustls-pemfile` line.
+# Tracking     : https://github.com/plotly/plotly.rs
+# Owner        : @joaquinbejar
+# Review by    : 2027-02-15
+# ---------------------------------------------------------------------------
+ignore = [
+    "RUSTSEC-2026-0235",
+    "RUSTSEC-2025-0119",
+    "RUSTSEC-2025-0134",
+]
 
-# Report unmaintained/unsound/notice advisories instead of dropping them
-# silently. Two are currently outstanding, both transitive and both without a
-# fixed release to move to:
-#   * RUSTSEC-2025-0119 — number_prefix 0.4.0 unmaintained, via `indicatif`.
-#   * RUSTSEC-2025-0134 — rustls-pemfile 1.0.4 unmaintained, via `reqwest`
-#     (only reachable with the `async` feature enabled).
+# Every other advisory class fails the run, so a new unmaintained or unsound
+# dependency cannot slip in as a silent warning: with these three waived, the
+# audit is expected to be clean, and `denyWarnings` in the workflow can be
+# turned on without turning the gate into noise.
 informational_warnings = ["unmaintained", "unsound", "notice"]
 severity_threshold = "none"
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -19,3 +19,16 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions-rust-lang/audit@v1
         name: Audit Rust Dependencies
+        with:
+          # Advisories are waived exclusively through `.cargo/audit.toml`,
+          # which carries the reachability rationale, owner and review date for
+          # each entry. Keep this input empty so that file stays the single
+          # source of truth.
+          ignore: ''
+          # Informational advisories (unmaintained / unsound / notice) must fail
+          # the run: with the three known-unreachable ones waived in
+          # `.cargo/audit.toml`, anything new is a real finding.
+          denyWarnings: 'true'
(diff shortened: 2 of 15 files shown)


joaquinbejar:
Pushed a third commit: `cargo test` no longer spawns a browser.

Two separate problems were hiding behind the "static export fails locally" symptom:

1. **Five tests need a matched WebDriver.** `plotly_static` drives a real browser for PNG/SVG export, so it needs a `chromedriver` whose major version matches the installed Chrome (locally: driver 138 vs Chrome 151, plus a `WEBDRIVER_PATH` pointing at a binary that no longer exists). They now carry `#[ignore = "needs a WebDriver matching the installed browser; run with --ignored"]`, joining the six sibling tests in `plotly_test.rs` that were already ignored for the same reason.
2. **One test really opened a browser window.** `test_render_browser` passes `OutputType::Browser`, which calls `Plot::show()` — that writes a temporary HTML file and hands it to the default browser. The comment above it said "this test doesn't actually open a browser"; it did, on every single run. Now ignored, with the comment corrected.

Four doc examples in `visualization/`, `curves/visualization/` and `surfaces/visualization/` also wrote a PNG (and one called `show()`) when executed; they are `no_run` now, so they still compile as documentation without needing a browser.

Nothing is lost: `make test-visual` runs exactly the ignored set, for when you do have a matching driver. `make test` also gained the comma it was missing in `--features static_export,plotly` — `plotly` was being parsed as a test-name filter, not a feature, so that third invocation was only running tests whose name contained "plotly".

`cargo test --all-features` is now **4495 passed, 0 failed, 79 ignored** locally, with no browser windows.

PR #426, opened 2026-08-20
fix(greeks): pass carry rate b = r - q into analytic Greeks d1/d2

changshenhan:
## Summary

Fixes a dividend-yield bug family in the analytic Greeks: **`d1`/`d2` were called with the bare risk-free rate `r` instead of the cost of carry `b = r − q`** (Black-Scholes with continuous dividends, Merton 1973). Every analytic Greek silently ignored `dividend_yield`.

### Root cause

The closed-form pricing kernels already use the carry model (`let b = option.risk_free_rate - option.dividend_yield.to_dec()`), but the **analytic Greeks passed `risk_free_rate` directly** into `d1`/`d2` — 17 call sites in `src/greeks/equations.rs`, plus `probability_keep_under_strike` in `src/pricing/utils.rs`. For any underlying with a nonzero dividend yield, every Greek was computed with the wrong `d1`/`d2`.

A second, related defect: **theta** did not discount the `S · n(d1)` term by `e^{-qT}`, overstating `|theta|` for dividend-paying underlyings.

### Changes

- `src/greeks/equations.rs` — pass `option.risk_free_rate - option.dividend_yield.to_dec()` at all 17 `d1`/`d2` call sites; add the `e^{-qT}` discount to theta's common term.
- `src/pricing/utils.rs` — `probability_keep_under_strike` now uses the carry rate.
- `src/greeks/utils.rs` — rename `d1`/`d2`'s third parameter `risk_free_rate` → `carry_rate` and document `b = r − q` so callers can't reintroduce the bug.
- Updated expected values across `model`, `strategies/build`, `volatility`, `chains`, `surfaces`, `curves`, and the equation tests to the corrected formulas.
- Added `test_dividend_high_q_carry_regression`: a long call with `q = 0.08` (carry `b = −0.03` over one year) pins **all 11 Greeks** to the closed-form reference, cross-checked against an independent Python implementation and finite differences.

### Why the signs of the updated values are preserved

The prior `vanna` expression `-(T · −q).exp()` evaluates to `-e^{-qT}`, which is the **correct** Merton vanna `∂Δ/∂σ = -e^{-qT}·n(d1)·d2/σ` (`∂d1/∂σ = -d2/σ`). I verified this against finite differences of the closed-form delta before touching any assertion — every Greek here changes only in magnitude (carry effect), never in sign.

### Verification

- `cargo test` — **3871 lib + 382 integration + 12 property + 207 doc tests, 0 failures**
- `cargo clippy --all-targets` — 0 warnings
- All 11 regression values match an independent Python Black-Scholes Merton reference
- Finite-difference audit of delta/gamma/vega/theta/rho/rho_d/vanna/veta against the option value

Co-Authored-By: Claude <noreply@anthropic.com>


joaquinbejar:
Merged as 18fc11fd. Thank you for this — it is a genuinely valuable contribution.

The diagnosis is correct and it found a real defect that had been sitting in the
library for a long time: the analytic Greeks were passing the bare risk-free rate
into `d1`/`d2` while `calculate_d_values` in the pricer already used `b = r - q`,
so the Greeks were not derivatives of the crate's own price for any dividend-paying
underlying. Before merging I had the eleven Greeks re-derived independently against
the Merton closed forms and against central finite differences of the price, at
S = K = 100, T = 1, sigma = 0.30, r = 0.05, q = 0.08. Every post-PR value agrees to
about 2.5e-15 relative, including the two discriminating cases: rho keeps `e^{-rT}`
rather than `e^{-bT}`, and theta carries both the newly added `e^{-qT}` on the
`S n(d1)` term and the `+q S e^{-qT} N(d1)` term. All 17 call sites are converted
with none missed, and the sites that legitimately use a different carry (Black-76
with b = 0, Garman-Kohlhagen with r_d - r_f) were correctly left alone. The vanna
rewrite is an exact algebraic no-op, as you said in the description.

A few follow-ups I am picking up myself rather than sending back to you, since most
of them predate your patch:

- `cargo fmt --all --check` fails on the branch with 47 hunks (clean on main). This
  is on me: `format_check.yml` runs `make fmt`, which formats and exits 0, so the
  checking target `fmt-check` is invoked by no workflow and this class of breakage
  is invisible in CI. I am fixing the workflow too.
- Four `charm` expectations in `src/strategies/build/model.rs` were left at their
  pre-PR values (lines 425, 483, 541) while line 599 was updated, so two identical
  put spreads are now pinned to different numbers, both sitting at about 98% of
  their 1e-4 tolerance.
- `vomma` and `veta` apply `option.quantity` twice, because `vega()` already carries
  it. That is a pre-existing bug, but the rewritten literals in
  `src/model/option.rs` encode 4x the one-lot value for a two-lot position, so it
  needs fixing on both sides at once.
- The 1e-28 tolerance in `test_dividend_high_q_carry_regression` is tighter than the
  values can support: they pass through a f64 CDF in `statrs`, so real precision is
  about 1e-16. The statrs 0.18 to 0.19 bump one commit before your base would have
  broken it without any change to the mathematics.
- The `d1` rustdoc formula block still reads `r` where the parameter is now
  `carry_rate`.

None of that takes away from the fix. Thanks again for taking the time to write it
up so carefully, and for the finite-difference audit in the description; that is
what made this quick to verify.

Issue #427, opened 2026-08-27
Store the full greek set on OptionData, per option style


joaquinbejar:
## Overview

`OptionData` stores three greeks: `delta_call`, `delta_put` and `gamma`
(`src/chains/optiondata.rs:121-129`). Consumers that render or risk-manage a
position need the full first-order set **and** the second-order set.

They *can* get there today: `impl TryFrom<&OptionData> for Options` is public
(`src/model/option.rs:810`) and the greek functions are re-exported
(`src/greeks/mod.rs:192-196`), so a consumer can convert and compute. But that
means every consumer pays twelve greek computations per style per strike on every
read, and each one re-derives the same conversion boilerplate, including the trap
that the `TryFrom` hardcodes `OptionStyle::Call`, `Side::Long` and
`quantity: ONE`.

All twelve greeks are already implemented here as public free functions. The gap
is that the chain data does not carry them, so the work is repeated per read
rather than done once at build time, where `calculate_delta` and
`calculate_gamma` already run.

This is an optimisation and an ergonomics fix, not a blocker: OptionChain-Simulator
is shipping the greeks over its own REST surface using the public `TryFrom`
(joaquinbejar/OptionChain-Simulator#73). Landing this lets it drop that
boilerplate and stop recomputing on the request path.

### What already exists

Twelve public free functions, all `fn(option: &Options) -> Result<Decimal, GreeksError>`:

| Greek | Order | Fn | Line |
|-------|-------|----|------|
| delta | 1st (∂V/∂S) | `delta` | `src/greeks/equations.rs:476` |
| theta | 1st (∂V/∂t) | `theta` | `src/greeks/equations.rs:800` |
| vega | 1st (∂V/∂σ) | `vega` | `src/greeks/equations.rs:948` |
| rho | 1st (∂V/∂r) | `rho` | `src/greeks/equations.rs:1079` |
| rho_d | 1st (∂V/∂q) | `rho_d` | `src/greeks/equations.rs:1230` |
| gamma | 2nd (∂²V/∂S²) | `gamma` | `src/greeks/equations.rs:659` |
| vanna | 2nd (∂²V/∂S∂σ) | `vanna` | `src/greeks/equations.rs:1366` |
| vomma | 2nd (∂²V/∂σ²) | `vomma` | `src/greeks/equations.rs:1491` |
| charm | 2nd (∂²V/∂S∂t) | `charm` | `src/greeks/equations.rs:1760` |
[...]

PR #429, opened 2026-08-27
fix(greeks): stop applying quantity twice in vomma and veta, and clean up after #426

joaquinbejar:
Follow-up to #426, covering the five items recorded in #427.

## The real bug: `vomma` and `veta` scale as quantity squared

`vomma` (`src/greeks/equations.rs`) and `veta` call `vega(option)?`, which
already returns `vega * quantity`, and then multiply by `quantity` a second
time. Both are first derivatives of vega and are linear in position size, so a
10-lot position reported 10x the true vomma, and the error propagated into
`PortfolioGreeks`.

The defect pre-dates #426. What made it urgent is that #426 rewrote the two
tests that pin it into fresh 28-digit literals encoding exactly 4x the one-lot
value for a 2-lot position, so it was cemented by green assertions.

Fixed on both sides, plus `test_vomma_and_veta_scale_linearly_in_quantity`,
which asserts the scaling law directly at 2, 5 and 10 lots rather than pinning
a magic number.

Nine `tests/unit/strategies/delta/*` integration tests were pinning the same
quantity-squared aggregates. Their expectations are re-derived, not refitted:

- The seven uniform 2-lot strategies have a correct value of exactly half the
  old one, which is the algebra of removing one factor of `quantity` per leg.
- The two butterflies have mixed leg quantities (1/2/1 and 3/6/3), so no single
  scale factor applies. Both were recomputed leg by leg from the Merton closed
  forms with mpmath at 40 significant digits. They match the implementation to
  ~3e-17 (vomma) and ~4e-18 (veta), and the old literals reproduce exactly the
  quantity-weighted buggy sum.

## Stale `charm` expectations

`src/strategies/build/model.rs` had four `charm` values left at their pre-#426
numbers while a fifth was updated. `test_strategy_bear_put_spread` and
`test_strategy_bull_put_spread` build the identical two legs, so their aggregate
charm is the same number, yet they asserted different values. The three stale
literals survived only on a 10x looser tolerance, at 98.0% and 98.8% of their
1e-4 budget: any sub-1e-6 perturbation turned them red with the mathematics
unchanged.

Re-derived at 50 dps: call spreads 0.2091949851158175385267637, put spreads
0.2091401920964687698946958, both now at the 1e-5 tolerance already used by the
bull put spread.

## `1e-28` regression tolerance

`test_dividend_high_q_carry_regression` asserted all eleven Greeks at `1e-28`.
Those values pass through `big_n`, which evaluates the normal CDF in f64 via
statrs and widens the result with `Decimal::from_f64`, so genuine precision is
~1e-16; on values of magnitude 0.1 to 0.5, `1e-28` is roughly one unit in the
last Decimal place. Measured gap to an mpmath 50-dps reference: delta
-7.085e-16, vega -9.238e-16, vanna -7.698e-16. The statrs 0.18 to 0.19 bump in
`fe82222c`, one commit before that PR's base, would have broken this test
without changing any of the mathematics.

Relaxed to `1e-12`, which still separates these values from the pre-#426
dividend-blind ones: they differ in the second significant digit (delta
0.5762569743 vs 0.4799640108). The provenance comment is corrected too, since
no f64 reference can supply digits 17 through 28.

## `d1` rustdoc

The formula block and legend still documented the third argument as `r`, the
risk-free rate, while #426 renamed the parameter to `carry_rate` and documented
it correctly three lines below. The same doc block stated two meanings for one
argument and the more prominent one was wrong. `d1`/`d2` are `pub` and in the
prelude, so a caller following it for S=K=100, T=1, sigma=0.2, r=0.05, q=0.03
would pass `0.05`, get d1=0.35 and a delta of 0.6180 where `option.delta()`
returns 0.5622.

## Formatting, and why CI never noticed

Merging #426 left `main` with 47 unformatted hunks, so `make check` failed.
`cargo fmt --all` applied.

The reason this was invisible: `.github/workflows/format_check.yml` ran
`make fmt`, which `Makefile` defines as `cargo +stable fmt --all`, the
*formatter*. It rewrote files inside the runner and exited 0 whatever the input,
so no pull request has ever actually been format-checked. The checking target
`fmt-check` was invoked by no workflow. The job now runs `make fmt-check`.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` clean
- 3872 lib + 12 property + 383 integration + 207 doc tests, 0 failures

Refs #427

### DIFF
--- a/.github/workflows/format_check.yml
+++ b/.github/workflows/format_check.yml
@@ -42,4 +42,4 @@ jobs:
         run: sudo apt-get install make
 
       - name: Format check
-        run: make fmt
+        run: make fmt-check
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.19.1] - 2026-08-28
+
+Correctness and housekeeping follow-up to the cost-of-carry fix in 0.19.0. No
+API changes.
+
+### Fixed
+
+- `vomma` and `veta` applied the position quantity twice, because `vega`
+  already carries it. Both are first derivatives of vega and are linear in
+  position size, so a 10-lot position reported 10x the true value and the error
+  propagated into `PortfolioGreeks`.
+- `d1`'s documentation still described its third argument as the risk-free rate
+  after it was renamed to `carry_rate`, stating two different meanings for the
+  same argument in one doc block. `d1` and `d2` are public and in the prelude,
+  so a caller following it disagreed with `option.delta()` by about 10%.
+- Four stale `charm` expectations meant two tests over the identical portfolio
+  asserted different numbers, both within 2% of their tolerance.
+- `test_dividend_high_q_carry_regression` pinned eleven greeks at `1e-28`, far
+  tighter than the f64 normal CDF behind them supports. Relaxed to `1e-12`.
+
+### Changed
+
+- The `format_check` CI job ran `make fmt`, which formats rather than checks and
+  exited 0 whatever the input, so no pull request was ever format-checked. It
+  now runs `make fmt-check`.
+
 ## [0.19.0] - 2026-08-17
 
 Dependency refresh: every dependency moved to its latest stable minor, and the
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "optionstratlib"
-version = "0.19.0"
+version = "0.19.1"
 edition = "2024"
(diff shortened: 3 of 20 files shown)


joaquinbejar (commented):
APPROVE. The quantity correction is linear, the regression expectations are re-derived, the patch bump is appropriate, and format, lint, build, and semver checks pass. The audit failure is inherited from the existing optional Plotly dependency chain.

joaquinbejar:
Thanks. One thing changed since you looked, the audit failure is gone. It was h2 0.3.27 arriving as a build dependency of plotly_static, waived in #433, and after rebasing on main all seven checks pass here including audit and coverage.

Issue #428, opened 2026-08-27
Greeks aggregation ignores Side for every greek except delta


joaquinbejar:
## Overview

`Greeks::greeks()` and the aggregate greek methods on the `Greeks` trait return
sign-inconsistent results for any strategy holding short legs. `delta` respects
`Side::Long` / `Side::Short`; **every other greek does not**. A short iron condor
therefore reports a correctly signed net delta next to a net gamma, theta, vega
and rho carrying the sign of the equivalent long position.

Any consumer aggregating position greeks through this trait is showing wrong
numbers, and wrong in the direction that matters: a short premium position looks
like it is losing to time decay when it is collecting it.

## Evidence

The long/short sign is computed in exactly one place in the whole file:

```
src/greeks/equations.rs:525:    let sign = if option.is_long() {
```

That is the only occurrence of `is_long()` or `is_short()` in
`src/greeks/equations.rs`, and it is inside `delta`, used at lines 541, 559 and
560.

Every other greek scales by `option.quantity` and by nothing else:

| Fn | Line | Scales by quantity | Applies long/short sign |
|----|------|--------------------|-------------------------|
| `delta` | 476 | `:563` | **yes**, `:525` |
| `gamma` | 659 | `:687` | no |
| `theta` | 800 | `:846` | no |
| `vega` | 948 | `:971` | no |
| `rho` | 1079 | `:1119` | no |
| `rho_d` | 1230 | `:1257` | no |
| `vanna` | 1366 | `:1394` | no |
| `vomma` | 1491 | `:1516` | no |
| `veta` | 1605 | `:1639` | no |
| `charm` | 1760 | `:1798` | no |
| `color` | 1909 | `:1951` | no |

The sign cannot ride on the quantity instead: `Options::quantity` is
`Positive` (`src/model/option.rs:149`), so it can never be negative. `Side` is
the only carrier of direction, and only `delta` reads it.

The aggregation path makes this observable. `impl Greeks for IronCondor`
(`src/strategies/iron_condor.rs:1161`) returns all four legs, including
`short_call.option` and `short_put.option` which carry `Side::Short`. The
default `Greeks::gamma()` sums them unsigned, so the two short legs contribute
with the wrong sign, and `greeks()` (`src/greeks/equations.rs:160`) packages the
result into a `Greek` alongside a delta that **was** signed correctly.

## Reproduction

Build any strategy with at least one short leg (a `ShortStraddle`, a
[...]

PR #430, opened 2026-08-27
feat(chains): store the full greek set on OptionData, per option style

### LINKED ISSUE #427: Store the full greek set on OptionData, per option style

joaquinbejar:
Closes #427. **Stacked on #429** — the base branch is `fix/greeks-followup-426`,
so #429 must merge first.

## What and why

`OptionData` carried three greeks. Any consumer needing the rest converted
through the public `TryFrom<&OptionData> for Options` and recomputed twelve
greeks per style per strike **on every read**, each re-deriving the same
boilerplate, including the trap that the conversion hardcodes `Side::Long` and
`quantity: ONE`.

Adds `greeks_call` and `greeks_put` as `Option<GreeksSnapshot>`, populated once
by a new `OptionData::calculate_greeks`.

## Prerequisite: `greeks()` used to fail at expiry

Ten of the twelve greeks early-return zero when time to expiry is zero. `rho_d`
had no guard at all and went straight to `d1`; `vanna` guarded a zero implied
volatility but not a zero expiry. `d1` rejects a zero expiry with
`InvalidTime`, and `Greeks::greeks()` is `?`-chained across all twelve, so the
whole set was lost at expiry.

Not a corner case: `ExpirationDate::get_years()` yields zero for any past date,
`load_from_json` calls `update_greeks()`, and the integration suite loads chain
fixtures that expired in 2024 and 2025. Both guards added, with tests.

## Performance, and why the feature is opt-in

Measured with the benchmark added in the first commit:

| strikes | before | with snapshots | factor |
|---------|--------|----------------|--------|
| 11 | 434 us | 3.00 ms | 7.0x |
| 21 | 839 us | 5.72 ms | 6.8x |
| 31 | 1.202 ms | 8.57 ms | 7.1x |

Criterion reports +598%, +583%, +586%, all at p = 0.00. The walk generators
were measured separately and are unaffected (1.152 -> 1.217 ms at 10 steps),
because the chains they build per step are far smaller.

So the snapshots are opt-in:

- `OptionChainBuildParams::with_greek_snapshots(true)` turns them on; the flag
  is off by default and is skipped during serialization while false, so
  existing payloads stay byte-identical and older ones still deserialize.
- `update_greeks` keeps its previous cost and scope, delta and gamma only. The
  full set has its own entry point, `OptionChain::update_greek_snapshots`.

With the flag off, `build_chain` measures back at baseline (-85% against the
unconditional version, i.e. 434 us / 885 us / 1.257 ms).

For OptionChain-Simulator this is still a net win: it pays those twelve greeks
on every read today, and they become a one-off cost per chain.

### Where the time actually goes

Per-call cost of the Decimal kernels, measured on this machine: `n` (the normal
pdf) 11.68 us, `exp` 6.61 us, `d1` 4.85 us, `big_n` 2.41 us, against a whole
`greeks()` of 473 us. Each greek re-derives all of them from scratch — roughly
27 `d1`, 12 `n`, 20 `exp` and 15 `big_n` per style, which is about 93% of the
total.

Caching `d1`/`d2` alone would recover only ~28%. Computing the whole shared set
once per option and threading it through all twelve should cut `greeks()` by
about 5x, which would bring the build-path cost down to roughly 1.4x. That is a
change to the numeric kernel every strategy depends on, so it belongs in its own
PR; follow-up issue linked below.

## Design notes

- `OptionData::new` is untouched. It already takes 18 positional arguments and
  is called from src, tests, examples and benches; snapshots are computed, never
  passed in.
- The mirror fields are computed independently rather than read back out of the
  snapshots. `delta` and `gamma` return a value at a zero implied volatility
  where the full set does not, so routing them through the snapshot would turn
  fields that are populated today into `None`.
- `alpha`'s `Decimal::MAX` sentinel is mapped to `None`. Publishing it would put
  a 29-digit number on the wire that consumers would read as a real value. The
  guard is defensive: `alpha` tests gamma first, so reaching the sentinel needs
  a non-zero gamma alongside an exactly cancelling theta.
- `set_volatility` and `set_extra_params` drop the snapshots rather than leave
  them stale. The mirror fields are left alone, since a broker feed populates
  those.
- `GreeksSnapshot` loses `#[serde(deny_unknown_fields)]`. It is a wire type now;
  with the attribute, a thirteenth greek would break deserialization of new
  payloads by older consumers. This is a deliberate deviation from the
  crate-wide serde convention.
- Field docs state the sign convention: the snapshot is for one long contract,
  so a consumer holding a short negates `delta` and nothing else (see #428).

## Versioning

This bumps the crate to **0.20.0**. `OptionData` has all-`pub` fields, no
`#[non_exhaustive]`, and is in the prelude, so adding two fields breaks any
downstream crate that builds one with an exhaustive struct literal.
`cargo-semver-checks` reports it as `constructible_struct_adds_field`. Under 0.x
a breaking change takes the minor, and the CHANGELOG carries a *Migration* note:
the fix is `..Default::default()` or the two new fields, and `OptionData::new`
needs no edit.

Adding `#[non_exhaustive]` would make future field additions non-breaking, at the
cost of removing literal construction for downstream crates entirely. Not done
here; worth considering separately.

## One thing still for you

**utoipa.** `Option<T>` forwards `T::schemas`, so a downstream
`components(schemas(OptionData))` picks up `GreeksSnapshot` transitively. This
crate has no `#[derive(OpenApi)]`, so nothing here proves it. If the REST
service registers `OptionData` explicitly, it must also register
`GreeksSnapshot` or the spec will carry a dangling `$ref`.

## Verification

- `cargo test --all-features --workspace`: 3879 lib + 12 property + 412


src/chains/optiondata.rs
```diff
@@ -157,6 +167,37 @@ pub struct OptionData {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     pub extra_fields: Option<Value>,
+
+    /// The full twelve-greek set for the call at this strike.
+    ///
+    /// # Sign and size convention
+    ///
+    /// The snapshot is computed for **one long contract**: the option is built
+    /// with [`Side::Long`] and `quantity: Positive::ONE`. A consumer holding a
+    /// short position must negate `delta` and **must not** negate the other
```
joaquinbejar: 🔴 **Critical**: A snapshot computed for one long contract is the derivative of the long position value, so applying it to a short position requires negating every Greek, not only delta. Encoding the current #428 bug as the consumer contract will invert gamma, theta, vega, rho, and higher-order risk for short holdings. Please keep these as per-long-contract values and apply the position side uniformly to all fields when turning them into exposure.

src/chains/optiondata.rs
```diff
@@ -157,6 +167,37 @@ pub struct OptionData {
...
+    /// eleven, because the greek functions apply the long/short sign in `delta`
+    /// alone (see issue #428). Scaling by position size is likewise the
+    /// consumer's job.
+    ///
+    /// # When this is `None`
+    ///
+    /// Populated by [`Self::calculate_greeks`], which the chain build path and
```
joaquinbejar: 🟠 **Warning**: This claim is false and leaves a stale-cache path in the public API. `OptionChain::update_expiration_date` mutates every row expiration and then calls `update_greeks()`, which refreshes only delta and gamma, so existing `greeks_call` and `greeks_put` snapshots remain from the old expiry. Invalidate or recompute the full snapshots on that path and point this documentation to `update_greek_snapshots` instead.

joaquinbejar (commented):
REQUEST_CHANGES. The version bump resolves the semver break, but the new snapshot API still defines an incorrect short-position sign contract and can retain stale snapshots after the chain expiration changes.

src/chains/optiondata.rs
joaquinbejar: You are right, a short is worth minus V so every derivative flips, not just delta. I had written the #428 aggregation behaviour into the contract, which is exactly backwards. Docs now say negate all twelve and note that the trait asymmetry is a separate bug these fields do not inherit.

src/chains/optiondata.rs
joaquinbejar: Confirmed, chain.rs writes the expiry straight onto each row so set_extra_params never runs. That path now invalidates both snapshots with a test pinning it, and the docs point at update_greek_snapshots.

joaquinbejar:
Both fixed in 71f1d116. The sign contract now says negate all twelve for a short, since the snapshot is a derivative of the long position value, and update_expiration_date drops the snapshots instead of leaving them pinned to the old expiry. Also worth noting the audit failure you mentioned is gone, it was waived in #433 and this branch is rebased on the new main.

PR #432, opened 2026-08-27
perf(greeks): compute the shared Black-Scholes kernels once per option

### LINKED ISSUE #431: Compute the shared Black-Scholes kernels once per option instead of per greek
## Overview

Each of the twelve greek functions in `src/greeks/equations.rs` takes an
`&Options` and re-derives every intermediate value it needs from scratch: `d1`,
`d2`, the normal pdf `n`, the normal cdf `big_n`, `exp(-qT)`, `exp(-rT)`,
`sqrt(T)`. `Greeks::greeks()` calls all twelve in sequence, so the same handful
of expensive `Decimal` transcendentals is computed dozens of times for one
option.

`alpha` (`src/greeks/equations.rs:1268`) is the clearest case: it calls
`gamma()` and `theta()` in full, both of which recompute `d1` and `d2`, when
`greeks()` already has both values in scope.

## Measurements

Per-call cost of each kernel, and of the whole aggregate, measured on an M-series
Mac with a 30-day ATM option:

| kernel | per call |
|--------|----------|
| `n` (normal pdf) | 11.68 us |
| `exp` | 6.61 us |
| `d1` | 4.85 us |
| `big_n` (normal cdf) | 2.41 us |
| `Greeks::greeks()` (all twelve) | 473 us |

Counting the calls inside `greeks()` — roughly 27 `d1`, 12 `n`, 20 `exp` and
15 `big_n` per option style — accounts for about 93% of the 473 us. The
remaining 7% is `Decimal` multiplication and branching.

Computing each shared value once (1 `d1`, 1 `d2`, 2 `n`, 3 `exp`, 4 `big_n`)
would cost roughly 60 to 90 us, so `greeks()` should get about **5x faster**.

## Why it matters now

#430 stores the twelve greeks per option style on `OptionData`. Computing them
made `build_chain` roughly 7x slower (1.202 ms to 8.57 ms on a 31-strike chain),
which is why that PR ships the snapshots behind an opt-in flag rather than
enabling them on the existing build path.

A 5x cut here would bring that cost to roughly 1.4x, at which point the flag
could plausibly be removed and the snapshots computed unconditionally, which is
what #427 originally specified.

## Shape of the change

Introduce a precomputed context — say `BlackScholesInputs` — holding `d1`, `d2`,
`n(d1)`, `big_n(±d1)`, `big_n(±d2)`, `exp(-qT)`, `exp(-rT)` and `sqrt(T)`,
computed once from an `&Options`. Give each greek an internal `*_with(&inputs)`
form and keep the twelve public `fn(&Options)` signatures as thin wrappers that
build the context and delegate, so nothing downstream changes.

`Greeks::greeks()` then builds the context once and calls the twelve internal
forms.

Note for `alpha`: the saving there must be taken in the free function or in the
context, **not** by changing `Greeks::alpha()` to divide the aggregate gamma by
the aggregate theta. The trait's `alpha()` sums per-option ratios while
`gamma()` and `theta()` sum numerator and denominator separately, and the two
coincide only for a single option.

## Constraints

- Every existing greek value must be bit-identical afterwards. The twelve are
  pinned by `test_dividend_high_q_carry_regression` and by the per-greek tests
  in `src/model/option.rs` and `tests/unit/strategies/delta/`.
- The public signatures of the twelve free functions must not change.
- Report the measured before and after on the `Chain Generators/build_chain`
  benchmark and on `Greeks::greeks()`.

## Acceptance criteria

- [ ] `Greeks::greeks()` computes `d1`, `d2`, `n`, `big_n`, `exp` and `sqrt`
      once each per option style
- [ ] Every greek returns a bit-identical value to today; the existing
      regression tests pass unchanged
- [ ] `alpha` no longer recomputes `gamma` and `theta` from scratch
- [ ] Measured speedup of `Greeks::greeks()` reported in the PR, with the
      `build_chain` delta
- [ ] A decision recorded on whether #430's `greek_snapshots` opt-in flag can be
      removed afterwards

joaquinbejar:
Closes #431. **Stacked on #430**, which is itself stacked on #429.

## The problem

Each of the twelve greek functions re-derived `d1`, `d2`, the normal pdf and
cdf, and the two discount factors from scratch. One `Greeks::greeks()` call
therefore evaluated the same handful of expensive `Decimal` transcendentals
dozens of times.

Measured per call on an M-series Mac:

| kernel | per call |
|--------|----------|
| `n` (normal pdf) | 11.68 us |
| `exp` | 6.61 us |
| `d1` | 4.85 us |
| `big_n` (normal cdf) | 2.41 us |
| `Greeks::greeks()` (all twelve) | 473 us |

Counting the calls inside `greeks()` — roughly 27 `d1`, 12 `n`, 20 `exp` and 15
`big_n` per option style — accounts for about 93% of the 473 us.

## The change

`BlackScholesKernels` holds those intermediates for one live European option.
Each greek gains an internal `*_with` form that reads them instead of
recomputing. The twelve public `fn(&Options)` signatures are unchanged and keep
their own degenerate branches, so nothing downstream moves.

Everything past `d1` is lazy. An earlier eager version made the
delta-and-gamma-only path about 20% slower, because those two paid for kernels
they never read; with `OnceCell` per field, each caller pays only for the
kernels its own greek touches, and the aggregate shares them all.

Two further savings:

- `d2` is derived from the cached `d1` as `d1 - sigma * sqrt(T)`, exactly as
  `greeks::d2` derives it, removing a second full `d1` per call.
- `alpha` no longer recomputes `gamma` and `theta`. The aggregate passes in the
  two it already holds, through the new `alpha_from`. As #431 required, that
  saving is **not** taken inside `Greeks::alpha()`: the trait sums per-option
  ratios while `gamma()` and `theta()` sum numerator and denominator separately,
  and the two coincide only for a single option.

`Greeks::greeks()` now aggregates option by option rather than greek by greek.
`Decimal` addition is exact, so the sums are identical to accumulating each greek
across every option in turn.

## Results

Same benchmark code on both branches, back to back:

| case | before (#430) | after | speedup |
|------|---------------|-------|---------|
| 11 strikes + greeks | 3.063 ms | 650 us | 4.7x |
| 21 strikes + greeks | 6.185 ms | 1.260 ms | 4.9x |
| 31 strikes + greeks | 9.266 ms | 1.954 ms | 4.7x |
| 11 strikes, no greeks | 447 us | 433 us | unchanged |
| 21 strikes, no greeks | 851 us | 869 us | unchanged |
| 31 strikes, no greeks | 1.321 ms | 1.262 ms | unchanged |

`Greeks::greeks()` itself: **473 us to 73.2 us, 6.5x**.

The cost of turning the snapshots on drops from about **7x** a chain build to
about **1.5x**.

## Bit-identical

No value changes. The whole suite passes unchanged, including
`test_dividend_high_q_carry_regression`, which pins eleven greeks to 1e-12, and
the nine per-strategy delta integration tests that pin aggregate values across
multi-leg positions. Those are the regression net for the reordered aggregation.

## The remaining decision, for you

#431's last acceptance criterion asks whether #430's `greek_snapshots` opt-in
flag can now be removed. At 1.5x rather than 7x, computing the snapshots
unconditionally is defensible and would match what #427 originally specified.
I have left the flag in place, because 1.5x is still real for simulation-heavy
paths that build chains and never read the greeks. Your call; happy to remove it
in a follow-up if you want the simpler API.

## Verification

- `cargo test --all-features --workspace`: 3879 lib + 12 property + 412

### DIFF
--- a/src/greeks/equations.rs
+++ b/src/greeks/equations.rs
@@ -4792,3 +4792,284 @@ pub mod tests_color_equations {
         );
     }
 }
+
+#[cfg(test)]
+mod tests_shared_kernel_equivalence {
+    use super::*;
+    use crate::greeks::utils::d2 as fresh_d2_fn;
+    use crate::model::types::OptionType;
+    use crate::{ExpirationDate, Options};
+    use positive::{Positive, pos_or_panic};
+    use rust_decimal_macros::dec;
+
+    /// Every branch `kernels_for` distinguishes, so the fast path and the
+    /// per-greek fallbacks are both exercised.
+    fn branches() -> Vec<(&'static str, OptionType, Positive, Positive)> {
+        vec![
+            // name, type, time to expiry in days, implied volatility
+            (
+                "live european",
+                OptionType::European,
+                pos_or_panic!(30.0),
+                pos_or_panic!(0.2),
+            ),
+            (
+                "at expiry",
+                OptionType::European,
+                Positive::ZERO,
+                pos_or_panic!(0.2),
+            ),
+            (
+                "zero volatility",
+                OptionType::European,
+                pos_or_panic!(30.0),
+                Positive::ZERO,
+            ),
+            (
+                "non european",
+                OptionType::American,
+                pos_or_panic!(30.0),
+                pos_or_panic!(0.2),
+            ),
+        ]
+    }
+
+    fn option_for(
+        option_type: OptionType,
+        days: Positive,
+        implied_volatility: Positive,
+        style: OptionStyle,
+        side: Side,
+        quantity: Positive,
+    ) -> Options {
+        Options::new(
+            option_type,
+            side,
+            "TEST".to_string(),
+            Positive::HUNDRED,
+            ExpirationDate::Days(days),
+            implied_volatility,
+            quantity,
+            pos_or_panic!(105.0),
+            dec!(0.05),
+            style,
+            pos_or_panic!(0.02),
+            None,
+        )
+    }
+
+    /// The aggregate shares one set of kernels across the twelve greeks and
+    /// reuses gamma and theta for alpha, while the public functions each build
+    /// their own. The two must agree exactly, in every branch, for both styles
+    /// and both sides — including on failure, so the fast path can neither
+    /// succeed where an individual greek errors nor error where they all
+    /// succeed.
+    ///
+    /// The zero-volatility branch is the one that fails today: seven of the
+    /// twelve have no guard for it, so `greeks()` errors there while `delta`
+    /// and `gamma` still return their discrete values.
+    #[test]
+    fn test_greeks_matches_the_individual_functions_exactly() {
+        for (name, option_type, days, iv) in branches() {
+            for style in [OptionStyle::Call, OptionStyle::Put] {
+                for side in [Side::Long, Side::Short] {
+                    let option = option_for(
+                        option_type.clone(),
+                        days,
+                        iv,
+                        style,
+                        side,
+                        pos_or_panic!(3.0),
+                    );
+                    let label = format!("{name} {style:?} {side:?}");
+
+                    let individual = [
+                        ("delta", delta(&option)),
+                        ("gamma", gamma(&option)),
+                        ("theta", theta(&option)),
+                        ("vega", vega(&option)),
+                        ("rho", rho(&option)),
+                        ("rho_d", rho_d(&option)),
+                        ("alpha", alpha(&option)),
+                        ("vanna", vanna(&option)),
+                        ("vomma", vomma(&option)),
+                        ("veta", veta(&option)),
+                        ("charm", charm(&option)),
+                        ("color", color(&option)),
+                    ];
+                    let all_ok = individual.iter().all(|(_, r)| r.is_ok());
+
+                    match (option.greeks(), all_ok) {
+                        (Ok(aggregate), true) => {
+                            let values = [
+                                aggregate.delta,
+                                aggregate.gamma,
+                                aggregate.theta,
+                                aggregate.vega,
+                                aggregate.rho,
+                                aggregate.rho_d,
+                                aggregate.alpha,
+                                aggregate.vanna,
+                                aggregate.vomma,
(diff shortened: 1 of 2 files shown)


src/greeks/equations.rs
```diff
@@ -201,18 +202,38 @@ pub trait Greeks {
...
-        let veta = self.veta()?;
-        let charm = self.charm()?;
-        let color = self.color()?;
+        // Aggregate option by option rather than greek by greek, so the shared
+        // Black-Scholes kernels are computed once per option instead of once
+        // per greek. `Decimal` addition is exact, so the sums are identical to
+        // accumulating each greek across every option in turn.
+        let options = self.get_options()?;
+        let mut delta = Decimal::ZERO;
+        let mut gamma = Decimal::ZERO;
```
joaquinbejar, line 211: 🟠 **Warning**: This numerical refactor changes all twelve public paths but adds no regression tests. The existing `test_greeks_implementation` checks one live long call at `1e-6` and omits alpha, so it cannot prove the claimed bit-identical behavior or protect expiry, zero-volatility, and non-European fallbacks. Please add exact before-and-after reference assertions or an independent legacy-formula baseline for all twelve functions and `greeks()` across those branches.

joaquinbejar (commented):
REQUEST_CHANGES. The shared-kernel refactor looks algebraically consistent and the targeted test suite passes, but the bit-identical branch coverage required by the issue is still missing.

src/greeks/equations.rs
joaquinbejar: Fair, added three tests in c04b9e1d. One walks live, expiry, zero volatility and non European across both styles and sides asserting greeks() matches the twelve functions exactly, including failure parity, another covers the reordered aggregation on a mixed multi leg, and a third pins every cached kernel against a freshly derived value.

joaquinbejar:
Coverage added in c04b9e1d, three tests across every branch the refactor distinguishes. I skipped pinned 28 digit constants on purpose since that is the bit lock we removed in #429, and asserted the fast and slow paths agree instead, which catches what this change can actually break.

PR #434, opened 2026-08-28
perf(greeks): compute the shared Black-Scholes kernels once per option

### LINKED ISSUE #431: Compute the shared Black-Scholes kernels once per option instead of per greek

joaquinbejar:
Closes #431. **Re-lands #432**, whose content never reached `main`.

## Why this PR exists

#432 was merged at 07:39:38Z into `issue-427-greeks-snapshot`. #430 had already
been squashed into `main` at 07:22:33Z, seventeen minutes earlier, so by then
that branch fed nothing. The kernel refactor landed on an orphan.

`main` has the snapshots from #430 but no `BlackScholesKernels`. This branch is
the same two commits cherry-picked onto current `main`; the diff against the
original branch is empty, so the content is identical to what was reviewed.

## What it does

Each of the twelve greek functions re-derived `d1`, `d2`, the normal pdf and
cdf, and the two discount factors from scratch, so one `Greeks::greeks()` call
evaluated the same expensive `Decimal` transcendentals dozens of times. Measured
per call, the pdf `n` at 11.68 us, `exp` at 6.61 us, `d1` at 4.85 us and the cdf
`big_n` at 2.41 us, against a whole `greeks()` of 473 us. Counting the calls
inside it, roughly 27 `d1`, 12 `n`, 20 `exp` and 15 `big_n` per style, accounts
for about 93% of that.

`BlackScholesKernels` holds those intermediates for one live European option,
computed lazily per field so a caller pays only for the kernels its own greek
touches. The twelve public signatures are unchanged and keep their degenerate
branches. `d2` comes from the cached `d1`, removing a second full `d1`, and
`alpha` no longer recomputes `gamma` and `theta`.

## Results

| case | before | after | speedup |
|------|--------|-------|---------|
| 11 strikes + greeks | 3.063 ms | 650 us | 4.7x |
| 21 strikes + greeks | 6.185 ms | 1.260 ms | 4.9x |
| 31 strikes + greeks | 9.266 ms | 1.954 ms | 4.7x |
| 31 strikes, no greeks | 1.321 ms | 1.262 ms | unchanged |

`Greeks::greeks()` itself goes from 473 us to 73.2 us, 6.5x. The cost of
enabling the snapshots drops from about 7x a chain build to about 1.5x.

## Coverage

Carries the tests added in review of #432. One walks every branch
`kernels_for` distinguishes, live European, at expiry, zero volatility and
non-European, across both styles and sides, asserting `greeks()` matches the
twelve public functions exactly and that the two paths agree on failure as well.
Another covers the reordered aggregation on a mixed multi-leg position. A third
pins every cached kernel against a freshly derived value.

No pinned 28-digit constants, deliberately: that is the bit-pattern lock removed
in #429, where a statrs ULP change could break a test without the mathematics
moving.

## Still open for you

With the overhead down to about 1.5x, #430's `greek_snapshots` opt-in flag could
be removed and the snapshots computed unconditionally, which is what #427
originally specified. Left in place here.

## Verification

- `cargo test --all-features --workspace`: 3882 lib + 12 property + 413

### DIFF
--- a/benches/chains/generators.rs
+++ b/benches/chains/generators.rs
@@ -112,6 +112,16 @@ pub fn benchmark_chain_generators(c: &mut Criterion) {
                 black_box(chain)
             })
         });
+
+        // Same build with the full twelve-greek snapshot per option style, the
+        // opt-in path used by consumers that read the snapshots.
+        let with_greeks = build_params.clone().with_greek_snapshots(true);
+        group.bench_function(format!("build_chain {strikes} strikes + greeks"), |b| {
+            b.iter(|| {
+                let chain = OptionChain::build_chain(black_box(&with_greeks));
+                black_box(chain)
+            })
+        });
     }
 
     let params_10 = chain_walk_params(10);
--- a/src/greeks/equations.rs
+++ b/src/greeks/equations.rs
@@ -4792,3 +4792,284 @@ pub mod tests_color_equations {
         );
     }
 }
+
+#[cfg(test)]
+mod tests_shared_kernel_equivalence {
+    use super::*;
+    use crate::greeks::utils::d2 as fresh_d2_fn;
+    use crate::model::types::OptionType;
+    use crate::{ExpirationDate, Options};
+    use positive::{Positive, pos_or_panic};
+    use rust_decimal_macros::dec;
+
+    /// Every branch `kernels_for` distinguishes, so the fast path and the
+    /// per-greek fallbacks are both exercised.
+    fn branches() -> Vec<(&'static str, OptionType, Positive, Positive)> {
+        vec![
+            // name, type, time to expiry in days, implied volatility
+            (
+                "live european",
+                OptionType::European,
+                pos_or_panic!(30.0),
+                pos_or_panic!(0.2),
+            ),
+            (
+                "at expiry",
+                OptionType::European,
+                Positive::ZERO,
+                pos_or_panic!(0.2),
+            ),
+            (
+                "zero volatility",
+                OptionType::European,
+                pos_or_panic!(30.0),
+                Positive::ZERO,
+            ),
+            (
+                "non european",
+                OptionType::American,
+                pos_or_panic!(30.0),
+                pos_or_panic!(0.2),
+            ),
+        ]
+    }
+
+    fn option_for(
+        option_type: OptionType,
+        days: Positive,
+        implied_volatility: Positive,
+        style: OptionStyle,
+        side: Side,
+        quantity: Positive,
+    ) -> Options {
+        Options::new(
+            option_type,
+            side,
+            "TEST".to_string(),
+            Positive::HUNDRED,
+            ExpirationDate::Days(days),
+            implied_volatility,
+            quantity,
+            pos_or_panic!(105.0),
+            dec!(0.05),
+            style,
+            pos_or_panic!(0.02),
+            None,
+        )
+    }
+
+    /// The aggregate shares one set of kernels across the twelve greeks and
+    /// reuses gamma and theta for alpha, while the public functions each build
+    /// their own. The two must agree exactly, in every branch, for both styles
+    /// and both sides — including on failure, so the fast path can neither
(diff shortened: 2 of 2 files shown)


joaquinbejar (commented):
APPROVE. The re-landed shared-kernel refactor preserves public signatures and fallback behavior, the added exact-equivalence tests address the remaining review concern across live and degenerate branches, and all build, coverage, format, lint, and semver checks pass.

PR #435, opened 2026-08-28
fix(greeks): respect Side in every greek, not just delta

### LINKED ISSUE #428: Greeks aggregation ignores Side for every greek except delta

joaquinbejar:
Closes #428.

## The bug

Only `delta` read `Side`. The other eleven greeks scaled by `option.quantity`
and nothing else, and `Options::quantity` is `Positive`, so it can never carry
direction. `Side` was the sole carrier and eleven functions ignored it.

24 strategies implement `Greeks`, so this was observable everywhere: any of them
holding a short leg reported gamma, theta, vega, rho and the higher-order greeks
with the sign of the equivalent long position, next to a delta that *was* signed
correctly. A short-premium position looked as though it was losing to time decay
when it was collecting it.

## The choice

Option 1 from the issue: apply the sign inside every function rather than
removing it from `delta`. The delta-neutral subsystem reads signed `delta()` at
six sites (`src/strategies/delta_neutral/portfolio.rs:119`,
`src/strategies/delta_neutral/model.rs:375,387,405,618,631`) to derive net delta
and hedge ratios, so removing it there would have broken that subsystem or
pushed the sign onto all six callers.

## How the sign lands

Not uniformly, and the split is the point. Applying it to all eleven the same
way would double-apply it in three of them, which is the exact bug shape fixed
in #429 for `quantity`:

- **Eight** apply it where they already applied quantity, through a new
  `signed_quantity` helper. Four multiply directly, four go through the checked
  `d_mul` before a `d_div`.
- **`vomma` and `veta` inherit it** from `vega`, which they are derivatives of,
  exactly as they already inherit quantity since #429.
- **`alpha` is invariant.** It is `gamma / theta`, and a short negates both, so
  the ratio is unchanged. That is the correct result rather than a missing case,
  and it is documented as such.

`delta` is untouched: it applies the sign *before* clamping to `[-1, 1]`, which
is the right order.

The convention is now stated in the `Greeks` trait docs and in all twelve
function doc comments. The file was previously silent about it and inconsistent
with itself. `alpha` gains the doc comment it never had and is re-exported from
`crate::greeks` alongside the other eleven.

## Updated expectations are derived, not refitted

About 60 pinned values change across `src/strategies/build/model.rs`,
`src/volatility/utils.rs` and `tests/unit/strategies/delta/`. Two things back
them:

- A new test per strategy family — single leg, vertical spread, condor,
  straddle — asserts the aggregate equals the manually signed per-leg sum
  `Σ greek(leg)` for all twelve values, computed independently of the aggregate.
  That is the acceptance criterion the issue asks for.
- The spread gamma was checked by hand against a 40-digit reference:
  ±0.0083145207388161153, against the old 0.0309 which was the *unsigned* sum.

Three existing tests became sharper rather than merely different:

- Two legs, long and short, of the identical contract now net to **exactly
  zero** across all twelve. Before, only delta did.
- A long call with a short put at the same strike is a synthetic forward, so the
  six greeks with no `option_style` branch cancel exactly while the five that do
  branch survive. The tests now assert that identity instead of a vague
  "non-zero".

New tests also pin that flipping a leg's side negates eleven greeks and leaves
`alpha` alone, and that a short-premium strategy reports positive theta.

## Not affected

`OptionData::greeks_call` and `greeks_put` are built through
`get_option(Side::Long, style)`, so they remain per-long-contract values. Their
docs now point at the new trait behaviour instead of at this bug.

## Versioning

No bump here, per the maintainer's call. The `## [Unreleased]` CHANGELOG entry
records the breaking behaviour change and carries a migration note: any consumer
that was compensating by negating the eleven unsigned greeks itself must stop.

## Verification

- `cargo test --all-features --workspace`: 3886 lib + 12 property + 418

### DIFF
--- a/src/greeks/equations.rs
+++ b/src/greeks/equations.rs
@@ -620,6 +620,7 @@ where
 ///
 /// Negating a `Decimal` is exact, so this needs no checked multiplication.
 #[inline]
+#[must_use]
 fn signed_quantity(option: &Options) -> Decimal {
     let quantity = option.quantity.to_dec();
     if option.is_long() {
@@ -779,7 +780,14 @@ fn greeks_for(option: &Options) -> Result<Greek, GreeksError> {
 /// by `quantity`. A short position reports the negative of the equivalent long.
 pub fn delta(option: &Options) -> Result<Decimal, GreeksError> {
     if !matches!(option.option_type, OptionType::European) {
-        return crate::greeks::numerical::numerical_delta(option);
+        // The numerical fallback prices through `price_option`, which takes the
+        // absolute value, so it returns a per-contract long sensitivity. Sign
+        // and scale it here to keep the documented convention.
+        return Ok(d_mul(
+            crate::greeks::numerical::numerical_delta(option)?,
+            signed_quantity(option),
+            "greeks::delta::numerical_position_weighted",
+        )?);
     }
     let expiration_date = option.expiration_date.get_years()?;
 
@@ -804,24 +812,27 @@ pub fn delta(option: &Options) -> Result<Decimal, GreeksError> {
     // This happens because at expiration, the option effectively becomes a direct position in the
     // underlying asset (**delta = 1 or -1**) if it is ITM, or has no value (**delta = 0**) if it is OTM.
     if expiration_date == Decimal::ZERO {
-        return match (
+        // These arms already carry the side, so they scale by the bare quantity
+        // rather than by `signed_quantity`, which would apply it twice.
+        let per_contract = match (
             &option.option_style,
             &option.side,
             &option.strike_price,
             &option.underlying_price,
         ) {
             // Call Options
-            (OptionStyle::Call, Side::Long, strike, price) if price > strike => Ok(Decimal::ONE),
-            (OptionStyle::Call, Side::Long, _, _) => Ok(Decimal::ZERO),
-            (OptionStyle::Call, Side::Short, strike, price) if price > strike => Ok(-Decimal::ONE),
-            (OptionStyle::Call, Side::Short, _, _) => Ok(Decimal::ZERO),
+            (OptionStyle::Call, Side::Long, strike, price) if price > strike => Decimal::ONE,
+            (OptionStyle::Call, Side::Long, _, _) => Decimal::ZERO,
+            (OptionStyle::Call, Side::Short, strike, price) if price > strike => -Decimal::ONE,
+            (OptionStyle::Call, Side::Short, _, _) => Decimal::ZERO,
 
             // Put Options
-            (OptionStyle::Put, Side::Long, strike, price) if price < strike => Ok(-Decimal::ONE),
-            (OptionStyle::Put, Side::Long, _, _) => Ok(Decimal::ZERO),
-            (OptionStyle::Put, Side::Short, strike, price) if price < strike => Ok(Decimal::ONE),
-            (OptionStyle::Put, Side::Short, _, _) => Ok(Decimal::ZERO),
+            (OptionStyle::Put, Side::Long, strike, price) if price < strike => -Decimal::ONE,
+            (OptionStyle::Put, Side::Long, _, _) => Decimal::ZERO,
+            (OptionStyle::Put, Side::Short, strike, price) if price < strike => Decimal::ONE,
+            (OptionStyle::Put, Side::Short, _, _) => Decimal::ZERO,
         };
+        return Ok(per_contract * option.quantity.to_dec());
     }
 
     let sign = if option.is_long() {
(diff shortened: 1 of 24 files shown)


src/greeks/equations.rs
```diff
@@ -744,6 +772,11 @@ fn greeks_for(option: &Options) -> Result<Greek, GreeksError> {
 ///     Err(e) => error!("Error calculating delta: {:?}", e),
 /// }
 /// ```
+///
+/// # Sign convention
+///
+/// Returns the sensitivity of the **position**: signed by [`Side`] and scaled
```
joaquinbejar, line 779: 🔴 **Critical**: The non-European early returns below bypass `signed_quantity`. `numerical_delta` and `numerical_gamma` differentiate `Positive` prices produced after `abs()` and never apply `Side` or `quantity`, so a supported exotic short reports the same exposures as its long equivalent (and quantities above one are under-scaled). Ensure the numerical results receive the same signed-quantity scaling and add an exotic long/short regression before documenting these APIs as position-signed.

joaquinbejar (commented):
REQUEST_CHANGES. The European analytical paths now apply `Side` consistently and the strategy coverage is strong. However, the non-European numerical delta/gamma paths still erase both side and quantity, so the PR does not yet satisfy the advertised position-signed contract for supported exotic options.

src/greeks/equations.rs
joaquinbejar: You were right and it is fixed. The signed quantity scaling landed in a8a8117f two minutes after you posted, and 146523b3 adds the exotic regression you asked for, a cash or nothing binary asserting that three lots long is exactly 3x the one lot baseline and three lots short is exactly -3x. My earlier branch matrix could not catch this because it used American, which errors in both paths and so asserted nothing.

joaquinbejar:
All addressed. The non-European paths are signed and scaled now, with an exotic long and short regression pinning it. An architecture pass on the same branch also caught that PortfolioGreeks::from_positions was re-applying quantity and sign, which cancelled the sign once the greeks became signed, so that is fixed with three tests it never had. Garman Kohlhagen and Black 76 are deliberately out of scope and tracked in #436.

PR #437, opened 2026-08-28
Fix Black-76 and GK position Greek signs (#436)

### LINKED ISSUE #436: Migrate Garman-Kohlhagen and Black-76 greeks to the position sign convention
## Overview

#428 made every greek in `src/greeks/equations.rs` return the sensitivity of the
**position**, signed by `Side` and scaled by `quantity`. The Garman-Kohlhagen
and Black-76 families were not migrated, so the crate now ships two
contradictory conventions for the same concept.

Created from the review of PR #435.

## Evidence

In both modules only `delta` reads `Side`, through a local `side_sign` helper
that already exists and is already used there:

| Module | Signs by `Side` | Scales by `quantity` only |
|--------|-----------------|---------------------------|
| `src/greeks/garman_kohlhagen.rs` | `delta_gk` (`:198`) | `gamma_gk` (`:246`), `vega_gk` (`:281`), `theta_gk` (`:354`), `rho_domestic_gk` (`:399`), `rho_foreign_gk` (`:442`) |
| `src/greeks/black_76.rs` | `delta_b76` (`:102`) | `gamma_b76` (`:151`), `vega_b76` (`:193`), `theta_b76` (`:269`), `rho_b76` (`:337`) |

`src/greeks/garman_kohlhagen.rs:215` even documents the old behaviour
explicitly: *"identical for calls and puts, independent of `Side`"*. That line
is now wrong relative to the rest of the crate.

This is exactly the defect #428 fixed, in a different family: an FX or futures
position with a short leg reports gamma, vega, theta and rho with the sign of
the equivalent long.

## The latent trap

`test_matches_bsm_greeks_when_q_is_zero` (`src/greeks/garman_kohlhagen.rs:662`)
asserts `gamma_gk == gamma` **bit-exactly**, and it passes today only because
its fixture is `Side::Long`. Flip that fixture to `Side::Short` and it fails
right now, because the two families disagree. That test is the clearest proof
the conventions have diverged, and it should be extended to cover both sides as
part of this work.

## Tasks

- [ ] Apply `side_sign` in `gamma_gk`, `vega_gk`, `theta_gk`, `rho_domestic_gk`
      and `rho_foreign_gk`, where they currently apply `quantity` alone
- [ ] Same for `gamma_b76`, `vega_b76`, `theta_b76` and `rho_b76`
- [ ] Check for the double-apply shape before committing: any function built on
      top of another that already carries the sign must not apply it again. That
      is the bug fixed in #429 for `quantity` and re-found in #435 at
      `PortfolioGreeks::from_positions`
- [ ] Correct the doc at `garman_kohlhagen.rs:215` and state the convention in
      every migrated function's doc comment, matching what `equations.rs` now
      does
- [ ] Extend `test_matches_bsm_greeks_when_q_is_zero` to run both sides, so the
      two families are pinned as consistent rather than accidentally agreeing
- [ ] Add a test per family asserting that flipping `Side` negates every greek
- [ ] Update any pinned expectations that involve a short position, deriving the
      new values rather than refitting them
- [ ] Note the breaking behaviour change in `CHANGELOG.md`

## Acceptance criteria

- [ ] Every greek in both families returns the position sensitivity, signed by
      `Side` and scaled by `quantity`
- [ ] `gamma_gk` equals `gamma` for a `Side::Short` option with zero dividend
      yield, not only for a long one
- [ ] No function applies the sign twice along any path

## Related

- #428 established the convention for the twelve greeks in `equations.rs`.
- #429 is the precedent for the double-apply failure mode this must avoid.

joaquinbejar:
## Summary

Black-76 and Garman-Kohlhagen exposed the same position Greeks as the core Black-Scholes family but only signed delta by `Side`. This change makes every Greek in both model families return the sensitivity of the full position, consistently signed by `Side` and scaled by `quantity`.

## Changes

- Apply the long/short sign exactly once to Black-76 gamma, vega, theta, and rho.
- Apply the long/short sign exactly once to Garman-Kohlhagen gamma, vega, theta, domestic rho, and foreign rho.
- Document the position-sign convention on every migrated function and in both module overviews.
- Extend long/short antisymmetry and quantity-linearity coverage to every Greek in both families.
- Verify Garman-Kohlhagen and Black-Scholes delta, gamma, and vega agree for both sides when the foreign rate is zero.
- Record the breaking behavioral change and migration guidance in the changelog.

## Testing

- [x] Reference regression test updated for both position sides
- [x] `cargo test --all-features --workspace` — 4,539 passed; 0 failed
- [x] `cargo clippy --all-targets --all-features --workspace -- -D warnings`
- [x] `cargo fmt --all --check`
- [x] `cargo build --release`
- [x] `cargo doc --all-features --no-deps`

## Checklist

- [x] All affected public items document the position-sign convention and retain their `# Errors` sections
- [x] No `.unwrap()` / `.expect()` / unchecked indexing added to production code
- [x] Checked arithmetic used for the new sign application
- [x] `rust_decimal::Decimal` remains the public numeric boundary
- [x] Module boundaries remain unchanged
- [x] No dependencies or feature flags added
- [x] `tracing` remains the only production logging API

Closes #436

### DIFF
--- a/src/greeks/garman_kohlhagen.rs
+++ b/src/greeks/garman_kohlhagen.rs
@@ -26,12 +26,16 @@
 //!
 //! Mirrors the BSM module:
 //!
-//! * `delta_gk` — per unit move in `S`, applies long/short sign.
+//! * `delta_gk` — per unit move in `S`.
 //! * `gamma_gk` — per unit move in `S` (second order).
 //! * `vega_gk`  — per **1%** change in volatility.
 //! * `theta_gk` — per **calendar day** (annual figure ÷ 365).
 //! * `rho_domestic_gk` — per **1%** change in `r_d`.
 //! * `rho_foreign_gk`  — per **1%** change in `r_f`.
+//!
+//! Every value is the sensitivity of the position: [`Side::Long`] keeps the
+//! analytic long-option sign, [`Side::Short`] negates it, and
+//! `option.quantity` scales it linearly.
 //!
 //! ## Two rhos
 //!
@@ -81,6 +85,7 @@ fn ensure_european(option: &Options) -> Result<(), GreeksError> {
     }
 }
 
+#[inline]
 fn side_sign(option: &Options) -> Decimal {
     if matches!(option.side, Side::Long) {
         Decimal::ONE
@@ -212,8 +217,14 @@ pub fn delta_gk(option: &Options) -> Result<Decimal, GreeksError> {
 ///
 /// # Formula
 ///
-/// `Γ = e^(-r_f·T) · n(d1) / (S · σ · √T)` — identical for calls and puts,
-/// independent of `Side`. Result is multiplied by `option.quantity`.
+/// `Γ = e^(-r_f·T) · n(d1) / (S · σ · √T)` — identical for long calls and
+/// long puts.
+///
+/// # Sign convention
+///
+/// Returns the sensitivity of the **position**: signed by [`Side`] and scaled
+/// by `option.quantity`. A short position reports the negative of the
+/// equivalent long.
 ///
 /// # Errors
 ///
@@ -243,7 +254,12 @@ pub fn gamma_gk(option: &Options) -> Result<Decimal, GreeksError> {
     let numer = d_mul(exp_neg_rf_t, n(d1_v)?, "greeks::gk::gamma::numer")?;
     let raw = d_div(numer, denom, "greeks::gk::gamma::raw")?;
 
-    let result = d_mul(raw, option.quantity.to_dec(), "greeks::gk::gamma::quantity")?;
+    let signed = d_mul(side_sign(option), raw, "greeks::gk::gamma::sign")?;
+    let result = d_mul(
+        signed,
+        option.quantity.to_dec(),
+        "greeks::gk::gamma::quantity",
+    )?;
     Ok(result)
 }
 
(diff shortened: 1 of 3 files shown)


src/greeks/garman_kohlhagen.rs
```diff
@@ -660,19 +724,33 @@ mod tests {
...
+
+        for side in [Side::Long, Side::Short] {
+            let mut option = create_fx_option(
+                100.0,
+                100.0,
+                dec!(0.05),
+                0.0,
+                180.0,
+                0.20,
+                OptionStyle::Call,
```
joaquinbejar, line 741: 💡 **Suggestion**: The parity loop now covers both sides, which is what the issue asked for, but it stays on Call. A put would be the more interesting case since the two families reach it differently, GK through `-e^(-r_f T)·N(-d1)` and BSM through `sign·(N(d1) - 1)·e^(-qT)`, so a divergence would show up there first.

Worth checking whether it holds under `assert_eq!` though. Those two forms are equal in exact arithmetic but `big_n` goes through f64, so `N(-d1)` and `1 - N(d1)` can differ in the last place and the put case may need a tolerance rather than exact equality.

src/greeks/garman_kohlhagen.rs
```diff
@@ -81,6 +85,7 @@ fn ensure_european(option: &Options) -> Result<(), GreeksError> {
     }
 }
 
+#[inline]
 fn side_sign(option: &Options) -> Decimal {
```
joaquinbejar, line 89: 🔵 **Nitpick**: `side_sign` picked up `#[inline]` here and in `black_76.rs:61` but not `#[must_use]`. Its counterpart `signed_quantity` in `greeks::equations` got both a couple of commits ago, so adding it keeps the three helpers consistent.

joaquinbejar (commented):
Clean migration, and the part I most wanted to see is right. The sign lands exactly once in each function, through the checked d_mul before the quantity multiplication, and nothing in either family is built on top of another greek, so there is no room for the double-apply that bit us in #429 and again at PortfolioGreeks in #435. I checked the aggregating traits too and they are pure delegation, and the only consumers outside these modules are two examples that use Side::Long, so nothing downstream was compensating for the old behaviour. I ran the suite, clippy and fmt myself and they are green, with the extended parity test passing for Short, which is exactly the assertion that would have failed before. Two small things below, neither blocking. COMMENT.

Issue #439, opened 2026-08-28
apply_spread withdraws a quote instead of widening it


joaquinbejar:
## Overview

`OptionData::apply_spread` (`src/chains/optiondata.rs:924`) **withdraws a quote**
rather than widening it. When `call_middle <= spread` it sets `call_ask`,
`call_bid` **and** `call_middle` to `None`; the put branch at `:960` is
identical.

So every contract worth less than one spread width disappears from the chain
entirely, quotes and mid alike. With a spread of `0.10`, nothing priced under ten
cents exists.

Two things are wrong with that.

**It is unrealistic.** A market maker does not stop quoting a five-cent option.
They quote it three at seven. Withdrawing the quote models a market that has no
counterparty at any price, which is not what a thin option is.

**The comparison is against the full spread, not the half.** The erasure branch
tests `call_middle > spread` while the quoting branch subtracts
`half_spread = spread / 2`. A contract with a mid of `0.06` and a spread of
`0.10` would quote `0.01 / 0.11` under the half-spread arithmetic, with
`sub_floor_zero` already there to keep the bid non-negative. It is erased
instead. The threshold is twice as aggressive as the arithmetic below it
requires.

## Why this matters now

It has become load-bearing for a downstream consumer.

Chainwalk (`gitlab.taunais.com/capitaldelta/chainwalk`) values every option
position at the touch: buying at the ask, selling at the bid, and marking an open
position at what it could be closed for. Under that convention the contracts that
fall below the erasure threshold are exactly the cheap out-of-the-money wings,
which are the long legs of every defined-risk structure. They vanish precisely as
they decay, which is precisely when the position needs to be marked and closed.

The same applies to OptionChain-Simulator, which drives `apply_spread` from its
chain builder (joaquinbejar/OptionChain-Simulator#80).

`sub_floor_zero` is already used on the bid side, so the intent to floor rather
than to erase is present in the code; it just is not reached.

## Proposal

Widen and floor instead of withdrawing.

- Quote at `mid ± half_spread` as today
- Floor the bid at a tick rather than allowing it to go negative;
  `sub_floor_zero` already does the non-negative half of this
- Never set `call_middle` / `put_middle` to `None` in this function: a mid that
  arrived is a mid, and clearing it destroys information the caller supplied
- Erase only when there was genuinely nothing to quote, which is the existing
  third match arm where both bid and ask are absent and there is no mid

The one judgement call is the bid floor for a contract whose mid is below
[...]

PR #441, opened 2026-08-28
Eliminate reachable panics in the Greeks and pricing core (#440)

### LINKED ISSUE #440: Eliminate reachable panics in the Greeks and pricing core
## Overview

The library aborts the caller's process for inputs that pass every validation
gate it has. A probe over the public Greeks and Black-Scholes surface — each
call wrapped in `catch_unwind`, driven by extreme but structurally valid
`Options` values — produced **119 distinct panics** across 12 Greeks and the
pricing entry points. `OptionStratLib` is embedded in long-running services
(`OptionChain-Simulator`, `IronCondor`), where a panic kills a tokio worker and
takes the in-flight request with it, so every one of these is a remote abort
reachable from a request body.

The root cause is systemic rather than local: `rust_decimal` and the `positive`
crate panic by default on the *operator* path and only offer safety on the
*checked* path. `Decimal`'s `+ - * /` panic on overflow, `/` panics on division
by zero, `exp()` panics with `Exp overflowed` / `Exp underflowed`, `ln()` panics
on zero and on negatives, `powd()` panics with `Pow overflowed`. Every
`Positive` operator (including its combinations with `Decimal` and `f64`) panics
on overflow, on division by zero, and when the result would break the positivity
invariant. `Positive` is `>= 0`, not `> 0`, so a `Positive` divisor never proves
a division is safe.

## Context

Reproduced panics, with the input that triggers each (probe messages quoted
verbatim):

| Input | Panic | Reached from |
|---|---|---|
| `iv = 1e-13` | `Positive arithmetic overflow in div_decimal_by_positive` | `vomma`, `greeks()` |
| `iv = 1e-27` | `Pow overflowed` | 9 Greeks + `greeks()` |
| `iv = Positive::MAX` | `Positive arithmetic overflow in powd` | all 12 Greeks, `black_scholes`, `time_value` |
| `T = 1e9 days` / `T = MAX` | `Exp underflowed` | `theta`, `rho`, `alpha`, `greeks()`, `black_scholes` |
| `T = MAX` | `Positive arithmetic overflow in mul_decimal_by_positive` | `charm`, `color` |
| `S = 1e-27` | `Unable to calculate ln for zero` | every Greek and every pricing entry point |
| `S = MAX` | `Positive arithmetic overflow in mul_decimal_by_positive` | `color`, `greeks()` |
| `K = 1e-27` | `Positive arithmetic overflow in div_decimal_by_positive` | every Greek and every pricing entry point |
| `r = ±1e6` | `Exp overflowed` / `Exp underflowed` | `theta`, `rho`, `alpha`, `black_scholes` |

`S = 1e-27` is the sharpest case: `S / K` rounds below the smallest
representable `Decimal`, and `ln(0)` panics. Nothing in `d1`'s validation
catches it — the function checks each input for zero individually, and each one
is non-zero.

The `vomma` path is the one already reported downstream: `vomma = vega · d1 · d2
/ sigma`, so a volatility approaching zero pushes the quotient out of `Decimal`'s
range and `Positive`'s `Div` panics. It has been reachable through `?greeks=`
since the greek query parameter shipped.

Relevant code:

- `src/greeks/utils.rs:87` — `d1`: `powd`, `Positive` division, `ln`, and three
  unchecked operators.
- `src/greeks/utils.rs:300` — `n`: `x.powd(2)` before the magnitude guard.
- `src/greeks/equations.rs:491` — `BlackScholesKernels`: `sqrt`, `d2`, and both
  discount factors use panicking operators.
- `src/greeks/equations.rs` — `vomma_with`, `vanna_with`, `veta_with`,
  `charm_with`, `color_with`, `gamma_with`, `theta_with`, `vega_with`,
  `alpha_from` divide by volatility, by time, or by the underlying with the raw
  operator.
- `src/model/decimal.rs:311` — the existing `d_add` / `d_sub` / `d_mul` /
  `d_div` checked helpers, which the fix extends and reuses.

Wider inventory across the 50,294 lines of production code (tests stripped):
611 divisions, 119 `.exp()`, 99 `.sqrt()`, 62 `.powd()`, 25 `.ln()`, plus 14
`unreachable!` and 5 `panic!`. There is no `.unwrap()` or `.expect()` in
production code — the panic surface here is arithmetic, not error handling.

## Tasks

- [ ] Extend `src/model/decimal.rs` with checked `d_exp`, `d_ln`, `d_powd`,
      `d_sqrt`, matching the existing `d_*` tagging convention
- [ ] Make `d1`, `d2` and `n` panic-free, including the `S / K` underflow case
- [ ] Make `BlackScholesKernels` panic-free: `sqrt(T)`, `d2`, `exp(-qT)`,
      `exp(-rT)`
- [ ] Convert all twelve Greeks to checked arithmetic
- [ ] Convert the Black-Scholes, Black-76 and Garman-Kohlhagen pricing and
      Greek kernels
- [ ] Sweep `src/chains/` and `src/volatility/` — the chain build path and the
      IV solvers are the next reachable surface
- [ ] Add a proptest that hammers the extreme domain and asserts every public
      Greek and pricing entry point returns, never aborts
- [ ] Add a `scan-banned` style guard so a banned construct reappearing in
      production code fails CI (same convention as the `positive` crate)

## Acceptance criteria

- [ ] The `catch_unwind` probe over the Greeks and pricing surface reports zero
      panics for every input in the table above
- [ ] Numerical results are unchanged for well-formed inputs — the existing unit
      tests pass untouched
- [ ] Failures surface as typed errors (`GreeksError`, `DecimalError`,
      `PricingError`), never as an abort, and each new fallible function
      documents its `# Errors`
- [ ] No `saturating_*` / `wrapping_*` introduced in financial math

## References

- `rules/global_rules.md` — checked arithmetic and error-handling rules
- `rust_decimal` 1.40 `src/maths.rs` — the `panic!("Exp overflowed")`,
  `panic!("Pow overflowed")` and `panic!("Unable to calculate ln for zero")`
  sites
- `positive` 0.6 `src/positive.rs` — `guard_nonzero_divisor` / `unwrap_or_panic`
  on every operator, and the `checked_*` counterparts

joaquinbejar:
## Summary

The Greeks and Black-Scholes surface aborted the calling process for inputs
that passed every validation gate the library has. A `catch_unwind` probe over
the public API produced 119 panics across the twelve Greeks and the pricing
entry points; the library runs inside long-running services, where each of
those kills a tokio worker and the in-flight request with it. This makes that
whole surface panic-free by moving it onto checked arithmetic, and adds a
property test plus a CI scan so it stays that way.

The reported `vomma` abort is one instance of the pattern: `vomma = vega · d1 ·
d2 / sigma`, so a volatility approaching zero pushed the quotient out of
`Decimal`'s range and `Positive`'s `Div` panicked. It was never a `vomma` bug —
`rust_decimal` and `positive` panic by default on the operator path, and the
kernels used operators throughout.

## Changes

- Four new checked helpers in `src/model/decimal.rs`, alongside the existing
  `d_add` / `d_sub` / `d_mul` / `d_div`: `d_exp`, `d_ln`, `d_powd`, `d_sqrt`.
  `d_exp` flushes an underflowing discount factor to zero — that is the value
  at the limit and the only representable one — and reports a genuine overflow
  as an error.
- `d1`, `d2` and `n` are panic-free, including the case that had no guard at
  all: `S / K` rounding below the smallest representable `Decimal`, where
  `ln(0)` panicked. Each input was checked for zero individually, and each one
  was non-zero.
- `BlackScholesKernels` computes `sqrt(T)`, `d2`, `exp(-qT)` and `exp(-rT)` on
  the checked path; `d2` and the two discount factors are now fallible and
  stay memoised in their `OnceCell`s.
- All twelve Greeks (`delta`, `gamma`, `theta`, `vega`, `rho`, `rho_d`,
  `alpha`, `vanna`, `vomma`, `veta`, `charm`, `color`) use checked arithmetic
  end to end, each operation tagged with its call site.
- Black-Scholes, Black-76 and Garman-Kohlhagen pricing and Greeks: every
  discount factor and every `sqrt(T)` goes through the checked helpers.
- `Options::time_value` no longer panics when the price sits near the top of
  the range against a negative intrinsic value.
- `calculate_d_values` computes the cost of carry `b = r - q` with a checked
  subtraction; `Decimal::MIN` as a rate reached the raw operator.
- Chain building: `adjust_volatility` computes log-moneyness on the checked
  path and degrades to a flat smile when it is not representable (matching the
  function's existing fallbacks), the strike grid uses checked `Positive`
  arithmetic, and a relative expiry that no calendar date can represent is
  rejected up front instead of reaching `DateTime + TimeDelta`, which panics.
- `tests/property/panic_freedom_test.rs`: proptest over the extremes of the
  `Decimal` domain — volatilities and strikes at the smallest representable
  scale, prices and times to expiry at the largest, rates far outside anything
  quotable — asserting the Greeks, the three pricing models and the chain
  builder return rather than abort. It found two panics this PR would
  otherwise have missed (`Decimal::MIN` rate, `time_value` overflow).
- `make scan-banned` (same convention as the `positive` crate, with
  `// scan-banned: allow -- <reason>` escapes), wired into the lint workflow
  and into `make check`.

## Technical decisions

- **Errors, not clamps.** Where a result is genuinely not representable the
  call returns `Err`; no silent saturation, no invented number. The two
  exceptions are mathematically defined limits: `e^x` underflowing to zero,
  and `ln(S/K)` for a moneyness that rounded to zero, which keeps the existing
  `Decimal::MIN` sentinel the `Positive::MAX` strike branch already used.
- **Numerics unchanged for well-formed inputs.** `d_div` rounds at
  `DIV_DEFAULT_SCALE = 28`, which is `Decimal`'s maximum scale, so replacing
  `/` with `d_div` is value-preserving. The existing unit tests pass
  unmodified; the only test edits are the three assertions in
  `tests_black_scholes_kernels` that had to follow `d2` / `exp_minus_qt` /
  `exp_minus_rt` becoming fallible.
- **Probe over inspection.** Every panic listed in #440 was reproduced before
  being fixed and re-checked afterwards, rather than argued from the source.

## Public API impact

`d1`, `d2`, `n` and the twelve Greek functions keep their signatures — they
were already fallible. `alpha_from` (crate-private) became fallible. No public
item was added, removed or renamed, so `README.tpl` needs no update.

## Testing

- [x] Integration tests added or updated (property test for panic freedom)
- [x] Reference regression test added (for pricing / Greek changes) — existing
      reference and put-call-parity tests pass unchanged, which is the
      regression bar for this PR
- [ ] Criterion benchmark added/updated (for perf changes)
- [x] `cargo test --all-features --workspace` clean
- [x] `make scan-banned` clean

The probe that produced the panic list is a throwaway crate with a path
dependency on this repo; the property test is the committed, repeatable form
of it.

## Checklist

- [x] Module boundaries respected (`model/` has no deps on `strategies/`, `backtesting/`, `visualization/`)
- [x] `tracing` only for logging — no `println!` / `eprintln!` / `dbg!` / `log` crate
- [x] `README.tpl` updated (and README regenerated via `make readme`) if public surface changed — not applicable, public surface unchanged

## Not in this PR

The sweep stops at the surface the probes reach. Still on raw operators, and
worth their own issues: `src/pricing/` exotics (asian, barrier, lookback,
compound, american, cliquet, chooser, rainbow, spread, quanto, power,
exchange), `src/curves/`, `src/surfaces/`, `src/strategies/`, and
`src/simulation/`. The current inventory of production code is 595 divisions,
102 `.exp()`, 91 `.sqrt()`, 60 `.powd()`, 23 `.ln()`, 14 `unreachable!` and 5
`panic!`. Two further items observed while working and deliberately left
alone: `Options::intrinsic_value` swallows a non-finite payoff through
`Decimal::from_f64(..).unwrap_or_default()`, returning zero instead of an
error, and `expiration_date` 0.3.0 panics on calendar overflow inside
`get_date_with_options`, which is guarded here at the call site but belongs
upstream.

Closes #440

---

**Blocks:** #443 (issue #439), which is stacked on this branch. Merge this one first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,13 +61,61 @@ clean:
 .PHONY: check
 check: test fmt-check lint scan-banned
 
-# Fails when a panicking construct reappears in production code (test modules
-# and doc comments are skipped). A reviewed exception carries a trailing
+# Fails when a panicking construct reappears in production code. A bare
+# `#[cfg(test)]` never truncates the scan; only the braced body of the item
+# it actually gates is skipped, brace counted (files may carry several):
+# `mod`/`fn`/`impl`/`trait`/`struct`/`enum`/`union` (any `pub(..)` visibility,
+# `async`/`unsafe`, generics or return type before the `{`, spanning multiple
+# signature lines if needed), and a bare `#[cfg(test)] { ... }` block. A
+# body-less item behind `#[cfg(test)]` (`use`, `const`, `static`, `type`
+# alias, or a `;`-terminated declaration such as a tuple struct or a trait
+# method signature) skips nothing; scanning resumes right after it as usual.
+# Doc comments are skipped too. A reviewed exception carries a trailing
 # `// scan-banned: allow -- <reason>` marker on the same line.
 .PHONY: scan-banned
 scan-banned:
 	@found=$$(for f in $$(find src -name '*.rs'); do \
-		awk -v file="$$f" '/#\[cfg\(test\)\]/ { exit } { print file ":" NR ":" $$0 }' "$$f"; \
+		awk -v file="$$f" ' \
+			BEGIN { skip = 0; depth = 0; pending = 0; awaiting = 0 } \
+			function braces(line,   tmp, o, c) { \
+				tmp = line; o = gsub(/{/, "{", tmp); \
+				tmp = line; c = gsub(/}/, "}", tmp); \
+				return o - c; \
+			} \
+			{ \
+				raw = $$0; \
+				if (skip) { \
+					depth += braces(raw); \
+					if (depth <= 0) { skip = 0; depth = 0 } \
+					next; \
+				} \
+				if (awaiting) { \
+					depth += braces(raw); \
+					if (depth > 0) { awaiting = 0; skip = 1; next } \
+					if (raw ~ /;[[:space:]]*$$/) { awaiting = 0; depth = 0 } \
+					next; \
+				} \
+				if (pending) { \
+					if (raw ~ /^[[:space:]]*#\[/) { next } \
+					pending = 0; \
+					trimmed = raw; sub(/^[[:space:]]+/, "", trimmed); \
+					if (trimmed ~ /^\{/) { \
+						depth = braces(raw); \
+						if (depth > 0) { skip = 1 } \
+						next; \
+					} \
+					if (trimmed ~ /^(pub([[:space:]]*\([^)]*\))?[[:space:]]+)?(async[[:space:]]+)?(unsafe[[:space:]]+)?(mod|fn|impl|trait|struct|enum|union)([[:space:]<(]|$$)/) { \
+						depth = braces(raw); \
+						if (depth > 0) { skip = 1 } \
+						else if (raw ~ /;[[:space:]]*$$/) { depth = 0 } \
+						else { awaiting = 1; depth = 0 } \
+						next; \
+					} \
+				} \
+				if (raw ~ /^[[:space:]]*#\[cfg\(test\)\][[:space:]]*$$/) { pending = 1; next } \
+				print file ":" NR ":" raw; \
+			} \
+		' "$$f"; \
 	done \
 		| grep -E '\.unwrap\(\)|\.expect\(' \
 		| grep -v -E ':[0-9]+:[[:space:]]*(///|//!|//|\*)' \
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,25 @@ clean:
 
 # Pre-push checks
 .PHONY: check
-check: test fmt-check lint
+check: test fmt-check lint scan-banned
+
+# Fails when a panicking construct reappears in production code (test modules
+# and doc comments are skipped). A reviewed exception carries a trailing
+# `// scan-banned: allow -- <reason>` marker on the same line.
+.PHONY: scan-banned
+scan-banned:
+	@found=$$(for f in $$(find src -name '*.rs'); do \
+		awk -v file="$$f" '/#\[cfg\(test\)\]/ { exit } { print file ":" NR ":" $$0 }' "$$f"; \
+	done \
+		| grep -E '\.unwrap\(\)|\.expect\(' \
+		| grep -v -E ':[0-9]+:[[:space:]]*(///|//!|//|\*)' \
+		| grep -v 'scan-banned: allow' || true); \
+	if [ -n "$$found" ]; then \
+		echo "Banned patterns found in production code:"; \
+		echo "$$found"; \
+		exit 1; \
+	fi; \
+	echo "OK: no .unwrap()/.expect() in production code"
 
 # Run the project
 .PHONY: run
--- a/src/greeks/black_76.rs
+++ b/src/greeks/black_76.rs
@@ -34,9 +34,11 @@ use crate::Options;
 use crate::error::PricingError;
 use crate::error::greeks::GreeksError;
 use crate::greeks::utils::{big_n, calculate_d_values_black_76, n};
-use crate::model::decimal::{d_div, d_mul, d_sub};
+use crate::model::decimal::{d_div, d_exp, d_mul, d_sub};
 use crate::model::types::{OptionStyle, OptionType, Side};
-use rust_decimal::{Decimal, MathematicalOps};
+use rust_decimal::Decimal;
+#[cfg(test)]
+use rust_decimal::MathematicalOps;
 use tracing::{instrument, trace};
 
 /// Reject any option type that is not European; Black-76 only prices European
@@ -66,8 +68,14 @@ fn side_sign(option: &Options) -> Decimal {
     }
 }
 
-fn discount_factor(option: &Options, t: Decimal) -> Decimal {
-    (-option.risk_free_rate * t).exp()
+/// `e^(-rT)`, shared by every Black-76 greek.
+///
+/// Fallible because `Decimal::exp` panics on overflow and on underflow; a
+/// discount factor below the representable scale flushes to zero, which is
+/// its limit.
+fn discount_factor(option: &Options, t: Decimal) -> Result<Decimal, GreeksError> {
+    let exponent = d_mul(-option.risk_free_rate, t, "greeks::b76::discount::exponent")?;
+    Ok(d_exp(exponent, "greeks::b76::discount")?)
 }
 
 /// Computes the delta of an option under the Black-76 model.
@@ -95,7 +103,7 @@ pub fn delta_b76(option: &Options) -> Result<Decimal, GreeksError> {
     let t = option.expiration_date.get_years()?.to_dec();
     let (d1, _d2) = calculate_d_values_black_76(option)?;
 
-    let df = discount_factor(option, t);
+    let df = discount_factor(option, t)?;
     let raw = match option.option_style {
         OptionStyle::Call => d_mul(df, big_n(d1)?, "greeks::black_76::delta::call")?,
         OptionStyle::Put => -d_mul(df, big_n(-d1)?, "greeks::black_76::delta::put")?,
@@ -140,10 +148,10 @@ pub fn gamma_b76(option: &Options) -> Result<Decimal, GreeksError> {
     let t = option.expiration_date.get_years()?;
     let (d1, _d2) = calculate_d_values_black_76(option)?;
 
-    let df = discount_factor(option, t.to_dec());
+    let df = discount_factor(option, t.to_dec())?;
     let f = option.underlying_price.to_dec();
     let sigma = option.implied_volatility.to_dec();
-    let sqrt_t = t.sqrt().to_dec();
+    let sqrt_t = t.checked_sqrt()?.to_dec();
 
     let denom = d_mul(
         f,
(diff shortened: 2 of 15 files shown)


src/greeks/garman_kohlhagen.rs
```diff
@@ -194,7 +196,10 @@ pub fn delta_gk(option: &Options) -> Result<Decimal, GreeksError> {
     let (d1_v, _d2) = calculate_d_values_gk(option)?;
 
     let r_f = option.dividend_yield.to_dec();
-    let exp_neg_rf_t = (-r_f * t).exp();
+    let exp_neg_rf_t = d_exp(
+        d_mul(-r_f, t, "greeks::gk::discount_foreign::exponent")?,
```
joaquinbejar, line 212: 🔴 **Critical**: This discount path is checked, but `calculate_d_values_gk` immediately above still reaches `cost_of_carry`, where `risk_free_rate - dividend_yield` uses raw `Decimal` subtraction. `delta_gk` will panic with `risk_free_rate = Decimal::MIN` and `dividend_yield = 1`, so make `cost_of_carry` fallible via `d_sub` and directly test the dedicated GK Greeks.

src/greeks/black_76.rs
```diff
@@ -246,12 +254,12 @@ pub fn theta_b76(option: &Options) -> Result<Decimal, GreeksError> {
     let t = option.expiration_date.get_years()?;
     let (d1, d2) = calculate_d_values_black_76(option)?;
 
-    let df = discount_factor(option, t.to_dec());
+    let df = discount_factor(option, t.to_dec())?;
```
joaquinbejar, line 257: 🔴 **Critical**: The discount factor is checked now, but `theta_b76` still evaluates `common + plus` with raw `Decimal` addition in both branches before calling `d_sub`. With `F = K = 0.98 * Decimal::MAX`, `r = -0.9`, `T = 3.65 days`, and `sigma = 0.3`, this will panic with `Addition overflowed`; compute the intermediate through `d_add(...)?` in both branches.

Makefile
```diff
@@ -59,7 +59,25 @@ clean:
...
-check: test fmt-check lint
+check: test fmt-check lint scan-banned
+
+# Fails when a panicking construct reappears in production code (test modules
+# and doc comments are skipped). A reviewed exception carries a trailing
+# `// scan-banned: allow -- <reason>` marker on the same line.
+.PHONY: scan-banned
+scan-banned:
+	@found=$$(for f in $$(find src -name '*.rs'); do \
+		awk -v file="$$f" '/#\[cfg\(test\)\]/ { exit } { print file ":" NR ":" $$0 }' "$$f"; \
```
joaquinbejar: 🟠 **Warning**: This exits at the first `#[cfg(test)]` anywhere in a source file, not only at the test module. Files such as `src/pricing/black_76.rs`, `src/greeks/black_76.rs`, and `src/greeks/garman_kohlhagen.rs` now have test-only imports near the top, so almost all later production code is skipped; a temporary production `.unwrap()` after such an import still makes this target print `OK`. Restrict the cutoff to the actual test module or use syntax-aware scanning.

joaquinbejar (commented):
The checked-arithmetic migration is incomplete in two reachable Greek paths, and the new guardrail can silently skip most production code, so the green checks do not establish the PR's panic-freedom claim. REQUEST_CHANGES. GitHub does not allow the PR author to set the formal request-changes state.

src/greeks/garman_kohlhagen.rs
joaquinbejar: Fixed in da86f880. `cost_of_carry` now goes through `d_sub` and returns a `Result`, so `calculate_d_values_gk` propagates the overflow instead of aborting. Added `tests_gk_extreme_rates`, which drives all six dedicated GK greeks with `risk_free_rate = Decimal::MIN` and `dividend_yield = 1` and asserts every one returns an error; I reverted the fix to confirm the test catches the exact `Subtraction overflowed` you predicted, and a second test pins that ordinary rates still price.

src/greeks/black_76.rs
joaquinbejar: Fixed in da86f880. Both branches now build the intermediate with `d_add(common, plus, ...)?` before the checked subtraction. Added `tests_b76_extreme_inputs` with your reproduction — forward and strike at 98% of `Decimal::MAX`, `r = -0.9`, 3.65 days, sigma 0.3 — and it fails with `Addition overflowed` on the old code, passes as an error now.

Makefile
joaquinbejar: Fixed in da86f880, and the blind spot was worse than the three files. The target scanned 68,949 of 179,040 lines because it exited at the first `#[cfg(test)]` anywhere; it now skips only the body of a `#[cfg(test)]` item — module, function, impl, or bare block, by brace counting — and resumes after it, covering 87,701 lines. Proof it works where it was blind: an injected `.unwrap()` placed right after that test-only import is now reported at its exact line, and the target stays clean once removed.

joaquinbejar:
All three are fixed in da86f880 and the review is right that the green checks were not establishing the claim. The two carry paths now go through `d_sub` and `d_add`, each with a regression test that I verified fails with the exact panic you named before the fix. The guardrail was the worst of the three: it was scanning 68,949 of 179,040 lines because it stopped at the first `#[cfg(test)]` anywhere, including the test-only imports this PR itself added, so it now skips only the body of a `#[cfg(test)]` item and covers 87,701 lines, proven against an injected `.unwrap()` placed exactly where it used to be blind. Suite is 3896 lib, 418 integration, 15 property and 207 doc tests green, with clippy, fmt and the scan clean.

PR #443, opened 2026-08-28
Widen a thin quote instead of withdrawing it (#439)

### LINKED ISSUE #439: apply_spread withdraws a quote instead of widening it

joaquinbejar:
## Summary

`OptionData::apply_spread` withdrew a quote instead of widening it: whenever a
contract's mid sat below one full spread it cleared `bid`, `ask` **and**
`middle`. That is not what a thin market is — a market maker does not stop
quoting a five-cent option, they quote it three at seven — and it is
load-bearing downstream, where positions are marked at the touch. Worse, since
`OptionChain::build_chain` stops generating strikes once both wings come back
unpriced, a wide spread also truncated the chain: with a spread of `2.0` a
21-strike chain came out with 3 strikes.

This widens and floors instead. The erasure threshold was also twice as
aggressive as the arithmetic below it required — it compared the mid against
the full spread while the quoting arm subtracted the half.

## Changes

- Both sides are quoted around the anchor and floored at one tick, where the
  tick is `10^-decimal_places`: `ask = max(round(mid + half), tick)` and
  `bid = clamp(round(mid - half), tick, ask)`. The ask is floored too, because
  an anchor small enough that `mid + half` rounds down to zero would otherwise
  produce a bid above the ask.
- `call_middle` / `put_middle` are never cleared when a mid was supplied. A mid
  that arrived is a mid; dropping it destroys information the caller gave us.
- The genuinely-empty case — no bid, no ask, no mid — still erases, and the doc
  comment states the difference between the two cases.
- The same widen-and-floor rule applies to the arm that had a two-sided book
  and no mid, so the invariants hold on both paths.
- Four module-private helpers (`tick_size`, `widen_around`, `widen_book`,
  `mid_of`) replace the call/put copy-paste.
- Panic safety on the lines being rewritten: the `Positive + Decimal` additions
  go through `checked_add_dec`, and `decimal_places` is clamped to
  `Decimal::MAX_SCALE` because `Positive::round_to` panics above it. On
  overflow the quote is left untouched with a `warn!` — `apply_spread` returns
  `()`, so turning it into a `Result` would be a public break.

- A supplied mid is held inside the widened quote instead of being left where
  it was: flooring the bid at a tick would otherwise leave a row whose mid sits
  below its own bid, and `OptionData` is the serialized wire type that feeds
  position premia. In the SP500 chain the 5250 put came out bid `0.01` against
  a mid of order `1e-30`. Clamping keeps the caller's mid wherever it is still
  inside the book, and still never clears it.
- Keeping the wings surfaced a latent panic one layer up:
  `ShortStraddle` / `ShortStrangle::update_break_even_points` computed the
  lower break-even as `strike - credit` with the raw `Positive - Decimal`
  operator, which panics when the credit exceeds the strike. The optimizer
  never built such a combination before because those strikes were being
  erased. The break-even is now floored at zero with the existing
  `sub_floor_zero`, which is where it belongs: the underlying cannot trade
  below zero, so the downside break-even does not exist. Six sibling strategies
  carried the identical expression — `IronCondor`, `IronButterfly`,
  `LongStraddle`, `LongStrangle`, `BearPutSpread`, `ShortButterflySpread`, and
  the iron structures are exactly the defined-risk ones with cheap long wings
  that #439 names — so all eight now share `lower_break_even` in
  `strategies/base.rs`. A lower break-even of `0.00` means the strategy has
  none, which is documented on both the helper and `BreakEvenable`.

## Technical decisions

The one judgement call in the issue is the bid floor for a contract whose mid
is below half a spread. The tick is derived from `decimal_places` rather than
added as a parameter: a two-decimal chain ticks at `0.01`, and the public
signature stays as it is, so `OptionChain::build_chain` and OptionChain-Simulator
keep compiling. The quote is then asymmetric around the mid — a tick bid
against a widened ask — which is exactly what a real thin market looks like.

## Known follow-up

`apply_spread` is applied more than once per row: `OptionData::calculate_prices`
applies it (once per side, and each call widens both sides) and
`build_chain` applies it again at `chain.rs:543`. That was invisible before
because the mid arm re-derives both sides from the stored mid, so repeated
application was idempotent as long as the mid never moved. Holding the mid
inside the quote makes a sub-tick row widen by one further tick on the second
pass — the deep wings settle at `0.01 / 0.02` instead of `0.01 / 0.01` — and
then it is stable. The quotes are coherent either way, but the pipeline should
apply the spread once; that belongs with the chains sweep (442c), not in a
behaviour fix.

`apply_spread` still returns `()`, so an arithmetic failure logs a warning and
leaves the quote untouched rather than surfacing a typed error. The reviewing
architect argued for `-> Result<(), ChainError>`: the error variant already
exists (`ChainError::PositiveError`), the only production caller sits inside a
function that already returns `Result<OptionData, ChainError>`, and it is
cheaper to break the signature at 0.20 than after 1.0. It is not done here
because the approved plan for this stack rules out public signature changes and
OptionChain-Simulator calls this method — that is a decision for the crate
owner, not for a bug fix. With the checked rounding and the mid clamp in place
the remaining failure mode is `checked_add_dec` within half a spread of
`Decimal::MAX`, which `build_chain` cannot reach. Carried into the chains sweep
(442c).

## Public API impact

No signature changes. The behaviour of `apply_spread` changes, which is what
the issue asks for, and the `CHANGELOG.md` entry under `[Unreleased]` records
that downstream row counts change.

## Testing

- [ ] Reference regression test added (for pricing / Greek changes) — not a
      pricing change
- [ ] Criterion benchmark added/updated (for perf changes)
- [x] `make pre-push` run locally and clean

New coverage:

- mid `0.06` against a spread of `0.10` is quoted, not erased: bid at the tick,
  ask above the mid.
- A supplied mid survives the call for every mid from `0.001` to `100`.
- A mid above the spread quotes exactly as before (`9.5 ± 0.3`), so the
  unchanged path is pinned.
- The tick follows `decimal_places` for 0, 1, 2 and 4 decimals.
- `tests/unit/chain/wide_spread_chain_test.rs`: a chain quoted with a spread
  wider than its cheap strikes' mids keeps the same strike count as a tightly
  quoted one, and every strike stays quoted and ordered. On the old code this
  fails 3 vs 7 — it is the regression the downstream consumers hit.
- `tests/property/quote_invariants_test.rs`: over mid × spread ×
  `decimal_places ∈ 0..=6`, `tick <= bid <= ask` and the mid is preserved, on
  both the mid-anchored and the book-anchored path.
- `test_sub_tick_corner_at_and_above_the_old_threshold` pins the boundary the
  issue's "unchanged above the spread" criterion cares about: `mid = 0.05`
  against `spread = 0.04` quotes exactly as before, and the sub-tick case is
  the deliberate deviation.
- Four unit tests on `lower_break_even`, including the credit-above-strike case
  that used to panic.
- The two existing tests that asserted `Some(Positive::ZERO)` for the erased
  case now assert the tick floor. That is the intended behaviour change.

Three existing tests encoded the truncation and were updated to the new
reality, each with a comment saying why:

- `chains::utils::tests_strike_step::long_discrepancy` asserted
  `len() - 1 == chain_size`; a chain that keeps every strike holds
  `2 * chain_size + 1` (57, not 29).
- `chains::chain::tests_chain_base::test_new_option_chain_build_chain_long`
  asserted `None` quotes on both wings; the chain is now 51 strikes with the
  wings quoted at the tick.
- The `short_straddle` / `short_strangle` `test_best_area` and
  `test_best_ratio` tests were panicking through the break-even fix above.

## Checklist

- [x] `tracing` only for logging — no `println!` / `eprintln!` / `dbg!` / `log` crate
- [x] `README.tpl` updated if public surface changed — not applicable

## Stack

- Base branch: `440-eliminate-panics-greeks-pricing` (PR #441)
- Stacked on: #441
- Stack: #441 → **#439 (this)** → #442a → #442b → #442c → #442d → #442e

The diff shown against `main` is cumulative until #441 merges; review it
against the base branch. Merge bottom-up.

Closes #439

### DIFF
--- a/src/strategies/bear_put_spread.rs
+++ b/src/strategies/bear_put_spread.rs
@@ -26,6 +26,7 @@ use super::base::{
     BreakEvenable, Optimizable, Positionable, Strategable, StrategyBasics, StrategyType, Validable,
 };
 use super::shared::SpreadStrategy;
+use crate::strategies::base::lower_break_even;
 use crate::{
     ExpirationDate, Options,
     chains::{StrategyLegs, chain::OptionChain, utils::OptionDataGroup},
@@ -344,9 +345,11 @@ impl BreakEvenable for BearPutSpread {
         self.break_even_points = Vec::new();
 
         self.break_even_points.push(
-            (self.long_put.option.strike_price
-                - self.get_net_cost()? / self.long_put.option.quantity)
-                .round_to(2),
+            lower_break_even(
+                self.long_put.option.strike_price,
+                self.get_net_cost()? / self.long_put.option.quantity,
+            )
+            .round_to(2),
         );
 
         Ok(())
--- a/src/strategies/bear_put_spread.rs
+++ b/src/strategies/bear_put_spread.rs
@@ -27,6 +27,7 @@ use super::base::{
 };
 use super::shared::SpreadStrategy;
 use crate::strategies::base::lower_break_even;
+use crate::strategies::base::price_gap;
 use crate::{
     ExpirationDate, Options,
     chains::{StrategyLegs, chain::OptionChain, utils::OptionDataGroup},
@@ -618,7 +619,10 @@ impl Strategies for BearPutSpread {
     }
     fn get_profit_area(&self) -> Result<Decimal, StrategyError> {
         let high = self.get_max_profit().unwrap_or(Positive::ZERO);
-        let base = self.break_even_points[0] - self.short_put.option.strike_price;
+        let base = price_gap(
+            self.break_even_points[0],
+            self.short_put.option.strike_price,
+        );
         Ok(Decimal::from_f64(high.to_f64() * base.to_f64() / 200.0).unwrap_or(Decimal::ZERO))
     }
     fn get_profit_ratio(&self) -> Result<Decimal, StrategyError> {
--- a/src/strategies/long_strangle.rs
+++ b/src/strategies/long_strangle.rs
@@ -19,6 +19,7 @@ use super::base::{
     BreakEvenable, Optimizable, Positionable, Strategable, StrategyBasics, StrategyType, Validable,
 };
 use super::shared::StrangleStrategy;
+use crate::strategies::base::lower_break_even;
 use crate::{
     ExpirationDate, Options,
     chains::{StrategyLegs, chain::OptionChain, utils::OptionDataGroup},
@@ -380,8 +381,11 @@ impl BreakEvenable for LongStrangle {
         let total_premium = self.get_net_cost()?;
 
         self.break_even_points.push(
-            (self.long_put.option.strike_price - (total_premium / self.long_put.option.quantity))
-                .round_to(2),
+            lower_break_even(
+                self.long_put.option.strike_price,
+                total_premium / self.long_put.option.quantity,
+            )
+            .round_to(2),
         );
 
         self.break_even_points.push(
--- a/src/strategies/long_strangle.rs
+++ b/src/strategies/long_strangle.rs
@@ -20,6 +20,7 @@ use super::base::{
 };
 use super::shared::StrangleStrategy;
 use crate::strategies::base::lower_break_even;
+use crate::strategies::base::price_gap;
 use crate::{
     ExpirationDate, Options,
     chains::{StrategyLegs, chain::OptionChain, utils::OptionDataGroup},
@@ -681,7 +682,9 @@ impl Strategies for LongStrangle {
             "First break even point: {} Last break even point: {}",
             first_option, last_option
         );
-        let start_price = first_option - diff.to_dec();
+        // No lower break-even puts `first_option` at zero, and the plot range
+        // cannot start below it.
+        let start_price = price_gap(first_option, diff);
         debug!("Start price: {}", start_price);
         let end_price = last_option + diff;
         debug!("End price: {}", end_price);
--- a/src/strategies/short_strangle.rs
+++ b/src/strategies/short_strangle.rs
@@ -21,6 +21,7 @@ use super::base::{
 };
 use super::shared::StrangleStrategy;
 use crate::strategies::base::lower_break_even;
+use crate::strategies::base::price_gap;
 use crate::{
     ExpirationDate, Options,
     chains::{StrategyLegs, chain::OptionChain, utils::OptionDataGroup},
@@ -740,7 +741,9 @@ impl Strategies for ShortStrangle {
     fn get_best_range_to_show(&self, step: Positive) -> Result<Vec<Positive>, StrategyError> {
         let max_profit = self.get_max_profit().unwrap_or(Positive::ZERO);
         let (first_option, last_option) = (self.break_even_points[0], self.break_even_points[1]);
-        let start_price = first_option - max_profit.to_dec();
+        // No lower break-even puts `first_option` at zero, and the plot range
+        // cannot start below it.
+        let start_price = price_gap(first_option, max_profit);
         let end_price = last_option + max_profit;
         Ok(calculate_price_range(start_price, end_price, step))
     }
--- a/src/strategies/short_strangle.rs
+++ b/src/strategies/short_strangle.rs
@@ -20,6 +20,7 @@ use super::base::{
     BreakEvenable, Optimizable, Positionable, Strategable, StrategyBasics, StrategyType, Validable,
 };
 use super::shared::StrangleStrategy;
+use crate::strategies::base::lower_break_even;
 use crate::{
     ExpirationDate, Options,
     chains::{StrategyLegs, chain::OptionChain, utils::OptionDataGroup},
@@ -367,8 +368,10 @@ impl BreakEvenable for ShortStrangle {
         let total_premium = self.get_net_premium_received()?;
 
         self.break_even_points.push(
-            (self.short_put.option.strike_price
-                - (total_premium / self.short_put.option.quantity).to_dec())
+            lower_break_even(
+                self.short_put.option.strike_price,
+                total_premium / self.short_put.option.quantity,
+            )
             .round_to(2),
         );
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **`OptionData::apply_spread` widens a thin quote instead of withdrawing it**
+  (#439). A contract whose mid sat below one full spread used to lose `bid`,
+  `ask` **and** `middle`. Both sides are now quoted around the mid — `mid ±
+  half_spread` — and floored at one tick (`10^-decimal_places`); a supplied mid
+  is kept, held inside the widened quote rather than cleared. Only a quote with
+  no mid and no two-sided book is still erased.
+- **Chains keep their cheap strikes, so row counts change.**
+  `OptionChain::build_chain` stops generating strikes once both wings come back
+  unpriced, so the erasure above also truncated the chain: a build with
+  `chain_size = n` returned `n + 1` rows and now returns the full `2n + 1`.
+  Downstream assertions on row counts, and anything iterating a chain, will see
+  the wings that used to disappear as they decayed.
+- `apply_spread` no longer changes the previous quote for a mid at or above the
+  tick. Below the tick it does deviate deliberately: a bid that used to round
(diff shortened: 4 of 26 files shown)


src/strategies/bear_put_spread.rs
```diff
@@ -344,9 +345,11 @@ impl BreakEvenable for BearPutSpread {
         self.break_even_points = Vec::new();
 
         self.break_even_points.push(
-            (self.long_put.option.strike_price
-                - self.get_net_cost()? / self.long_put.option.quantity)
-                .round_to(2),
+            lower_break_even(
```
joaquinbejar, line 349: 🔴 **Critical**: Returning zero here only moves the panic downstream. When the net debit exceeds the long strike, `get_profit_area()` evaluates `self.break_even_points[0] - self.short_put.option.strike_price` as a `Positive`, so a strategy that now constructs successfully still aborts; this is reproducible with long and short strikes `5/4`, quantity `1`, and long premium `50`. Handle the no-lower-break-even state in consumers, or represent it explicitly instead of as a numeric zero, and test the full strategy API.

src/strategies/long_strangle.rs
```diff
@@ -380,8 +381,11 @@ impl BreakEvenable for LongStrangle {
         let total_premium = self.get_net_cost()?;
 
         self.break_even_points.push(
-            (self.long_put.option.strike_price - (total_premium / self.long_put.option.quantity))
-                .round_to(2),
+            lower_break_even(
```
joaquinbejar, line 385: 🔴 **Critical**: This zero sentinel is incompatible with `LongStrangle::get_best_range_to_show`. When total cost exceeds the put strike, that method computes `diff = upper - 0` followed by `0 - diff`, which will panic in `Positive` arithmetic; put and call strikes `5/6`, quantity `1`, and premiums `3 + 3` reproduce it. Clamp or check the range start and add an end-to-end regression test.

src/strategies/short_strangle.rs
```diff
@@ -367,8 +368,10 @@ impl BreakEvenable for ShortStrangle {
         let total_premium = self.get_net_premium_received()?;
 
         self.break_even_points.push(
-            (self.short_put.option.strike_price
-                - (total_premium / self.short_put.option.quantity).to_dec())
+            lower_break_even(
```
joaquinbejar, line 372: 🔴 **Critical**: A zero lower break-even also makes `ShortStrangle::get_best_range_to_show` panic because it computes `start_price = first_option - max_profit`, which is negative when `first_option` is this sentinel and premium is positive. Put and call strikes `5/6`, quantity `1`, and premiums `3 + 3` reproduce it; consumers need to handle the no-lower-break-even state explicitly.

joaquinbejar (commented):
Flooring impossible lower break-evens to zero prevents construction-time subtraction panics, but several public strategy methods still treat that sentinel as an ordinary price and panic later; the change needs consumer-level handling and end-to-end tests before merge. REQUEST_CHANGES. GitHub does not allow the PR author to set the formal request-changes state.

src/strategies/bear_put_spread.rs
joaquinbejar: Fixed in 65221b06 with your reproduction as a test — long and short strikes 5/4, quantity 1, premium 50 — which aborts on the old code with `Positive invariant broken in sub`. The base of a profit area is a distance, so it now goes through a `price_gap` helper that floors at zero, and the same applies to `bear_call_spread`, `bull_call_spread` and the four single-leg strategies, which had the mirror image of the bug when the break-even sat above the strike.

src/strategies/long_strangle.rs
joaquinbejar: Fixed in 65221b06. The range start is a distance below the first break-even, so it goes through `price_gap` and stops at zero instead of going negative; with strikes 5/6, quantity 1 and premiums 3 + 3 the strategy now returns a range rather than aborting, and that is the end-to-end test.

src/strategies/short_strangle.rs
joaquinbejar: Fixed in 65221b06, same helper — `first_option - max_profit` is now `price_gap(first_option, max_profit)`. Your 5/6 strikes with premiums 3 + 3 is the third case in the new test file, and it panicked before the change.

joaquinbejar:
All three are fixed in 65221b06, and the diagnosis was exactly right: flooring the break-even only moved the abort into the consumers. Rather than patch the three call sites you found, every width measured against a break-even now goes through one `price_gap` helper that floors at zero — a profit-area base and a plot-range start are distances, and a distance that would come out negative is a region with no width. That also covers the mirror-image cases in the single-leg strategies and `custom`, which panicked when the break-even sat above the strike. Your three reproductions are now `tests/unit/strategies/no_lower_break_even_test.rs` and they abort on the old code. Suite is 3900 lib, 423 integration, 17 property and 207 doc green, with clippy, fmt and the scan clean.

Issue #442, opened 2026-08-28
Sweep the remaining modules for reachable panics


joaquinbejar:
## Overview

#440 made the Greeks, the three closed-form pricing models and the chain
builder panic-free. The rest of the library is still on `rust_decimal` and
`positive` operators, which panic on overflow, on division by zero, on `ln` of
zero, and on `exp` under/overflow. Nothing in these modules is proven safe;
they simply have not been probed yet.

Current inventory of production code (`#[cfg(test)]` blocks stripped,
50,595 lines): 595 divisions, 102 `.exp()`, 91 `.sqrt()`, 60 `.powd()`,
23 `.ln()`, 14 `unreachable!`, 5 `panic!`. Production code contains no
`.unwrap()` / `.expect()` — `make scan-banned` guards that in CI as of #440.

## Context

The method that worked in #440, to repeat per module: build a throwaway probe
crate with a path dependency on the repo, call the module's public API inside
`std::panic::catch_unwind` across the extremes of the `Decimal` domain, fix
what actually aborts, then encode the sweep as a proptest next to
`tests/property/panic_freedom_test.rs`.

The checked helpers to build on live in `src/model/decimal.rs`: `d_add`,
`d_sub`, `d_mul`, `d_div`, `d_sum`, `d_sum_iter`, `d_exp`, `d_ln`, `d_powd`,
`d_sqrt`. `positive` exposes `checked_add` / `checked_sub` / `checked_mul` /
`checked_div` / `checked_powd` / `checked_sqrt` / `checked_ln`.

## Tasks

- [ ] `src/pricing/` exotics — asian, barrier, lookback, compound, american,
      cliquet, chooser, rainbow, spread, quanto, power, exchange, binary,
      binomial, monte carlo, telegraph (the heaviest `.exp()` / `.powd()`
      concentration in the crate)
- [ ] `src/volatility/` — the IV solvers and the volatility models beyond the
      entry points already probed
- [ ] `src/curves/` and `src/surfaces/` — 44 and 39 divisions respectively,
      plus 26 `powd`, and the two `panic!` sites
- [ ] `src/strategies/` — per-strategy P&L, break-evens and the delta-neutral
      helpers; `src/strategies/base.rs` holds two `panic!`
- [ ] `src/simulation/` — walk and step kernels
- [ ] `src/chains/` beyond `build_chain` — `optiondata.rs`, `rnd.rs`,
      `generators.rs`
[...]

PR #445, opened 2026-08-28
Make the pricing exotics and numeric engines panic-free (#444)

### LINKED ISSUE #444: Panic sweep: src/pricing/ exotics and numeric kernels
## Overview

First module group of the sweep opened in #442. `src/pricing/` holds the
heaviest concentration of panicking arithmetic in the crate: 96 `.exp()`, 38
`.sqrt()`, 24 `.powd()`, 18 `.ln()` and 110 raw divisions across 20 files, all
on the operator path where `rust_decimal` and `positive` abort instead of
reporting.

`Decimal::exp()` panics with `Exp overflowed` / `Exp underflowed`, `ln()` panics
on zero and on negatives, `powd()` panics with `Pow overflowed`, and every
`Positive` operator panics on overflow, on division by zero, and when the result
would break the positivity invariant. #440 proved these are reachable from
ordinary-looking inputs on the Black-Scholes path; nothing makes the exotics
different, they simply have not been probed yet.

## Scope

- `american.rs`, `asian.rs`, `barrier.rs`, `binary.rs`, `chooser.rs`,
  `cliquet.rs`, `compound.rs`, `exchange.rs`, `lookback.rs`, `power.rs`,
  `quanto.rs`, `rainbow.rs`, `spread.rs`
- `binomial_model.rs`, `monte_carlo.rs`, `telegraph.rs`, `payoff.rs`,
  `utils.rs`

`black_scholes_model.rs`, `black_76.rs` and `garman_kohlhagen.rs` were done in
#440.

## Method

The one that worked in #440:

1. Probe crate with a path dependency on the repo, calling each pricing entry
   point inside `std::panic::catch_unwind` across the extremes of the `Decimal`
   domain: `1e-27`, `Positive::MAX`, `Decimal::MIN` / `MAX`, zero and huge time
   to expiry, rates at ±1e6, and the exotic-specific parameters (barrier
   levels, averaging windows, cap/floor, correlation).
2. Fix what actually aborts: reject an input with no financial meaning first,
   return the defined mathematical limit second, use checked arithmetic last.
   Reuse `d_add` / `d_sub` / `d_mul` / `d_div` / `d_exp` / `d_ln` / `d_powd` /
   `d_sqrt` from `src/model/decimal.rs` and the `checked_*` family on
   `Positive`.
3. Encode the sweep as a proptest next to
   `tests/property/panic_freedom_test.rs`.

## Tasks

- [ ] Probe every `pub` pricing entry point and record which inputs abort
- [ ] Convert the panicking arithmetic in the files above
- [ ] Add the proptest covering the exotic entry points
- [ ] Keep numerical results identical for well-formed inputs

## Acceptance criteria

- [ ] Numerical results unchanged for well-formed inputs — existing reference
      tests pass untouched
- [ ] Failures surface as `PricingError` / `DecimalError`, never as an abort,
      and every new fallible function documents its `# Errors`
- [ ] No `saturating_*` / `wrapping_*` introduced
      `make scan-banned` clean

## References

- #442 — umbrella
- #440 / #441 — the Greeks and Black-Scholes core, the pattern this follows
- `.claude/agents/panic-hunter.md` — method and the dependency panic surface

joaquinbejar:
## Summary

A probe over the public pricing surface — every entry point wrapped in
`catch_unwind`, driven across the extremes of the `Decimal` domain and over the
whole `OptionType` matrix — found **129 distinct panic sites** in
`src/pricing/`. They are the same class #441 fixed in the Black-Scholes core:
`rust_decimal` and `positive` abort on the operator path, so `Exp overflowed`,
`Pow overflowed`, `Unable to calculate ln for zero`, `Division by zero` and the
`Positive` invariant panics were all reachable from ordinary-looking inputs.

The single most reachable site fired 216,000 times in one probe run: a barrier
option with a zero barrier level divides by it. The most embarrassing one is
not an extreme input at all — an out-of-the-money chooser at `S = 100`,
`K = 110` aborted, and so did a power put whose `S^n` exceeds the strike.

The probe is now clean over **31,365,504 calls**, including a pass with a
`TRACE` subscriber installed.

## Changes

Every panicking site now goes through the checked helpers in
`src/model/decimal.rs`. Where a formula has a defined limit, the limit is
returned rather than an error:

- **`american.rs`** — `4M/K` with `K = 1 - e^(-rT)` is a 0/0 at `r = 0`, which
  is an ordinary input, so **every zero-rate American option aborted**.
  Replaced by its limit `8/(sigma^2 T)`.
- **`asian.rs`** — a non-positive Turnbull-Wakeman `M1`, or a matched variance
  that collapses, falls into the `sigma -> 0` branch the function already had.
- **`chooser.rs`** — a collapsed diffusion saturates the CDFs by the sign of
  the numerator; a genuine 0/0 returns a typed error rather than a guess.
- **`exchange.rs`, `spread.rs`, `quanto.rs`** — a zero second asset or strike,
  and a moneyness that rounds below the representable scale, take the same
  infinity limits.
- **`barrier.rs`, `binary.rs`, `cliquet.rs`, `compound.rs`, `lookback.rs`,
  `power.rs`, `rainbow.rs`** — discounts, reflection powers, moment matching
  and the intrinsics rebuilt on `d_exp` / `d_ln` / `d_powd` / `d_div` /
  `d_mul` / `d_sub`.
- **`binomial_model.rs`, `monte_carlo.rs`, `telegraph.rs`, `utils.rs`** — the
  lattice and simulation engines, including the indexing: `prices[i]` became
  `.get()` / `.get_mut()` with a typed error, an unchecked `.sum()` over the
  Monte-Carlo payoffs became `d_sum_iter`, and the reset loop in `cliquet`
  moved from `[i-1]` / `[i]` to `windows(2)`.

Domain rejects were added only where an input has no financial meaning: a zero
underlying, strike or barrier level in `barrier.rs`, and a non-positive
underlying or strike in the BAW helper.

### The root cause outside `src/pricing/`

`OptionType::payoff` aborted for an out-of-the-money chooser and for a power
put whose `S^n` exceeds the strike, and it reached the lattice, the telegraph
and the Monte-Carlo engines through their payoff evaluation. Both arms now form
the intrinsic on `Decimal` and floor it, which is the same arithmetic
`Positive::sub` was doing internally — every in-the-money price is
bit-identical, only the aborting path changed. The alternative, rejecting those
types at the engine entry, would have turned a currently-correct
exotic-at-expiry price into an unsupported-type error, so it was not taken.

## Technical decisions

The mathematical limits are listed above and each is the limit of the formula
the function already implements, not a substitute for it. The one place the
sweep refused to guess — the chooser's genuine 0/0, where both the diffusion
and the log-moneyness collapse — returns a typed `MethodError`.

`telegraph.rs::next_state` returns `i8` and could not become fallible without a
public change, so an unrepresentable `lambda * dt` degrades to the limit its
sign implies, mirroring the `to_f64`-failure degradation the function already
had.

Left alone deliberately: the `f64` Gauss-Legendre kernels in `compound.rs`,
`power.rs` and `rainbow.rs`. `f64` arithmetic does not panic, the indices are
provably in range on fixed-size arrays, and every `f64 -> Decimal` boundary
already produces a typed error for a non-finite value. Rewriting them risks
numeric drift for no panic-freedom gain.

## Public API impact

None. Every pricing entry point already returned `Result<_, PricingError>`.
`intrinsic_value` in `asian.rs` became fallible, but it is a private helper.

## Testing

- [x] Unit tests added or updated — 7 new in `src/model/types.rs::tests_payoff`
      covering both payoff arms at, above and below the money
- [x] Reference regression test added — the existing reference and parity tests
      pass **unedited**, which is the bar for a sweep that must not move a price
- [ ] Criterion benchmark added/updated
- [x] `make pre-push` equivalents run locally and clean

`tests/property/pricing_panic_freedom_test.rs` is the repeatable form of the
probe: three properties at 512 cases each, driving all twelve exotic entry
points over 31 `OptionType` shapes (zero barrier, empty reset schedule, rainbow
with zero assets, zero exponent, `Positive::MAX` choice date), three
`exotic_params` shapes, and the extreme numeric domain; plus the BAW `r = 0`
boundary and the lattice and simulation engines.

Suite on the rebased branch: **3913 lib, 423 integration, 20 property, 207 doc
— 0 failures**, with `cargo fmt --all --check`, `cargo clippy --all-targets
--all-features --workspace -- -D warnings` and `make scan-banned` clean.

## Checklist

- [x] `README.tpl` updated if public surface changed — not applicable

## Stack

- Base branch: `main` (#441 and #443 are merged)
- Stack: #441 → #443 → **#444 (this)** → 442b → 442c → 442d → 442e
- Part of the sweep opened in #442; the remaining module groups are `curves/`
  and `surfaces/`, `chains/` and `simulation/`, `strategies/`, and the
  close-out that extends `make scan-banned` to the math operators.

Closes #444

### DIFF
--- a/src/pricing/asian.rs
+++ b/src/pricing/asian.rs
@@ -25,6 +25,7 @@
 use crate::Options;
 use crate::error::PricingError;
 use crate::greeks::{big_n, d1, d2};
+use crate::model::decimal::{d_add, d_div, d_exp, d_ln, d_mul, d_powd, d_sqrt, d_sub};
 use crate::model::types::{AsianAveragingType, OptionStyle, OptionType};
 use positive::Positive;
 use rust_decimal::Decimal;
@@ -43,9 +44,16 @@ use rust_decimal_macros::dec;
 ///
 /// # Errors
 ///
-/// Returns `PricingError` if:
-/// - The option type is not Asian
-/// - Required parameters are invalid (zero volatility, etc.)
+/// - [`PricingError::MethodError`] when the option type is not
+///   `OptionType::Asian`, when the averaging type is an unsupported
+///   `#[non_exhaustive]` variant, or when the expiration cannot be converted
+///   to a year fraction.
+/// - [`PricingError::Decimal`] when an intermediate step leaves the
+///   representable `Decimal` range: the discount factor, the forward, either
+///   Turnbull-Wakeman moment, the moment-matched variance, or the final
+///   Black legs.
+/// - [`PricingError::Positive`] when the `σ / √3` geometric adjustment is not
+///   representable as a `Positive`.
 pub fn asian_black_scholes(option: &Options) -> Result<Decimal, PricingError> {
     match &option.option_type {
         OptionType::Asian { averaging_type } => match averaging_type {
@@ -84,26 +92,64 @@ fn geometric_asian_price(option: &Options) -> Result<Decimal, PricingError> {
         .map_err(|e| PricingError::other(&e.to_string()))?;
 
     if t == Positive::ZERO {
-        return Ok(intrinsic_value(option));
+        return intrinsic_value(option);
     }
 
     if sigma == Positive::ZERO {
-        // Deterministic case
-        let discount = (-r * t).exp();
-        let forward = s * ((r - q) * t).exp();
+        // Deterministic case: the average of a driftless-variance path is the
+        // forward itself, so the option collapses to a discounted intrinsic.
+        let t_dec = t.to_dec();
+        let discount = d_exp(
+            d_mul(-r, t_dec, "pricing::asian::geometric::det::neg_rt")?,
+            "pricing::asian::geometric::det::discount",
+        )?;
+        let carry = d_sub(r, q, "pricing::asian::geometric::det::carry")?;
+        let forward = d_mul(
+            s.to_dec(),
+            d_exp(
+                d_mul(carry, t_dec, "pricing::asian::geometric::det::carry_t")?,
+                "pricing::asian::geometric::det::growth",
+            )?,
+            "pricing::asian::geometric::det::forward",
+        )?;
         let intrinsic = match option.option_style {
-            OptionStyle::Call => (forward - k).max(Positive::ZERO).to_dec(),
-            OptionStyle::Put => (k - forward).max(Positive::ZERO).to_dec(),
+            OptionStyle::Call => {
+                d_sub(forward, k.to_dec(), "pricing::asian::geometric::det::call")?
+                    .max(Decimal::ZERO)
+            }
+            OptionStyle::Put => d_sub(k.to_dec(), forward, "pricing::asian::geometric::det::put")?
+                .max(Decimal::ZERO),
         };
-        return Ok(apply_side(intrinsic * discount, option));
+        let price = d_mul(intrinsic, discount, "pricing::asian::geometric::det::price")?;
+        return Ok(apply_side(price, option));
     }
 
     // Geometric average adjustments (Kemna-Vorst)
-    let sigma_sq = sigma * sigma;
+    let sigma_sq = d_mul(
+        sigma.to_dec(),
+        sigma.to_dec(),
+        "pricing::asian::geometric::sigma_sq",
+    )?;
     let sqrt_three = Positive::new(3.0_f64.sqrt())
         .map_err(|e| PricingError::method_error("geometric_asian_price", &e.to_string()))?;
-    let sigma_adj = sigma / sqrt_three;
-    let b_adj = (r - q - sigma_sq / dec!(6)) / dec!(2);
+    let sigma_adj = Positive::new_decimal(d_div(
+        sigma.to_dec(),
+        sqrt_three.to_dec(),
+        "pricing::asian::geometric::sigma_adj",
+    )?)?;
+    let b_adj = d_div(
+        d_sub(
+            d_sub(r, q, "pricing::asian::geometric::carry")?,
+            d_div(
+                sigma_sq,
+                dec!(6),
+                "pricing::asian::geometric::variance_drag",
+            )?,
+            "pricing::asian::geometric::b_numerator",
+        )?,
+        dec!(2),
+        "pricing::asian::geometric::b_adj",
+    )?;
 
     // Calculate d1 and d2 with adjusted parameters
     let d1_val = d1(s, k, b_adj, t, sigma_adj)
@@ -111,18 +157,49 @@ fn geometric_asian_price(option: &Options) -> Result<Decimal, PricingError> {
     let d2_val = d2(s, k, b_adj, t, sigma_adj)
         .map_err(|e: crate::error::GreeksError| PricingError::other(&e.to_string()))?;
 
-    let discount = (-r * t).exp();
+    let t_dec = t.to_dec();
+    let discount = d_exp(
+        d_mul(-r, t_dec, "pricing::asian::geometric::neg_rt")?,
+        "pricing::asian::geometric::discount",
+    )?;
+    // e^((b_adj - r) T): the geometric-average carry replaces the spot drift.
+    let carry_discount = d_exp(
+        d_mul(
+            d_sub(b_adj, r, "pricing::asian::geometric::carry_spread")?,
+            t_dec,
+            "pricing::asian::geometric::carry_spread_t",
+        )?,
+        "pricing::asian::geometric::carry_discount",
+    )?;
+    let s_leg = d_mul(
+        s.to_dec(),
+        carry_discount,
+        "pricing::asian::geometric::spot_leg",
+    )?;
+    let k_leg = d_mul(
+        k.to_dec(),
+        discount,
+        "pricing::asian::geometric::strike_leg",
+    )?;
 
     let price = match option.option_style {
         OptionStyle::Call => {
             let n_d1 = big_n(d1_val).unwrap_or(Decimal::ZERO);
             let n_d2 = big_n(d2_val).unwrap_or(Decimal::ZERO);
-            s.to_dec() * ((b_adj - r) * t).exp() * n_d1 - k.to_dec() * discount * n_d2
+            d_sub(
+                d_mul(s_leg, n_d1, "pricing::asian::geometric::call::spot")?,
+                d_mul(k_leg, n_d2, "pricing::asian::geometric::call::strike")?,
+                "pricing::asian::geometric::call",
+            )?
         }
         OptionStyle::Put => {
             let n_neg_d1 = big_n(-d1_val).unwrap_or(Decimal::ZERO);
             let n_neg_d2 = big_n(-d2_val).unwrap_or(Decimal::ZERO);
-            k.to_dec() * discount * n_neg_d2 - s.to_dec() * ((b_adj - r) * t).exp() * n_neg_d1
+            d_sub(
+                d_mul(k_leg, n_neg_d2, "pricing::asian::geometric::put::strike")?,
+                d_mul(s_leg, n_neg_d1, "pricing::asian::geometric::put::spot")?,
+                "pricing::asian::geometric::put",
+            )?
         }
     };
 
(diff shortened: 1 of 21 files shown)


src/pricing/spread.rs
```diff
@@ -144,47 +149,141 @@ fn kirk_approximation(
...
+    let ratio = d_div(s1, adjusted_strike, "pricing::spread::kirk::moneyness")?;
+
+    // `N(d1)`, `N(d2)`, `N(-d1)`, `N(-d2)`. A collapsed `σ√T` or a moneyness
+    // that underflowed below the representable scale drives the normal
+    // arguments to `±∞`, where the CDFs saturate: those are the limits of the
+    // formula, not substitutes for it.
+    let (n_d1, n_d2, n_neg_d1, n_neg_d2) = if denominator.is_zero() {
+        // σ√T → 0: the option is worth its discounted intrinsic, i.e. the
+        // step function at the forward.
+        if ratio >= dec!(1.0) {
```
joaquinbejar: 🔴 **Critical**: This will misprice the zero-volatility limit because exercise is selected from the undiscounted spot ratio. With `S1 = 100`, `S2 = 100`, `K = 10`, `r = 1`, `q1 = 0`, `T = 1` and both volatilities zero, it returns zero while the deterministic call is `100 - 110e^-1 ≈ 59.53`. Compare `s1_pv` with `adjusted_strike_pv`, or return their floored difference directly, and add this regression.

src/pricing/exchange.rs
```diff
@@ -124,30 +128,98 @@ fn margrabe_formula(
...
+
+    // `S2 = 0`: the exchange ratio diverges, `N(d1) = N(d2) = 1` and the
+    // option collapses to the present value of the first asset.
+    if s2.is_zero() {
+        return Ok(s1_pv.max(dec!(0.0)));
+    }
+    let ratio = d_div(s1, s2, "pricing::exchange::ratio")?;
+    // `S1 = 0` (or a ratio that rounded below the representable scale): both
+    // normal arguments diverge to `-∞`, `N(d1) = N(d2) = 0`, price zero.
+    if ratio.is_zero() {
```
joaquinbejar: 🔴 **Critical**: A ratio rounded to zero does not imply that both normal arguments go to negative infinity because the `(q2 - q1)T` carry can offset the log ratio. With `S1 = 1e-27`, `S2 = 100`, `q2 = 100`, `T = 1` and both volatilities `0.2`, this returns zero even though `S2`'s present value has vanished and the exchange option is worth `S1`'s present value. Compute log-moneyness as `ln(S1) - ln(S2)` without rounding the ratio first. The same shortcut in `src/pricing/spread.rs` at lines 236 and 376 needs the same fix.

src/pricing/asian.rs
```diff
@@ -147,94 +224,255 @@ fn arithmetic_asian_price(option: &Options) -> Result<Decimal, PricingError> {
...
+        )?;
+        let term1 = d_div(
+            d_mul(
+                d_mul(dec!(2), s_sq, "pricing::asian::arithmetic::m2::two_s_sq")?,
+                term1_exp,
+                "pricing::asian::arithmetic::m2::term1_numerator",
+            )?,
+            d_mul(
+                d_mul(
+                    b_plus_var,
```
joaquinbejar, line 410: 🟠 **Warning**: This will reject valid inputs such as `r = 0`, `q = 0.04`, `sigma = 0.20` and `T = 1` because `b + sigma^2` becomes zero. The Turnbull-Wakeman second moment has removable limits at both `b = -sigma^2` and `b = -sigma^2 / 2`, so the checked division must not surface an error here. Handle both limits analytically or evaluate the defining integral, then add regressions for both boundaries.

joaquinbejar (commented):
The checked arithmetic sweep is a solid reliability improvement, but the new limit handling still produces two incorrect prices and rejects a valid Turnbull Wakeman boundary. The diff is over 5,000 lines, so please split future sweeps into smaller module groups to keep numerical review tractable. REQUEST_CHANGES. GitHub does not allow the PR author to set that formal review state, so this review is submitted as COMMENT.

joaquinbejar:
Quant review of the seven limits is in: six correct, one wrong, and the wrong one was introduced by this sweep rather than exposed by it. Kirk's collapsed-diffusion branch tested the spot moneyness `S1 / (S2 + K)` with the step test that belongs to quanto, where the ratio is already a forward, so it dropped the carry `(r - q1)T` from the sign. It is reachable from well-formed inputs — Kirk's adjusted volatility is exactly zero at `rho = 1, sigma1 = w * sigma2` — and with `S1 = S2 = 100`, `K = 25`, `sigma2 = 0.25`, `r = 25%` the call priced at 0 against a discounted intrinsic of 2.65. Fixed in 5c4963fd by comparing the two present values, which is the shape `exchange.rs` already uses, with three regression tests; the first fails on the previous commit.

The other six hold, with numerical confirmation on the BAW limit (201.0017, 200.0100, 200.0000 against 200 as `r` goes to zero) and a line-by-line diff of every rewritten expression against its original. Two rounding notes worth having on record: the only reassociation in the whole sweep is `s * sigma * sqrt(t)` becoming `s * (sigma * sqrt(t))` in three lookback expressions, worth about 1e-27 absolute on a price of order 10, and `d_div` is exactly `/` because `round_dp_with_strategy` is a no-op at `MAX_SCALE`.

Five pre-existing defects surfaced next to the sweep without being part of it, each now its own issue: #447 (the Asian zero-volatility branch prices the terminal forward instead of the average, 2.53 against 1.26), #448 (the complex chooser does not implement the Rubinstein formula it documents, 7% off Haug's worked example), #449 (zero-volatility binomial drops the early-exercise premium, 5.13 against 10 on an American put), #450 (`Point2D` orders on `(x, y)` but compares equal on `x` alone) and #451 (bilinear interpolation near the upper boundary). None of them is touched here.

joaquinbejar (commented):
The Kirk present value fix and its regressions are correct, but the current head still reproduces the two unresolved blocking threads on exchange ratio underflow and the Turnbull Wakeman removable boundary. REQUEST_CHANGES. GitHub does not allow the PR author to set that formal review state, so this review is submitted as COMMENT.

src/pricing/spread.rs
joaquinbejar: Fixed in 5c4963fd, before this second review landed. The branch now compares `s1_pv` against `adjusted_strike_pv` instead of the undiscounted ratio, and three regressions cover the call, the put and the flip to a negative rate; the first fails on the previous commit with a price of 0.

src/pricing/exchange.rs
joaquinbejar: Fixed in 865b2f1b at all three sites you named. The log-moneyness is now `ln(S1) - ln(S2)` rather than `ln(S1 / S2)`, so the quotient never rounds away the information the carry needs, and your case prices at `S1`'s present value. A ratio that genuinely vanishes in the limit still returns zero and has its own test alongside it.

src/pricing/asian.rs
joaquinbejar: Fixed in 865b2f1b. Both removable singularities are handled by going back to the defining double integral and evaluating the degenerate exponential directly instead of dividing by its rate, so the values are exact rather than approximations, and your `r = 0, q = 0.04, sigma = 0.20, T = 1` case prices. Each boundary has a continuity test against the price a hair either side of it, which is what proves the limit rather than just exercising the branch.

joaquinbejar:
All three are fixed and the two blocking ones landed in 865b2f1b. The exchange and Margrabe sites now take the log before the ratio, which is the part that mattered: `ln(S1) - ln(S2)` is exact exactly where `ln(S1 / S2)` has already thrown the information away, and your carry case prices at `S1`'s present value instead of zero. The Turnbull-Wakeman boundaries are handled from the defining integral rather than by dividing by a vanishing rate, so both are exact, and each has a continuity test against the price either side of the boundary rather than a bare assertion that it returns something.

Taking the process point: the next sweeps go out one module group per PR. This one was 5,000 lines because `src/pricing/` was treated as a single group, and both defects you found here were in the parts where the sweep was choosing a limit rather than removing a panic — precisely the material that a diff that size hides. Splitting also lets the unchanged-behaviour claim be made per file instead of in aggregate, which is what would have caught these before you had to.

Issue #451, opened 2026-08-29
Define bilinear interpolation near the upper boundary


joaquinbejar:
## Overview

`Curve::bilinear_interpolate` reads a four-point window starting at the bracket
index, so it runs off the end for any `x` in the last three intervals. On a
four-point curve `(0,0), (1,1), (2,4), (3,9)` it fails at `x = 1.5` and at
`x = 2.5`; a five-point curve fails at `x = 2.5`.

#446 made it return `InterpolationError::Bilinear` there instead of aborting.
That was the conservative call: it changes no currently-working result. But it
leaves bilinear interpolation unusable across most of its domain, which is not
a satisfying end state.

## Context

- `src/curves/curve.rs:538-541` — the four-point window
- `find_bracket_points` supplies the index
- `cubic_interpolate` in the same file already handles its upper boundary by
  clamping the window to `(len-4, len-3, len-2, len-1)`, which is the obvious
  precedent

## The decision to make

Clamping the window start to `len - 4` follows the in-file precedent and makes
the method usable everywhere, but it decides a numeric answer that is currently
undefined — the interpolant near the right edge stops being local. That needs
validating against a reference implementation rather than being asserted, which
is why it was not folded into a panic sweep.

## Tasks

- [ ] Decide what bilinear interpolation should mean on a 1-D curve near the
      upper boundary
- [ ] Implement it, and say in the doc comment what the boundary behaviour is
- [ ] Validate against a reference for at least three interior points and both
      boundaries
- [ ] Point `InterpolationError::Bilinear`'s doc at the resolution

## Acceptance criteria

- [ ] Bilinear interpolation returns a value across the whole interior domain,
      or documents precisely where it does not and why
- [ ] Values at the interior points match the reference
- [ ] The existing interpolation tests are unchanged

Issue #450, opened 2026-08-29
Point2D orders on (x, y) but compares equal on x alone


joaquinbejar:
## Overview

`Point2D` orders on `(x, y)` but compares equal on `x` alone, so two points
that `PartialEq` calls equal can order differently. That breaks the contract
`Ord` requires (`a == b` iff `a.cmp(b) == Ordering::Equal`), and the collections
built on it inherit the breakage: a `BTreeSet<Point2D>` can legitimately hold
two points sharing an abscissa, which is why every `/ (p2.x - p1.x)` in the
interpolators is reachable at all.

Found by `panic-hunter` while sweeping `src/curves/` for #446. That sweep
returns `InterpolationError::DegenerateInterval` for the division, which is the
right answer for the arithmetic, but the underlying inconsistency is a domain
modelling question and was deliberately left alone.

## Context

- `src/curves/types.rs` (or wherever `Point2D` derives / implements `Ord`,
  `PartialEq` and `Eq`)
- `Curve` stores its points in a `BTreeSet<Point2D>`
- Rust's own docs on the `Ord` / `PartialEq` consistency requirement

## Tasks

- [ ] Decide the intended semantics: is a curve a function of `x` (at most one
      point per abscissa, so `Eq` on `x` is right and `Ord` must match) or a
      set of points in the plane (so `Eq` must include `y`)
- [ ] Make `Ord`, `PartialOrd`, `PartialEq`, `Eq` and `Hash` mutually
      consistent with that decision
- [ ] If a curve is a function of `x`, reject or merge a duplicate abscissa at
      construction instead of storing both

## Acceptance criteria

- [ ] `a == b` iff `a.cmp(&b) == Ordering::Equal` for every pair, covered by a
      proptest
- [ ] A duplicated abscissa is either impossible by construction or handled
      explicitly with a documented rule
- [ ] Existing curve tests unchanged

joaquinbejar:
Exact locations from the sweep: `src/curves/types.rs:69-75` implements `PartialEq`/`Eq` on `self.x == other.x` alone, `src/curves/types.rs:90-97` implements `Ord` on `(x, y)`, and `src/curves/types.rs:77-82` hashes only `x`, so `Hash` agrees with `Eq` but not with `Ord`. `BTreeSet` dispatches on `Ord`, which is why a curve holds `(1, 2)` and `(1, 3)` at once while a `HashSet` would treat them as one point.

[...]

PR #452, opened 2026-08-29
Make curves, surfaces and geometrics panic-free (#446)

### LINKED ISSUE #446: Panic sweep: curves, surfaces and geometrics
## Overview

Second module group of the sweep opened in #442, after #444 took `src/pricing/`.
`src/curves/`, `src/surfaces/` and `src/geometrics/` hold 112 raw divisions, 26
`.powd()`, 7 `.sqrt()` and the two remaining `panic!` sites in production code,
all on the operator path where `rust_decimal` and `positive` abort instead of
reporting.

Interpolation and surface construction take user-supplied points, so the same
inputs that broke the Greeks in #441 — a value at the edge of the `Decimal`
range, two points that collapse onto the same abscissa, an empty point set —
reach this arithmetic directly.

## Scope

- `src/curves/` — `curve.rs` (44 divisions, 7 `powd`, 1 `panic!`), `traits.rs`,
  `utils.rs`, and the interpolation implementations
- `src/surfaces/` — `surface.rs` (39 divisions, 19 `powd`, 1 `panic!`),
  `types.rs`, `utils.rs`
- `src/geometrics/` — interpolation, intersection and the grid helpers

## Method

The one that worked in #441 and #444:

1. Probe crate with a path dependency on the repo, calling each public curve
   and surface entry point inside `std::panic::catch_unwind` over the extremes
   of the `Decimal` domain, plus the structural extremes this module has:
   an empty point set, a single point, duplicated abscissae, points already
   sorted and reversed, and `Positive::MAX` coordinates.
2. Fix what actually aborts: reject an input with no geometric meaning first,
   return the defined limit second, use checked arithmetic last. Reuse
   `d_add` / `d_sub` / `d_mul` / `d_div` / `d_exp` / `d_ln` / `d_powd` /
   `d_sqrt` from `src/model/decimal.rs` and the `checked_*` family on
   `Positive`.
3. Resolve the two `panic!` sites: each becomes a typed error or gains a
   `// scan-banned: allow -- <reason>` marker with the argument for why it
   cannot fire.
4. Encode the sweep as a proptest next to
   `tests/property/pricing_panic_freedom_test.rs`.

## Tasks

- [ ] Probe every public curve, surface and geometrics entry point
- [ ] Convert the panicking arithmetic in the three modules
- [ ] Resolve the two `panic!` sites
- [ ] Add the proptest covering curve and surface construction and interpolation
- [ ] Keep numerical results identical for well-formed inputs

## Acceptance criteria

- [ ] Numerical results unchanged for well-formed inputs — existing tests pass
      untouched
- [ ] Failures surface as `CurveError` / `SurfaceError` / `DecimalError`, never
      as an abort, and every new fallible function documents its `# Errors`
      `make scan-banned` clean

## References

- #442 — umbrella
- #444 / #445 — the pricing sweep, the pattern this follows
- `.claude/agents/panic-hunter.md` — method and the dependency panic surface

joaquinbejar:
## Summary

A probe over the public curve and surface API found **372 panics in 4,108
calls**: 134 `Subtraction overflowed`, 117 `Multiplication overflowed`, 40
`Addition overflowed`, 34 `Division by zero`, 21 `Pow overflowed` and 26
out-of-bounds indexes. The probe is now at **zero**.

Most of them came from ordinary inputs rather than extreme ones, because this
module's degeneracies are geometric, not numeric: an empty point set, a single
point, two points sharing an abscissa, a one-column surface. Every interpolator
divided by `x2 - x1` with the raw operator, and a repeated abscissa is not
hypothetical — `Point2D` orders on `(x, y)` but compares equal on `x` alone, so
a `BTreeSet<Point2D>` legitimately holds both. That inconsistency is left alone
here and tracked in #450; the division now returns
`InterpolationError::DegenerateInterval`, which is what the variant already
documented.

## Breaking change

`create_linear_curve` and `create_constant_curve` now return
`Result<Curve, CurveError>` instead of `Curve`. Both aborted the caller's
process on a span or an ordinate that leaves the `Decimal` range:
`create_linear_curve(-Decimal::MAX, Decimal::MAX, dec!(1))` panicked with
`Subtraction overflowed`, and `create_linear_curve(dec!(0), dec!(10),
Decimal::MAX)` with `Multiplication overflowed`. There is no way to keep an
infallible signature without inventing a value, so **`cargo-semver-checks` will
flag this PR, and that is the intended outcome, not a failure to work around**.
Callers add `?` or `.expect(..)`.

## Other paths that now report instead of aborting

- `compute_trend_metrics` on both curves and surfaces returns
  `MetricsError::TrendError` when the least-squares denominator is zero — every
  abscissa equal, or abscissas so small their squares underflow. A one-column
  surface is the ordinary case that used to abort.
- `bilinear_interpolate` returns `InterpolationError::Bilinear` when its
  four-point window runs past the end of the curve, which is any `x` in the
  last three intervals. Conservative on purpose: it changes no
  currently-working result. Making it usable across the whole domain means
  deciding a numeric answer that is undefined today, which is #451.
- `Surface::linear_interpolate` reports a surface with fewer than three points,
  and three collinear nearest neighbours.
- The surface metrics gained the empty-input and zero-dispersion guards the
  curve metrics already had.
- `generate_statistical_curve` rejects `num_points` above the curve's own
  x-value count instead of indexing past the end.
- `verify_curve_metrics` reports a gap to a target metric that is itself
  unrepresentable.

## The two `panic!` sites

Both are `impl Index<usize>` (`curves/curve.rs`, `surfaces/surface.rs`).
`std::ops::Index::index` returns `&Self::Output` with no fallible channel, so
the choice is to keep the panic or remove a public trait impl. They keep it,
behind `// scan-banned: allow` with the `Vec::index` contract as the argument —
and every internal `self[i]` in the library is gone, which is where 16 of the
372 panics came from. Four files shed their
`#![allow(clippy::indexing_slicing)]` entirely: `curves/curve.rs`,
`curves/traits.rs`, `surfaces/surface.rs` and
`geometrics/interpolation/traits.rs`.

## Technical decisions

The 43 call sites that gained a mechanical `.unwrap()` for the signature change
are all inside `#[cfg(test)] mod` blocks — 39 in `curves/curve.rs`, 4 in
`geometrics/operations/axis.rs`. No assertion or expected value was touched,
and no production call site needed one.

`calculate_prominence` looked reachable and turned out not to be: for its
subtraction to overflow the y-spread must exceed 7.9e28, but
`compute_shape_metrics` computes the variance first and fails above 2.8e14. A
second probe confirmed it over 27 runs, so that file is untouched and keeps its
existing allow.

## Public API impact

`create_linear_curve` and `create_constant_curve` change their return type, as
above. The re-export in `curves/mod.rs` needed no change. Everything else in
the module keeps its signature; the new failures travel through existing error
variants.

## Testing

- [x] Reference regression test added — the existing curve and surface tests
      pass **unedited**, which is the bar for a sweep that must not move a value
- [ ] Criterion benchmark added/updated
- [x] `make pre-push` equivalents run locally and clean

`tests/property/curves_panic_freedom_test.rs` adds 11 properties covering
construction, interpolation, metrics and the statistical helpers over both the
`Decimal` extremes and the structural degeneracies. They also pass at
`PROPTEST_CASES=8192`. One of them found a panic the hand-written probe had not
thought to reach: `verify_curve_metrics` with a `Decimal::MIN` target mean
against an ordinary curve mean.

Suite: **4574 passed, 0 failed, 79 ignored**, with `cargo fmt --all --check`,
`cargo clippy --all-targets --all-features --workspace -- -D warnings` and
`make scan-banned` clean.

## Checklist

- [x] `README.tpl` updated if public surface changed — the two constructors are
      not named in it, so no regeneration is needed

## Stack

- Base branch: `stack/2-444-panic-sweep-pricing` (PR #445)
- Stacked on: #445
- Stack: #441 → #443 → #445 → **#446 (this)** → 442c → 442d → 442e
- The diff is cumulative until #445 merges; review it against the base branch.

Closes #446

### DIFF
--- a/src/surfaces/surface.rs
+++ b/src/surfaces/surface.rs
@@ -1,9 +1,3 @@
-// Scoped allow: bulk migration of unchecked `[]` indexing to
-// `.get().ok_or_else(..)` tracked as follow-ups to #341. The existing
-// call sites are internal to this file and audited for invariant-bound
-// indices (fixed-length buffers, just-pushed slices, etc.).
-#![allow(clippy::indexing_slicing)]
-
 //! This module provides functionality for visualizing and plotting 3D surfaces.
 //! It leverages the `plotters` crate for rendering and offers a flexible API for
 //! customizing plot appearance and saving outputs.  It supports plotting single
@@ -28,13 +22,15 @@
 //!
 
 use crate::curves::{Curve, Point2D};
+use crate::error::decimal::DecimalError;
 use crate::error::{InterpolationError, MetricsError, SurfaceError};
 use crate::geometrics::{
     Arithmetic, AxisOperations, BasicMetrics, BiLinearInterpolation, ConstructionMethod,
     ConstructionParams, CubicInterpolation, GeometricObject, GeometricTransformations, Interpolate,
     InterpolationType, LinearInterpolation, MergeAxisInterpolate, MergeOperation, MetricsExtractor,
-    RangeMetrics, RiskMetrics, ShapeMetrics, SplineInterpolation, TrendMetrics,
+    RangeMetrics, RiskMetrics, ShapeMetrics, SplineInterpolation, TrendMetrics, powu_checked,
 };
+use crate::model::decimal::{d_add, d_div, d_mul, d_sub, d_sum_iter};
 use crate::surfaces::Point3D;
 use crate::surfaces::types::Axis;
 use crate::utils::Len;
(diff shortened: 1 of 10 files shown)


joaquinbejar (commented):
The panic sweep is thorough and the property checks are clean, but two new surface branches silently return incorrect financial values. The delta is over 2,400 lines against its stacked base, so future sweeps should be split into smaller module groups. REQUEST_CHANGES. GitHub does not allow the PR author to set that formal review state, so this review is submitted as COMMENT.

src/surfaces/surface.rs
```diff
@@ -1200,28 +1530,55 @@ impl MetricsExtractor for Surface {
...
+        let sum_sq = central_moment(&z_values, mean, 2, op).map_err(risk_err)?;
+        let variance = d_div(sum_sq, Decimal::from(z_values.len()), op).map_err(risk_err)?;
+        let volatility = variance.sqrt().unwrap_or(Decimal::ZERO);
+
+        // A flat surface has no dispersion, so the Sharpe ratio is undefined;
+        // mirror the zeroed answer `Curve::compute_risk_metrics` returns.
+        if volatility == Decimal::ZERO {
+            return Ok(RiskMetrics {
+                volatility,
+                value_at_risk: Decimal::ZERO,
```
joaquinbejar: 🔴 **Critical**: This zero-volatility shortcut silently changes the function's own VaR limit. For a flat surface at `z = 5`, the formula below gives `mean - 1.645 * 0 = 5`, but this returns `value_at_risk = 0`. Keep the deterministic VaR at the mean and only zero fields that are genuinely undefined, then add a flat nonzero regression.

src/surfaces/surface.rs
```diff
@@ -1306,23 +1668,36 @@ impl Arithmetic<Surface> for Surface {
...
+                                        let acc = acc?;
+                                        if val == Decimal::ZERO {
+                                            Ok(acc)
+                                        } else {
+                                            d_div(acc, val, op).map_err(construction_err)
+                                        }
+                                    },
                                 )
-                                .reduce(|| first, |a, _b| a)
+                                .reduce(|| Ok(first), |a, _b| a)?
```
joaquinbejar: 🔴 **Critical**: This reducer always keeps its left input, so every fold result and any checked-arithmetic error are discarded. Merging constant surfaces with `z` values `8`, `2`, and `2` returns a surface near `8` everywhere instead of `8 / 2 / 2 = 2`. Division is not associative, so use a sequential `try_fold` from the first value and add a non-unit regression.

src/surfaces/surface.rs
joaquinbejar: Fixed in 2fd6bf1a. The early return is gone rather than patched, so every field now falls out of the ordinary formula and only `sharpe_ratio` is special-cased as the one genuinely undefined at zero dispersion; a flat surface at 5 reports a VaR of 5 and a flat one at -2.5 reports -2.5, both covered by regressions that fail on the previous commit. The same guard was in `Curve::compute_risk_metrics` — it is where the surface copied it from — so it is fixed there too.

src/surfaces/surface.rs
joaquinbejar: Fixed in 2fd6bf1a with a sequential `try_fold`, and it was worse than estimated: your 8, 2, 2 case returned exactly 8, because rayon starts every partial from the `Ok(first)` identity and the winning partial had no division applied at all. The swallowed-error half has its own regression, since two surfaces whose quotient passes `Decimal::MAX` used to return `Ok` with the overflow discarded. `Curve::merge` folds reciprocals and was already correct; a three-object test now pins that, because the existing two-curve test could not tell the two shapes apart.

joaquinbejar:
Both are fixed in 2fd6bf1a, and both were the right things to catch — a wrong number is worse than the aborts this sweep removed, because a wrong number gets used. The VaR guard turned out to be a copy: `Curve::compute_risk_metrics` had the same early return and is where the surface inherited it, so it is fixed in both, with flat positive and flat negative regressions. The existing `test_constant_curve_risk_metrics` asserted volatility, beta and sharpe ratio but never `value_at_risk`, which is precisely the gap that let it through.

On the reducer, your estimate was conservative: 8, 2, 2 returned exactly 8, not near it, because every rayon partial starts from the `Ok(first)` identity and the leftmost never had a division applied. The half that worries me more is that the same reducer discarded checked-arithmetic errors, so an overflow inside the fold came back as `Ok` — that undercuts the sweep's whole claim, and it now has its own regression. `Curve::merge` folds reciprocals and was already correct; a three-object test pins it, since the existing two-curve test could not distinguish the two shapes.

One thing I left alone deliberately and filed as #453: the Multiply arms still reduce in parallel, and `Decimal` rounding past 28 places makes that non-deterministic in the last digits for three or more objects. Pre-existing, same family of problem, but trading parallelism for determinism deserves Criterion numbers rather than a line slipped into a panic sweep.

Process point taken, and the two defects prove it: both were in the parts of the diff that preserved or mirrored existing behaviour rather than removing a panic, which is exactly what a 2,400-line pass skims. From 442c onward it is one module group per PR, which also lets the unchanged-behaviour claim be made per file instead of in aggregate.

PR #457, opened 2026-08-29
Guard the public API with a checked-in snapshot (#455)

### LINKED ISSUE #455: semver CI passes a changed public return type
## Overview

The `semver` CI job does not catch a changed return type on a public function,
so the gate that is supposed to protect the published API silently passes a
breaking change.

Observed on PR #452. `curves::create_linear_curve` and
`curves::create_constant_curve` changed from `-> Curve` to
`-> Result<Curve, CurveError>`. Both are public API: `src/lib.rs:1260` has
`pub mod curves`, and `src/curves/mod.rs:13` re-exports them with
`pub use utils::{create_constant_curve, create_linear_curve}`. Every external
caller breaks and has to add `?` or `.expect(..)`.

`cargo-semver-checks` reported:

```
Checking optionstratlib v0.20.0 -> v0.20.0 (no change; assume minor)
Summary no semver update required
```

The job went green.

## Why it matters

The break itself was deliberate and approved — the two constructors aborted the
caller's process on inputs outside the `Decimal` range, and the crate is
pre-1.0, which is when that break is cheap. What is not acceptable is that the
gate did not see it. The plan for #452 was written expecting the check to flag
it; if it can miss this, it can miss the next one that was not deliberate, and
the crate ships a breaking change in a patch release.

This is the same shape as the `scan-banned` defect found in #441: a check that
returns green while covering far less than its name suggests.

## Context

- `.github/workflows/semver.yml` — `obi1kenobi/cargo-semver-checks-action@v2`,
  `feature-group: all-features`
- The baseline is the published 0.20.0 and the working version is also 0.20.0,
  so the tool assumes a minor bump — under which a breaking change must be
  reported, and was not
- Worth confirming against the current `cargo-semver-checks` lint set whether a
  return-type change is covered at all, and at which version

## Tasks

- [ ] Reproduce locally: run `cargo semver-checks` against the published
      baseline on the #452 branch and confirm it reports nothing
- [ ] Determine whether the lint exists and is disabled, or does not exist
- [ ] If it does not exist, add a public-API snapshot to CI — `cargo public-api`
      diffing against the published release is the usual answer — so a signature
      change has to be acknowledged in the diff rather than discovered by a user
- [ ] Re-run the guard against #452's commit and confirm it flags the two
      constructors

## Acceptance criteria

- [ ] A changed return type on a public function fails CI, proven against
      #452's actual commit
- [ ] The guard covers the whole public surface, not just the items someone
      remembered to list
- [ ] A deliberate break is acknowledged explicitly rather than passing silently

joaquinbejar:
## Summary

The `semver` job passed #452, which changed `curves::create_linear_curve` and
`create_constant_curve` from `-> Curve` to `-> Result<Curve, CurveError>`. Both
are public — `pub mod curves` in `src/lib.rs`, re-exported from
`src/curves/mod.rs` — so every external caller breaks, and the gate whose whole
job is to catch that reported `no semver update required`.

The break was deliberate and approved. The gate missing it was not.

## The cause

It is a coverage gap, not a misconfiguration. Reproduced locally against the
published 0.20.0 baseline, on the installed 0.48.0 and again on 0.50.0 (what
the unpinned action actually installs):

```
Checking optionstratlib v0.20.0 -> v0.20.0 (no change; assume minor)
196 checks: 196 pass, 58 skip
Summary no semver update required
```

Of the 257 lints in 0.50.0, none covers a function's return type changing in
the general case. The only return-type lints are about `()` specifically —
`function_now_returns_unit`, `exported_function_return_value_added`,
`trait_method_return_value_added`. Upstream confirms it as a known gap:
[obi1kenobi/cargo-semver-checks#1613](https://github.com/obi1kenobi/cargo-semver-checks/issues/1613),
where the maintainer answers "This is a known issue." No feature-group or
configuration change would have caught it.

So the answer is an additional check, not a setting.

## Changes

- `public-api/optionstratlib.txt` — a checked-in snapshot of the entire public
  surface, generated with `--all-features` to match the semver job's
  feature group.
- `make public-api-check` — fails when the crate's public API drifts from the
  snapshot, printing the diff and telling the author how to acknowledge it.
  Shaped like the existing `scan-banned` target.
- `make public-api-update` — regenerates the snapshot. This is the
  acknowledgement mechanism: an intentional API change is committed together
  with its snapshot diff, so it lands in the review rather than in a user's
  build.
- `.github/workflows/public_api.yml` — one job, mirroring `lint.yml`.
  `semver.yml` is untouched; this is additive.

## Technical decisions

**Why a snapshot rather than a lint.** The gap is upstream and open. A snapshot
covers the whole surface by construction rather than a list somebody has to
remember to extend, and it makes a deliberate break visible as a diff instead
of silenceable with a flag.

**Why `-sss`.** Blanket, auto-trait and auto-derived impls otherwise drag in
dozens of unreviewable lines per public type from `nalgebra`, `utoipa`,
`zerocopy` and `tower_http`. Nothing is lost by omitting them: a dropped `Send`
or `Debug` is caught by `cargo-semver-checks`' own `auto_trait_impl_removed`
and `derive_trait_impl_removed` lints, which work correctly — this PR only
covers what that tool cannot see.

**Nightly.** `cargo-public-api` builds rustdoc JSON, which requires nightly. It
does not need to be the active toolchain, so `rust-toolchain.toml` stays pinned
to stable and the workflow installs nightly alongside it. The tool itself is a
CLI installed in the job, exactly as `cargo-readme`, `cargo-tarpaulin` and
`cargo-criterion` already are. No crate dependency is added.

## Testing

Verified in both directions, because a guard that cannot fail is worse than no
guard:

- **Fails on the real break.** With the snapshot taken from `6e196e63` — the
  commit before the break — placed against `e7f28ced`'s source, the check fails
  with exactly the two changed signatures and nothing else, confirmed against
  `cargo public-api diff 6e196e63..e7f28ced`.
- **Fails on an added item.** A `pub fn` added to the crate root is reported by
  name.
- **Passes clean** on the committed snapshot.

- [ ] Unit tests added or updated — not applicable, this is CI tooling
- [x] `make public-api-check` verified locally in both directions
- [x] `cargo fmt --all --check`, `cargo clippy --all-targets --all-features
      --workspace -- -D warnings` and `make scan-banned` unaffected and clean

## Public API impact

None. No file under `src/` is touched. The snapshot committed here already
contains the `Result` signatures from #452, so this PR's own diff to it is the
record of that break.

## Checklist

- [x] No new dependencies added — `cargo-public-api` is a CLI, not a crate dependency
- [x] `semver.yml` left intact; the new check is additional
- [x] The snapshot's regeneration command is documented in the Makefile comment

## Stack

- Base branch: `main`
- Independent of the other open work: it touches only `Makefile`,
  `.github/workflows/` and the new `public-api/` directory.

Closes #455


Makefile
```diff
@@ -127,6 +127,49 @@ scan-banned:
 	fi; \
 	echo "OK: no .unwrap()/.expect() in production code"
 
+# Installs the tooling `public-api-check`/`public-api-update` need: the
+# `cargo public-api` subcommand, plus a nightly toolchain to build rustdoc
+# JSON from (it does not need to be the active/default toolchain, so it
+# coexists with the `stable` channel pinned in rust-toolchain.toml).
+.PHONY: check-cargo-public-api
+check-cargo-public-api:
+	@command -v cargo-public-api > /dev/null || (echo "Installing cargo-public-api..."; cargo install cargo-public-api --locked)
```
joaquinbejar: 🟠 **Warning**: This accepts any `cargo-public-api` already on `PATH` and otherwise installs whatever release is current. Because the CLI owns the snapshot format—and CI caches `~/.cargo/bin` only by `Cargo.lock`—two clean environments can render the same source with different versions. Pin a specific CLI version, verify or reinstall when the installed version differs, and include that version in the cache key.

.github/workflows/public_api.yml
```diff
@@ -0,0 +1,59 @@
...
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: stable
+
+      # cargo-public-api builds rustdoc JSON, which requires nightly. It does
+      # not need to be the active toolchain (rust-toolchain.toml keeps
+      # `stable` selected for everything else), it just needs to be
+      # installed so `cargo public-api` can find and use it.
+      - name: Install a nightly toolchain for rustdoc JSON
+        run: rustup toolchain install nightly --profile minimal
```
joaquinbejar: 🟠 **Warning**: `nightly` is a moving compiler, yet its rustdoc JSON is the input to this checked-in snapshot. The same commit can therefore start producing a different result—or become incompatible with the installed CLI—after a nightly update. Install a dated compatible nightly and invoke `cargo +nightly-YYYY-MM-DD public-api` consistently from both the update and check paths.

joaquinbejar (commented):
The snapshot catches the missed signature changes and the check passes, but its producer is currently nondeterministic: both `cargo-public-api` and the nightly rustdoc toolchain float across environments. Please pin and enforce both inputs so an unchanged commit cannot produce a different snapshot. REQUEST_CHANGES. GitHub does not allow the PR author to set that formal review state, so this review is submitted as COMMENT.

Makefile
joaquinbejar: Pinned it to 0.52.0, and the target now compares the installed version instead of accepting whatever is on PATH, so a cache restore that hands back an older binary gets reinstalled. The version is part of the CI cache key too.

.github/workflows/public_api.yml
joaquinbejar: Pinned the toolchain to nightly-2026-08-28, and both public-api-update and public-api-check now go through cargo +nightly-2026-08-28 public-api. CI reads that value straight from the Makefile so it cannot drift from a local run.

joaquinbejar:
Good catch, both producers are pinned now. cargo-public-api sits at 0.52.0 and the rustdoc toolchain at nightly-2026-08-28, the Makefile reinstalls the CLI when the installed version differs, and the workflow reads both pins from the Makefile so CI and a local run cannot diverge. Regenerating the snapshot with the pinned pair produced no diff, so the checked-in file stands.

joaquinbejar (commented):
The pinned CLI and dated nightly now make the snapshot reproducible, the workflow consumes the same pins and the local public API check passes cleanly. The previous findings are resolved. APPROVE. GitHub cannot record a formal approval from the PR author, so this review is submitted as COMMENT.

PR #458, opened 2026-08-29
Make the chain internals, simulation and series panic-free (#456)

### LINKED ISSUE #456: Panic sweep: chains internals, simulation and series
## Overview

Third module group of the sweep opened in #442, after #444 took `src/pricing/`
and #446 took curves and surfaces. This one covers the chain internals beyond
`build_chain`, plus simulation and series: roughly 85 raw divisions, 22
`.sqrt()`, 6 `.exp()` and the indexing in the walk and step kernels.

Same reasoning as the two before it. `rust_decimal` and `positive` abort on the
operator path, and the inputs here come straight from user data — a chain row
parsed from a file, a walk seeded with caller-supplied parameters — so the
extremes are reachable without anything exotic.

## Scope

- `src/chains/` beyond `build_chain`: `optiondata.rs`, `rnd.rs`,
  `generators.rs`, `utils.rs`
- `src/simulation/`: the walk and step kernels, `walk_driver.rs`
- `src/series/`

## Method

The one that worked in #444 and #446:

1. A probe crate with a path dependency on the repo, calling each public entry
   point inside `std::panic::catch_unwind` across the `Decimal` extremes and
   this module's own degeneracies: an empty chain, a single strike, a walk of
   zero steps, a series with one point, duplicated timestamps.
2. Fix in order — reject the meaningless input, return the defined limit, then
   checked arithmetic. Reuse `d_add` / `d_sub` / `d_mul` / `d_div` / `d_exp` /
   `d_ln` / `d_powd` / `d_sqrt` from `src/model/decimal.rs` and the `checked_*`
   family on `Positive`.
3. A proptest beside `tests/property/pricing_panic_freedom_test.rs`.

## Carried over from #439

`OptionData::apply_spread` is applied more than once per row:
`calculate_prices` applies it once per side, and `OptionChain::build_chain`
applies it again. It was invisible while the mid arm re-derived both sides from
the stored mid, but the pipeline should apply a spread once. Fix it here, and
say what the resulting quotes are for a sub-tick row.

## Tasks

- [ ] Probe every public entry point in the three modules
- [ ] Convert the panicking arithmetic and the unchecked indexing
- [ ] Apply the spread once per row
- [ ] Add the proptest
- [ ] Keep numerical results identical for well-formed inputs, and state that
      claim per file rather than in aggregate

## Acceptance criteria

- [ ] Existing tests pass unedited
- [ ] Failures surface as `ChainError` / `SimulationError` / `DecimalError`
      `make scan-banned` clean

## References

- #442 umbrella; #444 and #446 for the pattern
- `.claude/agents/panic-hunter.md`

joaquinbejar:
## Summary

A probe over the public API of the chain internals, simulation and series found
**24,691 panics in 71,356 calls**. It now finds **2**, and both are deliberate:
the `impl Index` operator on `RandomWalk` and `Simulator`, which keep the
`Vec::index` contract — defensible now that `get_step` and `get_random_walk`
return `Option` for a caller who does not want to risk it.

Most of what it found was not extreme input.

- **`heston_walk` iterated `0..params.size - 1`.** A zero step count underflows
  `usize`. In debug that panics; in release it is `0..usize::MAX` — a hang, not
  a crash, which is the failure mode nobody watches a test suite for.
- **`RNDStatistics` divided by `std_dev.powi(4)`.** Densities at `100` and
  `100.0000000000001` give a standard deviation near `7.1e-14`, whose fourth
  power is about `2.5e-53`. That is below `Decimal`'s scale-28 floor, so it
  rounds to exactly zero and the next line divides by it. An index chain near
  expiry looks exactly like that.
- **`OptionData`'s `Display` aborted** on a gamma at the top of the range, so
  printing a chain row inside a log line could take the process down.
- **`calculate_rnd` bumped a strike with `k + h`**, which aborts when the
  bumped strike leaves the `Positive` range. It appeared twice, once inside a
  `#[cfg(test)]` `debug!`, so a test build would have aborted even after
  production was fixed. A strike that cannot be bumped has no upper wing and
  is skipped, exactly as a missing quote already is.
- **The quote spread was applied three times per row.** `calculate_prices`
  called `apply_spread` once per side, and since `apply_spread` widens *both*
  sides on every call, the put arm re-widened the call mid the call arm had
  just stored. `build_chain` then applied a third, at a hardcoded scale of 2
  rather than the configured one.

## Changes

Every panicking operator in `src/simulation/`, `src/chains/optiondata.rs`,
`src/chains/rnd.rs`, `src/chains/utils.rs`, `src/series/` and the two chain
functions that were still raw (`calculate_skew`, `calculate_rnd`) now goes
through the checked helpers in `src/model/decimal.rs`.

Two decisions worth calling out because clamping would have been easier and
wrong:

- `ExitPolicy` thresholds decide from the **sign of the unrepresentable
  product** — the sign of a product is the product of the signs, which is
  exact — rather than clamping the threshold and firing an exit at a price
  nobody set.
- The kurtosis returns zero at the representable-scale limit, which is the
  same answer the function's own `variance == ZERO` guard already gives for a
  point mass. It is that guard reached at the floor, not a new invention.

`apply_spread` now applies once, at the scale the build params carry. No
well-formed row moves — the same chain still quotes `630.07 / 630.09` on its
in-the-money strikes — but a sub-tick row goes from `0.01 / 0.02` to
`0.01 / 0.01`, because half a spread there is exactly one tick and the old ask
was the second pass widening a mid the first had already lifted. Whether
`widen_around` should guarantee `ask > bid` is a wire-semantics decision, not a
panic fix, and is not made here.

## Breaking changes

Stated here rather than left to CI, because `cargo-semver-checks` does not
catch a changed return type — that is #455, whose guard is in flight as #457.
**Once #457 merges, this branch needs `make public-api-update` before it can
go in.**

- `chains::utils::strike_step` → `Result<Positive, ChainError>`
- `chains::RNDStatistics::new`, `chains::RNDResult::new` → `Result<Self, ChainError>`
- `RandomWalk::{get_step, get_step_mut}` → `Option<&Step<X, Y>>` / `Option<&mut …>`
- `Simulator::{get_random_walk, get_random_walk_mut}` → `Option<…>`
- `SimulationStats::update` → `Result<(), SimulationError>`
- `volatility::generate_ou_process` → `Result<Vec<Positive>, SimulationError>`

None had an honest infallible answer. `strike_step` overflows in *both*
directions depending on which input is extreme — with `k = MAX, σ = 0` the true
product is zero, with `k = 4, σ = MAX` it is enormous — so there is no limit to
return and anything infallible would be invented.

## Scope reached outside these modules, deliberately

`utils::calculate_log_returns` (792 panics) and `volatility::generate_ou_process`
(781) are not in this module group, but they are called by walk kernels that
are, and a sweep cannot claim panic-freedom for a caller while the callee
aborts. Fixing them is what let the proptest cover **all eleven** `WalkType` variants
and the full zero-to-`MAX` price range instead of excluding some with a
comment. There is no hedging note left in the file, because there is nothing
left to hedge.

## Testing

- [x] Reference regression test — the existing tests pin the numbers:
      `strike_step` still returns 5.0, 10.0 and 25.0 on its three cases,
      `RNDStatistics` still asserts `mean == 101` and skewness to 1e-5, and
      `build_chain` still produces 51 strikes at 630.09 / 630.07
- [x] `cargo test --all-features --workspace`, clippy `-D warnings`,
      `cargo fmt --all --check` and `make scan-banned` clean

`tests/property/chains_panic_freedom_test.rs` adds 8 properties at 256 cases
each, covering the quote accessors, `apply_spread` applied twice,
`calculate_prices`, `adjust_volatility`, the eight walk kernels, both drivers,
`Xstep` advance and rewind, and `expanding_window_vols`.

**Two existing tests asserted contracts this PR removes** and were rewritten
rather than deleted:

- `test_random_walk_get_step_out_of_bounds` was `#[should_panic(expected =
  "index out of bounds")]`; it now asserts `None` for the shared *and* the
  mutable accessor, which it did not cover before. The panicking contract still
  has its home in `test_random_walk_index_operator`.
- `tests_log_returns::test_zero_price` was `#[should_panic]`; it now asserts the
  `Err` that the function's own documentation always promised.

Suite: **3934 lib, 423 integration, 39 property, 207 doc — 0 failures.**

## Checklist

- [x] `# Errors` on every function whose error set changed
- [x] Checked arithmetic only — no `saturating_*` / `wrapping_*` in financial
      math. The one `saturating_sub` is on a `usize` step count, not a value,
      and is commented as such
- [x] No new dependencies

## Stack

- Base branch: `main`
- Part of #442. `src/pricing/` was #445, curves and surfaces #452; the
  remaining groups are `strategies/` and the close-out.

Closes #456


src/chains/rnd.rs
```diff
@@ -359,39 +420,67 @@ impl RNDStatistics {
...
         // a panic if a future invariant change ever produces NaN-like input.
         let std_dev = variance_dec.sqrt().unwrap_or(Decimal::ZERO);
-        let std_dev_4 = std_dev.powi(4);
+        // `sigma^4` underflows to exactly zero once sigma drops below about
+        // 1e-7 — two strikes a hair apart is enough — and dividing by it
+        // aborted with `Division by zero`. A distribution that narrow is the
+        // point mass the `variance == ZERO` guard above already answers with
+        // zero excess kurtosis; this is the same degenerate case reached at a
+        // representable-scale limit rather than exactly.
+        let Some(std_dev_4) = std_dev.checked_powi(4).filter(|value| !value.is_zero()) else {
```
joaquinbejar: 🔴 **Critical**: This returns zero for a narrow but nondegenerate distribution, which silently corrupts the tail metric. With equal mass at 100 and 100.0000000000001, the standardized deviations are ±1 and excess kurtosis is -2 even though σ⁴ underflows in Decimal. Normalize each deviation by `std_dev` first and accumulate `normalized_diff⁴`, as the skewness path already does, instead of treating underflow as a point mass.

src/simulation/stats.rs
```diff
@@ -69,9 +71,21 @@ impl SimulationStats {
...
+    ///
+    /// Returns [`SimulationError::Decimal`] when the running P&L total leaves
+    /// the representable `Decimal` range. Skipping the addition instead would
+    /// leave every later reader looking at a total that is quietly wrong,
+    /// which is worse than the abort this replaces; the run is reported and
+    /// the caller decides.
+    pub fn update(&mut self, result: SimulationResult) -> Result<(), SimulationError> {
         self.total_simulations += 1;
-        self.total_pnl += result.pnl.realized.unwrap_or(dec!(0.0));
+        self.total_pnl = d_add(
```
joaquinbejar: 🟠 **Warning**: This fallible update runs after `total_simulations` has already advanced, so an addition error leaves the accumulator partially mutated while results, exit counts and the average remain unchanged. A `Decimal::MAX` PnL followed by any positive PnL is enough to create that inconsistent state. Compute the new total and checked counters first, then commit all fields only after every fallible operation succeeds.

src/simulation/exit.rs
```diff
@@ -305,6 +318,55 @@ impl fmt::Display for ExitPolicy {
...
+/// Both the factor and the product overflow for an extreme `pct`, and neither
+/// has an honest representable stand-in: clamping the threshold would fire the
+/// exit at a price the caller never set. What IS knowable is the SIGN of the
+/// unrepresentable value — the sign of a product is the product of the signs —
+/// and that decides the comparison on its own, because `current` is a
+/// `Decimal` and therefore sits below any threshold past `Decimal::MAX` and
+/// above any threshold past `Decimal::MIN`.
+///
+/// `add` selects `1 + pct` (a threshold above the premium) over `1 - pct`.
+fn cmp_to_scaled_premium(current: Decimal, initial: Decimal, pct: Decimal, add: bool) -> Ordering {
```
joaquinbejar, line 333: 🔴 **Critical**: This makes the percentage branches safe, but the same public entry point still panics in `FixedPrice` because `(current_premium - price).abs()` overflows for `Decimal::MIN` and any positive target. Add a checked distance helper for `FixedPrice`, treating an overflowed distance as outside the one cent tolerance, and cover this branch in the panic freedom matrix.

joaquinbejar (commented):
The panic sweep is broad and the tests pass, but it still leaves one public exit policy panic, returns a mathematically incorrect kurtosis for narrow distributions, and can partially mutate simulation statistics on an error. These need fixing before merge. REQUEST_CHANGES. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

src/chains/rnd.rs
joaquinbejar: Fixed, each deviation is standardized by std_dev before it is raised now, same as the skewness path, so nothing depends on sigma^4 surviving. Two masses 1e-7 apart come back as exactly -2 and the 1e-13 case you describe comes back around -1.91, where Decimal::sqrt itself runs out of precision, both covered by regression tests.

src/simulation/stats.rs
joaquinbejar: Right, the update is now resolve then commit, every fallible step lands in a local first and the fields are only written once they have all succeeded. The counters go through a checked_increment as well, and there is a test that feeds Decimal::MAX then a positive PnL and checks all nine fields are untouched.

src/simulation/exit.rs
joaquinbejar: Good catch, FixedPrice goes through a within_fixed_price_tolerance helper built on checked_sub now, and an unrepresentable distance counts as outside the one cent tolerance since it is further from zero than anything representable. The panic freedom matrix drives all fourteen ExitPolicy variants nested two levels deep with premiums at Decimal::MIN and MAX.

joaquinbejar:
All three are fixed in 3edaa890. The kurtosis no longer divides by sigma^4, it standardizes each deviation first like the skewness path does; SimulationStats::update resolves every fallible step into locals and commits the fields only once they have all succeeded; and the FixedPrice branch measures its distance with checked_sub, treating an unrepresentable one as outside the tolerance. I also repointed the new Errors sections at ChainError::OptionDataError, since ChainError has no DecimalError variant and those links were breaking cargo doc under the crate's deny on broken intra doc links.

joaquinbejar (commented):
The snapshot now records every public signature changed by this sweep, and the public API check passes on the current head. APPROVE. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

Issue #460, opened 2026-08-29
Panic sweep: strategies, P&L, risk, metrics and backtesting


joaquinbejar:
## Overview

Fourth module group of the sweep opened in #442, after #444 (`pricing/`), #446
(`curves/`, `surfaces/`) and #456 (chain internals, simulation, series). This is
the composition layer and the largest group left: roughly 165 raw divisions,
plus the two remaining `panic!` sites in production code.

The pattern from the three sweeps before it holds. The panics that matter are
not the exotic ones — they are the ordinary inputs nobody guarded: a strategy
with no lower break-even, a P&L with a zero cost basis, a metric over an empty
position set.

## Scope

- `src/strategies/` — every strategy, the optimizer and the delta-neutral
  helpers, including the two `panic!` in `base.rs`
- `src/pnl/`
- `src/risk/`
- `src/metrics/`
- `src/backtesting/`

## Method

Same as #444, #446 and #456:

1. A probe crate calling each public entry point inside
   `std::panic::catch_unwind`, across the `Decimal` extremes and this layer's
   own degeneracies: a strategy with zero quantity, a position with no
   premium, an empty leg set, break-evens at zero, a backtest over an empty
   series.
2. Fix in order — reject the meaningless input, return the defined limit, then
   checked arithmetic. Reuse `d_add` / `d_sub` / `d_mul` / `d_div` / `d_exp` /
   `d_ln` / `d_powd` / `d_sqrt` from `src/model/decimal.rs`, the `checked_*`
   family on `Positive`, and `lower_break_even` / `price_gap` already in
   `src/strategies/base.rs`.
3. A proptest beside `tests/property/chains_panic_freedom_test.rs`.

## Tasks

- [ ] Probe every public entry point in the five modules
- [ ] Convert the panicking arithmetic and the unchecked indexing
- [ ] Resolve the two `panic!` sites, each becoming a typed error or keeping
      the panic behind a `// scan-banned: allow` marker with the argument
- [ ] Add the proptest
- [ ] State the unchanged-behaviour claim per file, not in aggregate

## Acceptance criteria

- [ ] The probe reports zero panics, or only ones with a stated contract
- [ ] Existing tests pass unedited, except any that assert a removed contract,
      which are rewritten rather than deleted
- [ ] `cargo clippy --all-targets --all-features --workspace -- -D warnings`,
      `cargo fmt --all --check`, `cargo test --all-features --workspace`,
[...]

PR #461, opened 2026-08-29
Take the Asian moments from the integral, not a shortcut (#454, #447)

### LINKED ISSUE #447: Asian zero-volatility branch prices the terminal forward, not the average
## Overview

`asian_black_scholes` prices the average, but its `sigma == 0` short-circuit
prices the terminal spot. The two disagree by the whole difference between the
terminal forward and the average forward, which is not small: at `S = 100`,
`b = 0.05`, `T = 0.5`, `K = 100` the branch uses `S·e^{bT} = 102.53` where the
formula it short-circuits uses `M1 = S(e^{bT} - 1)/(bT) = 101.26`, so a
zero-volatility call comes out at an intrinsic of 2.53 instead of 1.26 — twice
the correct value.

The geometric branch has the same defect: Kemna-Vorst's own limit is
`S·e^{bT/2}`, not `S·e^{bT}`.

Found by `quant-expert` while reviewing #445, which added a `deterministic_price`
helper for the *other* degenerate paths (a non-positive `M1`, a collapsed
matched variance). Those new call sites pass `M1` and are correct; this is the
pre-existing branch beside them, and the inconsistency is now visible in the
same file.

## Context

- `src/pricing/asian.rs:236-247` — the arithmetic `sigma == 0` branch
- `src/pricing/asian.rs:98-125` — the geometric branch
- `src/pricing/asian.rs:449-465` — `deterministic_price`, the shape the fix
  should reuse
- The formula both branches short-circuit ends in Black on `F = M1`
  (`asian.rs:400-409`)

## Tasks

- [ ] Pass the average forward `M1` to the arithmetic `sigma == 0` branch
- [ ] Pass `S·e^{bT/2}` to the geometric branch
- [ ] Reuse `deterministic_price` rather than open-coding the intrinsic again
- [ ] Add a continuity test: the price at `sigma = 1e-12` and at `sigma = 0`
      must agree to a tolerance, which is what fails today

## Acceptance criteria

- [ ] A zero-volatility Asian call at `S = 100, b = 0.05, T = 0.5, K = 100`
      prices at the average-forward intrinsic, not the terminal-forward one
- [ ] The zero-volatility price is continuous with the limit from above for
      both the arithmetic and the geometric variant
- [ ] Existing reference tests unchanged

### LINKED ISSUE #454: Driftless Asian uses the wrong second moment
## Overview

The driftless branch of the arithmetic Asian uses the wrong second moment.
`src/pricing/asian.rs:286` short-circuits `|b| < 1e-10` to
`M2 = S^2 e^{sigma^2 T}`, but the `b -> 0` limit of the integral the function
actually implements is

    M2 = (2 S^2 / (sigma^2 T^2)) * [ (e^{sigma^2 T} - 1) / sigma^2 - T ]

At `sigma = 0.25`, `T = 0.5` that is `1.0111 S^2` against the branch's
`1.0317 S^2`. The moment-matched volatility is therefore too high, and every
contract with `r = q` — a forward-priced or fully-carried underlying, not an
edge case — is mispriced.

This is the third removable boundary of the Turnbull-Wakeman second moment. The
other two, at `b = -sigma^2` and `b = -sigma^2 / 2`, were fixed in #445 by
returning to the defining double integral; this one predates that work and was
left alone because correcting it changes prices for every `r = q` contract,
which was outside what a panic sweep should decide.

## Context

- `src/pricing/asian.rs:286` — the `|b| < 1e-10` branch
- `src/pricing/asian.rs:285` onwards — the two boundaries fixed in #445, and
  the comment deriving them from `E[A^2] = (2 S^2 / T^2) ∫₀^T e^{bu} ∫₀^u
  e^{at} dt du`, which is where this limit comes from too
- Related: #447, the same file's zero-volatility branch pricing the terminal
  forward instead of the average

## Tasks

- [ ] Replace the driftless `M2` with the limit above, derived the same way as
      the other two boundaries
- [ ] Check the geometric branch for the same defect
- [ ] Add a continuity test against the general branch either side of `b = 0`,
      matching the tolerance used for the other two boundaries
- [ ] Record the price change for a representative `r = q` contract in the
      changelog, since this moves a currently-returned value

## Acceptance criteria

- [ ] The driftless price is continuous with the general branch as `b -> 0`
      from both sides
- [ ] `M2 -> S^2` as `sigma -> 0` still holds
- [ ] The moment-matched volatility at `sigma = 0.25, T = 0.5` matches the
      integral rather than the shortcut
- [ ] The price change for `r = q` is stated explicitly with a before and after

joaquinbejar:
## Summary

Three branches of the Asian pricing short-circuited to an expression that is
not the limit of the formula they short-circuit. This closes both #454 (the
driftless second moment) and #447 (the zero-volatility forward), because they
land on the same lines and splitting them would have meant rebasing one onto
the other for no reviewer benefit.

**The driftless second moment used `M2 = S² e^{σ²T}`.** The `b → 0` limit of
the integral the function actually implements,

```
E[A²] = (2 S² / T²) ∫₀^T e^{bu} ∫₀^u e^{at} dt du
```

is `(2 S² / (σ²T²)) · [ (e^{σ²T} − 1)/σ² − T ]`, derived the same way #445
derived the other two removable boundaries: evaluate the degenerate exponential
rather than divide by its vanishing rate.

The gap is not cosmetic. At `σ = 20%`, `T = 1` the old branch handed the moment
matching `σ_adj = 0.20` where the correct value is `0.11566` — **it was pricing
an arithmetic Asian at full spot volatility**. Every contract with `r = q` was
affected, which is a forward-priced or fully-carried underlying, not an edge
case.

**The zero-volatility short-circuits had the same shape of error in a different
variable.** Both returned the terminal forward `S e^{bT}`. The arithmetic
average of a deterministic path is `M1`; the geometric average is `S e^{bT/2}`.
The geometric branch disagreed with its own Kemna-Vorst legs everywhere except
exactly at `b = 0` — the cleanest evidence available that it was wrong rather
than a different convention.

## Prices moved

| contract | before | after |
|---|---|---|
| `S=K=100, T=1, σ=20%, r=q=4%` | 7.6532330880 | **4.4308753052** |
| `S=K=100, T=0.5, σ=25%, r=q=5%` | 6.8693005996 | **3.9746021865** |
| `σ=0, S=K=100, T=1, r=5%, q=0` geometric | 4.8770575499 | **2.4080487528** |
| same, arithmetic | 4.8770575499 | **2.4182085485** |

Every "before" figure was measured from the crate at `HEAD`, not reconstructed.
Both entries are in `CHANGELOG.md` under `[Unreleased]`.

## What is unchanged, and why that is structural

`M1`'s own driftless branch is correct: `(e^{bT} − 1)/(bT) → 1`, so `M1 → S`.
It is unchanged in value, and is now extracted into a shared
`arithmetic_average_forward` helper so the zero-volatility branch reuses the
expression instead of duplicating a guard.

The Kemna-Vorst legs have no defect: `b` enters only through
`b_adj = (b − σ²/6)/2`, which is linear, so `b = 0` is an ordinary point and
the function is analytic there.

**The general `else` branch is untouched.** Prices away from these boundaries
are therefore unchanged by construction rather than by testing, and the
existing reference tests pass unedited.

For `σ < 1e-14` the variance underflows `Decimal`'s scale; the bracket is
`σ²T²/2 + O(σ⁴)`, so the nested limit is exactly `S²` and the branch returns
it rather than dividing by the variance that vanished.

## Verification

Three independent ways, because a limit that is only exercised is not proven:

1. **Series limit of the closed form** — `(e^{bT} − 1)/b → T`, `a, c → σ²`
   gives the same expression, and the `else` branch was confirmed to be
   algebraically the reference bracket, so it is the right thing to take the
   limit of.
2. **Quadrature on the double integral** (40 digits): `M2 = 1.0134677405 S²`
   at `σ = 20%, T = 1`, price `4.4308753049687568`. The crate now returns
   `4.4308753052434734` — agreement to **2.7e-10**. This is the check that does
   not assume the closed form is right.
3. **Continuity against the general branch either side of the boundary**:

| shift in `r` | gap after | gap before |
|---|---|---|
| −1e-9 | 2.152e-8 | 3.2224 |
| +1e-9 | 2.097e-8 | 3.2224 |
| −1e-10 | 2.400e-9 | 3.2224 |
| +1e-10 | 1.850e-9 | 3.2224 |

The stated tolerance is `1e-7`, which clears the genuine `dP/dr ≈ 21`
sensitivity with room. The test also asserts monotonicity in `r` across the
boundary.

## Testing

- [x] Reference regression test — the existing Asian reference tests pass
      **unedited**; no assertion was changed
- [x] `cargo test --all-features --workspace` — 3942 lib, 423 integration,
- [x] `cargo fmt --all --check`, `cargo clippy --all-targets --all-features
      --workspace -- -D warnings`, `make scan-banned` and `make
      public-api-check` clean

## Public API impact

None. No signature changes; the new helper is private.

## Stack

- Base branch: `main`
- Independent of #460 (the `strategies/` sweep), which touches no pricing file.

Closes #454
Closes #447

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Every arithmetic Asian option with `r = q` was mispriced** (#454). The
+  Turnbull-Wakeman second moment short-circuited its removable `b = 0`
+  singularity to `M2 = S² e^{σ² T}`, which is `E[S_T²]`: the second moment of
+  the *terminal* price, not of its average. The `b → 0` limit of the double
+  integral the function actually implements is
+  `M2 = (2 S² / (σ² T²)) [ (e^{σ² T} − 1) / σ² − T ]`. The stand-in handed the
+  moment matching `σ_adj = σ` where the average carries `σ_adj ≈ σ / √3`, so
+  the price came out far too high and the branch was discontinuous: a one-year
+  ATM call on `S = K = 100`, `σ = 20%`, `r = q = 4%` returned
+  **7.6532330880**, against `4.4308752837` and `4.4308753262` for a carry a
+  billionth either side. It now returns **4.4308753052**, which agrees with
+  quadrature on the defining integral to `3e-10` and with both neighbours to
+  `2.2e-8`. A forward-priced or fully-carried underlying is an ordinary
+  contract, so every pinned `r = q` Asian value moves.
+- **A zero-volatility Asian option priced off the terminal forward instead of
+  the average** (#454). Both kernels short-circuited `σ = 0` to
+  `(S e^{bT} − K)⁺ e^{−rT}`, but a deterministic path still has to be
+  averaged: the geometric mean of `S e^{b t}` over the window is `S e^{bT/2}`
+  and the arithmetic mean is `S (e^{bT} − 1) / (bT)`, which are also the
+  `σ → 0` limits of the Kemna-Vorst and Turnbull-Wakeman formulas the branches
+  stand in for. A one-year ATM call on `S = K = 100`, `r = 5%`, `q = 0`
+  returned **4.8770575499** from both kernels and now returns
+  **2.4080487528** (geometric) and **2.4182085485** (arithmetic). The old
+  value was only correct at `b = 0`, where every average collapses to `S`.
 - **Reachable panic in the lower break-even of eight strategies.**
   `strike - credit` was computed with the raw `Positive - Decimal` operator,
   which panics when the credit (or, on a long structure, the debit) exceeds the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `2.2e-8`. A forward-priced or fully-carried underlying is an ordinary
   contract, so every pinned `r = q` Asian value moves.
 - **A zero-volatility Asian option priced off the terminal forward instead of
-  the average** (#454). Both kernels short-circuited `σ = 0` to
+  the average** (#447). Both kernels short-circuited `σ = 0` to
   `(S e^{bT} − K)⁺ e^{−rT}`, but a deterministic path still has to be
   averaged: the geometric mean of `S e^{b t}` over the window is `S e^{bT/2}`
   and the arithmetic mean is `S (e^{bT} − 1) / (bT)`, which are also the
--- a/src/pricing/asian.rs
+++ b/src/pricing/asian.rs
@@ -51,9 +51,9 @@ use rust_decimal_macros::dec;
 /// - [`PricingError::Decimal`] when an intermediate step leaves the
 ///   representable `Decimal` range: the discount factor, the forward, either
 ///   Turnbull-Wakeman moment, the moment-matched variance, or the final
-///   Black legs. The two removable singularities of the Turnbull-Wakeman
-///   second moment, `b = -σ²` and `b = -σ²/2`, are evaluated at their limits
-///   and never raise.
+///   Black legs. The three removable singularities of the Turnbull-Wakeman
+///   second moment, `b = 0`, `b = -σ²` and `b = -σ²/2`, are evaluated at
+///   their limits and never raise.
 /// - [`PricingError::Positive`] when the `σ / √3` geometric adjustment is not
 ///   representable as a `Positive`.
 pub fn asian_black_scholes(option: &Options) -> Result<Decimal, PricingError> {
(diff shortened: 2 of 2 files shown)


src/pricing/asian.rs
```diff
@@ -291,22 +278,82 @@ fn arithmetic_asian_price(option: &Options) -> Result<Decimal, PricingError> {
...
+        } else {
+            let exp_var_t = d_exp(
+                d_mul(
+                    sigma_sq,
+                    t_dec,
+                    "pricing::asian::arithmetic::m2::b_limit::var_t",
+                )?,
+                "pricing::asian::arithmetic::m2::b_limit::exp_var_t",
+            )?;
+            let bracket = d_sub(
```
joaquinbejar, line 346: 🔴 **Critical**: This nested subtraction loses the second order term while `sigma_sq` is still representable. For `S = K = 100`, `T = 1`, and `r = q = 4%`, `sigma = 1e-7` returns `31.0136012069` and `sigma = 1e-6` returns zero, although both prices must converge to the zero volatility value. The `sigma_sq.is_zero()` guard does not cover cancellation in `exp(x) - 1 - x`. Evaluate `2 S² (exp(x) - 1 - x) / x²` through a stable small `x` series such as `S² (1 + x/3 + x²/12 + ...)`, and add a continuity regression below `1e-6`.

CHANGELOG.md
```diff
@@ -29,6 +29,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  moment matching `σ_adj = σ` where the average carries `σ_adj ≈ σ / √3`, so
+  the price came out far too high and the branch was discontinuous: a one-year
+  ATM call on `S = K = 100`, `σ = 20%`, `r = q = 4%` returned
+  **7.6532330880**, against `4.4308752837` and `4.4308753262` for a carry a
+  billionth either side. It now returns **4.4308753052**, which agrees with
+  quadrature on the defining integral to `3e-10` and with both neighbours to
+  `2.2e-8`. A forward-priced or fully-carried underlying is an ordinary
+  contract, so every pinned `r = q` Asian value moves.
+- **A zero-volatility Asian option priced off the terminal forward instead of
+  the average** (#454). Both kernels short-circuited `σ = 0` to
```
joaquinbejar: 🔵 **Nitpick**: This zero volatility fix closes #447, but the changelog points to #454 and sends readers to the driftless moment issue above. Change this reference to #447.

joaquinbejar (commented):
The corrected limiting formulas and boundary tests are well motivated, but the new driftless implementation becomes numerically unstable at small representable volatility and can return zero or a very large price instead of the continuous limit. This needs a stable small argument evaluation before merge. REQUEST_CHANGES. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

src/pricing/asian.rs
joaquinbejar: Fixed in 3de1d9ed, the branch sums the series 1 + x/3 + x^2/12 + ... below x = 1e-6 instead of forming exp(x) - 1 - x, so the price stays proportional to sigma down the decades. Covered by a series against closed form crossover test and a tiny volatility proportionality test.

CHANGELOG.md
joaquinbejar: Right, that entry closes #447. Fixed in 991217a0.

joaquinbejar:
Two updates, one of them a caveat on this PR's own claim.

Pushed 3de1d9ed on top. The first cut of the driftless branch was sound at ordinary volatilities and wrong below sigma of about 1e-5 — at 1e-7 a worthless contract priced at 31.01, and at 1e-13 it returned an error where the old code returned a value. `e^x` is stored to 28 decimal places, so the `x^2/2` leading `e^x - 1 - x` sinks under that floor as `x = sigma^2 T` shrinks and the closed form loses exactly the term it needs. The branch is now collected on `x` and takes a two-term series below `x = 1e-6`, a threshold derived from where the truncation error `x/4` and the cancellation error `6e-28/x^3` cross, at `x` around 2.2e-7, then rounded up one decade so the choice survives the error model being off by a few factors. The headline value did not move: sigma 20% at r = q = 4% still returns 4.4308753052434733824678445998 byte for byte, so every figure above stands.

The caveat: measuring that ladder turned up a pre-existing defect in the branch this PR deliberately leaves untouched. The general arithmetic second moment loses its precision as maturity shortens — 73% too high at 14 minutes to expiry, 612 times too high at 86 seconds — through the same cancellation, in the `T` direction instead of the `sigma` direction. This PR's claim stands exactly as written, that prices away from these boundaries are unchanged by construction, but a reader should not infer from that that they are correct in every regime. Filed as #462 with the measured table against quadrature, and not touched here.

PR #464, opened 2026-08-29
Implement the chooser formula the module documents (#448)

### LINKED ISSUE #448: Complex chooser does not match the Rubinstein formula it documents
## Overview

`chooser_black_scholes` does not implement the Rubinstein complex-chooser
formula it documents. Two things are off:

- the `y` legs discount at the choice date `t` rather than at expiry `T`
  (`chooser.rs:255-274`)
- `y1` uses `b·t` where the formula has `b·T + sigma²·t/2`
  (`chooser.rs:154-155`)

On Haug's own worked example — `S = X = 50`, `t = 0.25`, `T = 0.5`,
`r = b = 0.08`, `sigma = 0.25`, reference value **6.1071** — the file returns
**6.5241**, about 7% high.

Found by `quant-expert` while reviewing the limits added in #445. The limits
themselves are correct for the formula as written; this issue is that the
formula is not the one in the doc comment.

## Context

- `src/pricing/chooser.rs:154-155` and `:255-274`
- Haug, *The Complete Guide to Option Pricing Formulas*, complex chooser
- The `t_choice -> 0` correction added in #445 comes out undiscounted because
  of the same defect, so fixing this also fixes that limit

## Tasks

- [ ] Correct the discounting of the `y` legs to expiry
- [ ] Correct the `y1` drift term to `b·T + sigma²·t/2`
- [ ] Regress against Haug's example to 4 decimal places
- [ ] Re-check the `sigma·sqrt(t_choice) -> 0` limit added in #445 against the
      corrected formula

## Acceptance criteria

- [ ] Haug's example prices at 6.1071 within 1e-4
- [ ] `price_at_choice_equals_expiry` still holds
- [ ] The simple-chooser identity `max(C, P) = C + max(0, P - C)` holds at the
      choice date

joaquinbejar:
## Summary

The module header documents Rubinstein's simple chooser. The body did not
implement it, and the price was **6.83% high** against Haug's published
example.

Two defects, which are the same mistake seen twice:

| | before | after |
|---|---|---|
| `y1` numerator | `ln(S/K) + (b + σ²/2)·t` | `ln(S/K) + b·T + σ²·t/2` |
| `y` leg discounts | `K·e^{-rt}`, `S·e^{-qt}` | `K·e^{-rT}`, `S·e^{-qT}` |

Parity at the choice date gives
`max(C_t, P_t) = C_t + max(0, K·e^{-r(T−t)} − S_t·e^{-q(T−t)})`, so a chooser
is a `T`-expiry call plus `e^{-q(T−t)}` puts expiring at `t`, struck at the
discounted forward. The embedded put's own d-values *are* `y1`/`y2`, and its
`e^{-rt}` discount recombines with the strike adjustment `e^{-b(T−t)}` back
into `e^{-rT}`. That is why the carry runs to expiry while only the variance
runs to the choice date.

## The number

| | value |
|---|---|
| before | 6.524120579621281140947825628 |
| after | **6.107077498162327583120889761** |
| Haug, published | 6.1071 |
| independent recomputation | 6.107077498162337 |

Reference: Haug, *The Complete Guide to Option Pricing Formulas*, 2nd ed.,
§2.5.1; original result Rubinstein (1991), "Options for the Undecided". The
"before" figure was reproduced by reverting the two numeric defects in place,
not reconstructed.

## The #445 limit changes with it

The zero-diffusion limit keeps its saturation structure, but its discriminant
depends on the numerator this PR corrects. The sign test is now
`ln(S/K) + b·T` — the sign of `C − P = e^{-rT}(S·e^{bT} − K)` — rather than the
spot moneyness `ln(S/K)`.

With no diffusion left the holder chooses immediately and takes whichever
branch the forward favours, so the old test was wrong for every spot between
the strike and the forward. At `S = 99, K = 100, T = 0.5, r = b = 8%` it priced
the put at 5.5075 and now prices the call at 8.4285.

That also makes the `Ordering::Equal` error spurious. A zero numerator now
means `S·e^{bT} = K`, forward parity, where both saturations return the same
number — so the branch returns that common value instead of reporting an
indeterminacy that no longer exists. The `# Errors` doc drops the clause; no
signature change.

## Naming

This is the **simple** chooser, not the complex one, and the issue title is
wrong. `OptionType::Chooser { choice_date }` carries one strike and one
expiry, and Haug's 6.1071 is his simple-chooser example. His complex chooser
takes separate call and put strikes and expiries, needs a bivariate normal and
a root solve for the indifference spot, and cannot be expressed by this
variant at all.

## Testing

- [x] Reference regression test — `test_simple_chooser_haug_example_matches_reference`
      pins 6.1071 at 1e-4
- [x] **No existing test was re-baselined.** Nothing pinned the wrong 6.5241,
      which is why the defect survived
- [x] `cargo test --all-features --workspace` — 3945 lib, 423 integration,
      `make public-api-check` clean

New coverage: the Haug regression; the zero-horizon limit following the
forward rather than the spot, in both directions; forward parity pricing the
tie; and the `max(C, P) = C + max(0, P − C)` identity with and without
dividends, so the `e^{-q(T−t)}` scale on the embedded puts is exercised.

## Public API impact

None. No signature changes; the file stays panic-free as of #445.

## Stack

Closes #448

### DIFF
--- a/src/pricing/chooser.rs
+++ b/src/pricing/chooser.rs
@@ -627,7 +627,8 @@ mod tests {
         );
         let put_price = crate::pricing::black_scholes_model::black_scholes(&put).unwrap();
 
-        assert_decimal_eq!(chooser_price, call_price + put_price, dec!(1e-6));
+        let parity = d_add(call_price, put_price, "test::chooser::parity").unwrap();
+        assert_decimal_eq!(chooser_price, parity, dec!(1e-6));
     }
 
     /// Same identity with a dividend yield, so the `e^(-q(T-t))` scale on the
@@ -657,7 +658,9 @@ mod tests {
         let put_price = crate::pricing::black_scholes_model::black_scholes(&put).unwrap();
         let scale = f64_to_decimal((-q * gap_years).exp()).unwrap();
 
-        assert_decimal_eq!(chooser_price, call_price + put_price * scale, dec!(1e-6));
+        let scaled_put = d_mul(put_price, scale, "test::chooser::scaled_put").unwrap();
+        let parity = d_add(call_price, scaled_put, "test::chooser::parity").unwrap();
+        assert_decimal_eq!(chooser_price, parity, dec!(1e-6));
     }
 
     /// A choice horizon of zero leaves no diffusion, so `σ√t` collapses and
@@ -731,7 +734,8 @@ mod tests {
         let put = european(s, k, expiry_days, sigma, r, 0.0, OptionStyle::Put);
         let put_price = crate::pricing::black_scholes_model::black_scholes(&put).unwrap();
 
-        assert_decimal_eq!(price, call_price + put_price, dec!(1e-8));
+        let straddle = d_add(call_price, put_price, "test::chooser::straddle").unwrap();
+        assert_decimal_eq!(price, straddle, dec!(1e-8));
     }
 
     /// The closed form and the `price_at_choice_equals_expiry` branch have to
--- a/src/pricing/chooser.rs
+++ b/src/pricing/chooser.rs
@@ -11,17 +11,42 @@
 //!
 //! # Simple Chooser (Rubinstein 1991)
 //!
-//! At the choice date t, the holder chooses max(Call, Put).
+//! At the choice date t, the holder chooses max(Call, Put). Both branches
+//! share one strike K and one expiry T, which is what makes the chooser
+//! *simple*. A complex chooser carries a separate strike and expiry per
+//! branch, needs a bivariate normal and a root solve for the indifference
+//! spot, and is not implemented here.
+//!
 //! The value is:
 //!
-//! `V = S*e^(-qT)*N(d1) - K*e^(-rT)*N(d2) + K*e^(-rt)*N(-y2) - S*e^(-qt)*N(-y1)`
+//! `V = S*e^(-qT)*N(d1) - K*e^(-rT)*N(d2) + K*e^(-rT)*N(-y2) - S*e^(-qT)*N(-y1)`
 //!
 //! Where:
 //! - T = time to final expiration
 //! - t = time to choice date
 //! - d1, d2 are standard BS d-values for T
-//! - y1 = [ln(S/K) + b*t + (σ²/2)*t] / (σ√t)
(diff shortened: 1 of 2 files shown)


src/pricing/chooser.rs
```diff
@@ -521,6 +526,235 @@ mod tests {
...
+            choice_days,
+            sigma,
+            r,
+            q,
+            OptionStyle::Put,
+        );
+        let put_price = crate::pricing::black_scholes_model::black_scholes(&put).unwrap();
+        let scale = f64_to_decimal((-q * gap_years).exp()).unwrap();
+
+        assert_decimal_eq!(chooser_price, call_price + put_price * scale, dec!(1e-6));
```
joaquinbejar: 🟠 **Warning**: This assertion uses raw `Decimal` addition and multiplication, and the same unchecked addition appears on lines 630 and 734. The project rules require checked arithmetic for financial values. Build the expected value with `d_mul` and `d_add` before asserting.

joaquinbejar (commented):
The corrected chooser formula matches the published benchmark and the decomposition tests, but three new assertions use raw Decimal arithmetic in violation of the checked arithmetic rule. The branch also currently conflicts with main and must preserve the latest changelog entries when rebased. REQUEST_CHANGES. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

src/pricing/chooser.rs
joaquinbejar: Fixed, the three parity expectations are built with d_add and d_mul now instead of the raw operators. The branch is also rebased on main with both changelog entry sets kept, the Asian ones from main and the chooser ones from here.

joaquinbejar:
Both points are addressed in b1ccd67e. The three assertions build their expected value through d_add and d_mul, and the rebase onto main keeps the Asian entries that landed in #461 alongside the chooser ones from this branch.

joaquinbejar (commented):
The follow up replaces all three raw Decimal expressions with d_add and d_mul, preserves the rebased changelog, and all 19 chooser tests pass. The branch is mergeable and the build, format, lint, public API, and semver checks are green. APPROVE. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

PR #465, opened 2026-08-29
Keep the early-exercise premium at zero volatility (#449)

### LINKED ISSUE #449: Zero-volatility binomial drops the early-exercise premium
## Overview

`price_binomial` short-circuits a zero-volatility option to
`calculate_discounted_payoff`, which returns `e^{-rT}·payoff(S·e^{rT})` — the
European value. For an American or a Bermuda contract that drops the
early-exercise premium.

A zero-volatility American put at `S = 90`, `K = 100`, `r = 5%`, `T = 1` is
worth `K - S = 10` by exercising immediately. The deterministic path returns
`e^{-0.05}(100 - 90·e^{0.05})⁺ = 5.13`, roughly half.

Pre-existing (`binomial_model.rs:137-139`), not introduced by #445 — that PR
only mirrored the same branch for the newly-detected collapsed lattice, which
`quant-expert` confirmed is consistent. The defect lives on the far more
reachable `sigma == 0` input.

## Context

- `src/pricing/binomial_model.rs:137-139` — the `sigma == 0` branch
- `src/pricing/utils.rs:446-498` — `calculate_discounted_payoff`
- The collapsed-lattice branch added in #445 takes the same path, so both are
  fixed by the same change

## Tasks

- [ ] For an American or Bermuda style, return
      `max(immediate exercise, discounted European value)` in the
      zero-volatility branch
- [ ] For Bermuda, respect the exercise-date schedule rather than assuming
      continuous exercise
- [ ] Test the zero-volatility American put above against its analytic value

## Acceptance criteria

- [ ] A zero-volatility American put at `S = 90, K = 100, r = 5%, T = 1`
      prices at 10
- [ ] The European zero-volatility price is unchanged
- [ ] The price is continuous with the limit from `sigma -> 0+`

joaquinbejar:
## Summary

`price_binomial` short-circuited a zero-volatility option to the discounted
European payoff, so an **American or Bermuda contract silently lost its
early-exercise premium**. A zero-volatility American put at `S = 90`,
`K = 100`, `r = 5%`, `T = 1` is worth `K − S = 10` by exercising immediately;
it returned **5.1229**.

The collapsed-lattice guard added in #445 takes the same branch, so both are
fixed by the same change.

## The numbers

| contract (`K = 100, T = 1, σ = 0`) | before | after |
|---|---|---|
| American put, `S = 90`, `r = +5%` | 5.122942450071404 | **10** |
| American call, `S = 110`, `r = −5%` | 4.872890362397595 | **10** |
| American call, `S = 110`, `r = +5%` | 14.877057549928595 | unchanged |
| American put, `S = 90`, `r = −2%` | 12.020134002675647 | unchanged |

The call at a positive rate is unchanged because a call on a
non-dividend-paying underlying is never exercised early while rates are
positive — that is the correct answer, not a missed case. Flipping the rate
sign exercises that branch and mirrors it on the put. All four are pinned
against closed forms: `K − S` exactly for the immediate-exercise cases,
`S − K·e^{-rT}` and `K·e^{-rT} − S` at 1e-12 for the others.

## How the European path is protected

Both entries route through a private `price_deterministic`, which values each
admissible exercise time by calling the **existing** `calculate_discounted_payoff`
with `expiry` set to that time. The European path is therefore the same call
with the same arguments as before, and is bit-identical by construction rather
than by testing.

## Bermuda reads its schedule

Exercise opportunities are expiry plus every scheduled date at or before it,
read exactly rather than snapped to a lattice step. At `S = 90, K = 100,
r = 5%`, where `K·e^{-rt} − S` decays in `t` so the earliest date wins:

- `[0.25, 0.5, 0.75]` → 8.757780049388145 = `100·e^{-0.0125} − 90`, strictly
  between the European 5.1229 and the American 10
- `[0.5]` alone → 7.530991202833263, **strictly below** the `[0.25, 0.5]`
  schedule. This is the test that distinguishes a real schedule from
  continuous exercise: a branch ignoring the dates returns the same number for
  both
- `[]` and `[2.0]` (entirely past expiry) → equal the European, exactly
- `[0.0]` → equals the American, exactly

## Continuity, and a constraint worth recording

Against the limit from `σ → 0+`: largest measured gap **4.9e-15** across five
cases at `σ = 0.005`, `n = 200`, against a stated tolerance of `1e-12`. That is
`Decimal` round-off through 200 induction steps, not model error — and about
twelve orders of magnitude below the 4.88 the defect was moving the put by.

The finding that constrains that test: **a CRR lattice is only arbitrage-free
while `σ > |r|·√dt`.** Below that the risk-neutral probability leaves `[0, 1]`
and the lattice stops approximating anything. Measured outside the valid
region the American call is 1.09 out at `σ = 0.002, n = 400` and 2.36 out at
`σ = 0.001, n = 800`. Those are lattice breakdown, not a discontinuity in the
new branch, and the bound is recorded in the test's doc comment so nobody later
"tightens" the tolerance into the broken regime.

## Testing

- [x] Unit tests added — 11 in `tests_zero_volatility_early_exercise`, 5 in
      `tests_zero_volatility_continuity`
- [x] Reference regression test — every case pinned against a closed form
- [x] **No existing test touched.** None had pinned a European value for an
      American contract, so no stop was needed
- [x] `cargo test --all-features --workspace` — 3953 lib, 423 integration,
      `make public-api-check` clean

One of the 11 covers the collapsed lattice at `σ = 1e-28`, the smallest
non-zero `Decimal`, verified with a temporary probe to actually reach that
branch rather than assumed to.

## Public API impact

None.

## Stack

Closes #449

### DIFF
--- a/src/pricing/binomial_model.rs
+++ b/src/pricing/binomial_model.rs
@@ -97,7 +97,10 @@ pub struct BinomialPricingParams<'a> {
 /// # Special cases
 ///
 /// - If `expiry` is 0, the function returns the intrinsic value of the option.
-/// - If `volatility` is 0, the function calculates the option price deterministically.
+/// - If `volatility` is 0, the underlying is deterministic and the price is
+///   computed in closed form, which still honours early exercise: an American
+///   is worth the better of exercising now and holding to expiry, a Bermuda
+///   the best of its schedule, a European its discounted payoff.
 ///
 /// # Notes
 ///
@@ -135,7 +138,7 @@ pub fn price_binomial(params: BinomialPricingParams) -> Result<Decimal, PricingE
         return Ok(intrinsic_value);
     }
     if params.volatility == Decimal::ZERO {
-        return calculate_discounted_payoff(params);
+        return price_deterministic(&params);
     }
 
     let no_steps_raw = params.no_steps.get();
@@ -146,7 +149,7 @@ pub fn price_binomial(params: BinomialPricingParams) -> Result<Decimal, PricingE
         // `σ√dt` underflowed below the representable scale, so the lattice has
         // collapsed onto a single deterministic path: same answer as the
         // zero-volatility branch above.
-        return calculate_discounted_payoff(params);
+        return price_deterministic(&params);
     }
     let p = calculate_probability(params.int_rate, dt, d, u)?;
     let discount_factor = calculate_discount_factor(params.int_rate, dt)?;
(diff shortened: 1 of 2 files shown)


src/pricing/binomial_model.rs
```diff
@@ -224,6 +227,64 @@ pub fn price_binomial(params: BinomialPricingParams) -> Result<Decimal, PricingE
...
+            expiry: time,
+            ..params.clone()
+        })
+    };
+
+    let expiry_value = exercise_value(params.expiry)?;
+    match params.option_type {
+        OptionType::American => {
+            let immediate = exercise_value(Positive::ZERO)?;
+            Ok(expiry_value.max(immediate))
```
joaquinbejar: 🔴 **Critical**: This maximization is applied after `calculate_discounted_payoff` receives the original side. That helper signs the payoff and then negates a short result again, so this arm and the Bermuda arm return a positive liability. A zero volatility short American put with `S = 90`, `K = 100`, and `r = 5%` returns `+10` instead of `-10`. Evaluate every candidate as long, take the maximum, then apply the requested side once, and add long and short symmetry coverage.

joaquinbejar (commented):
The long side early exercise values and Bermuda schedule regressions check out, but the new deterministic maximization returns positive values for short American and Bermuda positions, so it corrupts the position sign. REQUEST_CHANGES. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

src/pricing/binomial_model.rs
joaquinbejar: Fixed in 5946b527, every candidate is valued long, the maximum is taken there and the requested side is applied once to the winner. Your case now returns -10, and long against short symmetry is covered across the American, Bermuda and European arms on both styles.

joaquinbejar:
Sign bug fixed in 5946b527. The maximization runs on long values and the side is applied once at the end, which also puts the European arm of this branch back on the long equals minus short convention the lattice holds to.

joaquinbejar (commented):
The follow up values every exercise candidate on the long side, applies the requested side once after maximization, and restores exact long and short symmetry across American, Bermuda, and European contracts. All 13 focused tests pass, the branch is mergeable, and the build, format, lint, public API, and semver checks are green. APPROVE. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

Issue #466, opened 2026-08-29
Duplicate abscissae in Curve and Surface have no defined rule


joaquinbejar:
## Overview

`Curve` and `Surface` store points in a `BTreeSet`, and duplicate abscissae get
there legitimately — projecting a grid onto one axis produces several points
sharing an `x` by construction. There is no rule for what that means, and the
crate assumes two contradictory things about it.

Most of the crate treats a `Curve` as a function of its abscissa:
`Curve::get_point` takes the first `p.x == *x`, `contains_point` is `p.x == *x`,
`Curve::merge` resamples every curve onto one shared grid with one `y` per `x`.
But `Surface::get_curve` returns a projection, which is genuinely multi-valued,
as a `Curve`.

## How this surfaced

#450 fixed the `Ord`/`Eq`/`Hash` contract on `Point2D` and `Point3D`. Before
that fix, the disagreement between `Eq` (on `x` alone) and `Ord` (on `(x, y)`)
meant a `BTreeSet` of these points **had no well-defined contents**:
`BTreeSet::from_iter` deduplicates adjacent elements with `PartialEq`, while
`BTreeSet::insert` deduplicates with `Ord`. Measured at the commit before the
fix, on the same four `Point3D` stacked on one `(x, y)`:

```
from_iter  len = 1
insert     len = 4
```

`Surface::get_curve(Axis::X)` on the test surface returned **3** of its 5
projected points through `collect`, silently keeping only the greatest ordinate
per abscissa. `(0, 0)` and `(1, 1)` were simply absent, and the tests that
probed for them passed only because `x`-only equality made `(0, 1)` match a
probe for `(0, 0)` — the same defect caused the loss and concealed it.

With the contract fixed, the projection now returns all five points. That is
the correct behaviour for a projection and the tests assert it, but it leaves
the design question open.

## The decision to make

Three coherent answers, none of which belongs in a panic or contract fix:

- **`get_curve` returns the wrong type.** A projection is multi-valued; give it
  a type that says so, and leave `Curve` as a function of its abscissa.
- **`Curve` legitimately holds duplicates**, and every consumer that assumes
  one point per abscissa — `get_point`, `contains_point`, `merge`, the
  interpolators — documents and enforces its own rule.
- **Construction normalizes**, with a stated rule for which point survives.
  Note that `points` is a `pub` field on both types, so a struct literal
  bypasses any constructor: normalization can be a convenience but never an
  invariant.

[...]

PR #467, opened 2026-08-29
Make the point types agree with themselves (#450)

### LINKED ISSUE #450: Point2D orders on (x, y) but compares equal on x alone

joaquinbejar:
## Summary

`Point2D` compared equal on `x` alone while ordering on `(x, y)`, and hashed on
`x` alone. `Point3D` compared equal on `(x, y)` while ordering on `(x, y, z)`,
and had no `Hash` at all. That breaks the contract `Ord` requires — `a == b`
iff `a.cmp(&b) == Ordering::Equal` — and both types are stored in a `BTreeSet`.

**The consequence is that a `BTreeSet` of these points had no well-defined
contents.** `BTreeSet::from_iter` deduplicates adjacent elements with
`PartialEq`; `BTreeSet::insert` deduplicates with `Ord`. Measured at `main`, on
the same four `Point3D` stacked on one coordinate:

```
from_iter  len = 1
insert     len = 4
```

## What it cost

- **`Surface::merge_indexes` collapsed every column of a grid into one cell.**
  It collects its axis through a `HashSet`, so with `Eq` on `x` alone, four
  distinct `(x, y)` indices became two — and a checked-in, passing test
  asserted exactly that loss.
- **`Surface::get_curve` returned three of its five projected points**, keeping
  only the greatest ordinate per abscissa. `(0, 0)` and `(1, 1)` were simply
  absent, and the tests probing for them passed only because `x`-only equality
  made `(0, 1)` match a probe for `(0, 0)`. The same defect caused the loss and
  concealed it.

## The decision, and its evidence

Points carry their full coordinates; the function-of-abscissa invariant belongs
to the container, not to the point.

The evidence is the asymmetry in how the two containers index themselves.
`Curve` uses `AxisOperations<Point2D, Decimal>` — a bare abscissa, because only
`x` identifies a curve sample. `Surface` uses
`AxisOperations<Point3D, Point2D>` — an `(x, y)` pair, where **both coordinates
are load-bearing**. Taking `y` out of `Ord` instead would have kept `Eq` and
`Hash` on `x` alone and left the surface bug fully intact.

`Hash` also stops reading `mantissa()` and `scale()` directly. `Decimal`
compares numerically, so `dec!(1.0) == dec!(1.00)` while those two hashed
differently — `Hash` disagreed with `Eq` independently of everything above.

## What this PR deliberately does not do

An earlier cut normalized duplicates away at construction. That was removed,
because "greatest ordinate wins" happens to be exactly what `collect` was doing
by accident, and blessing an accident as the semantics of a projection is worse
than leaving the question open. A projection of a grid is genuinely
multi-valued, and `points` is a `pub` field on both types, so a struct literal
bypasses any constructor — normalization could never be a real invariant.

`Curve::new`, `Surface::new` and `get_curve` now document what they store, and
**#466** tracks the design decision. `get_curve` has no production callers and
is uncallable from outside the crate — its `Axis` parameter type is not
re-exported — so the blast radius is the three tests that assert it.

## Testing

- [x] `tests/property/point_contract_test.rs` — asserts `a == b` iff
      `a.cmp(&b) == Equal`, and that `Hash` agrees with `Eq`, for both types
- [x] **Six existing tests asserted the old contract and were rewritten**, not
      deleted. Two of them asserted both halves of the violation 286 lines apart
      in the same file. `test_merge_indexes` now asserts all four indices, with
      a comment saying the old `2` was the data loss rather than the
      specification
- [x] The three projection tests are back to their original assertions
      verbatim, and now pass for the right reason
- [x] `cargo test --all-features --workspace` — 3941 lib, 423 integration,
      55 property, 207 doc, 0 failures
      `make public-api-check` clean

## Public API impact

One addition: `impl Hash for Point3D`. The snapshot diff in
`public-api/optionstratlib.txt` is that impl and nothing else.

Behaviour that moves for callers: a `HashSet` or `HashMap` keyed on either
point type now keeps entries it used to merge. There is exactly one such site
in the crate — the `HashSet` in `merge_indexes` — and widening it is the fix.

## Stack

Closes #450

### DIFF
--- a/src/curves/types.rs
+++ b/src/curves/types.rs
@@ -79,7 +79,10 @@ impl Display for Point2D {
 /// onto one cell whenever those indices pass through a `HashSet`.
 ///
 /// The "one ordinate per abscissa" rule is a property of a *curve*, not of a
-/// point; [`crate::curves::Curve`] enforces it at construction.
+/// point, and it is a container-level invariant that nothing currently
+/// enforces: [`crate::curves::Curve::new`] stores a repeated abscissa
+/// unchanged, and whether such a curve should be rejected, normalized or
+/// supported is deferred to issue #466.
 impl PartialEq for Point2D {
     fn eq(&self, other: &Self) -> bool {
         self.x == other.x && self.y == other.y
--- a/src/curves/types.rs
+++ b/src/curves/types.rs
@@ -66,9 +66,23 @@ impl Display for Point2D {
     }
 }
 
+/// Two points are equal when both coordinates are equal.
+///
+/// `PartialEq`, `Ord` and `Hash` all read the full `(x, y)` pair, so
+/// `a == b` holds exactly when `a.cmp(&b)` is [`Ordering::Equal`], and equal
+/// points hash alike. `Ord` requires that agreement.
+///
+/// Comparing on `x` alone would also break `Point2D` in its second role: it
+/// is the *index* type of a surface (`AxisOperations<Point3D, Point2D>`),
+/// where it names a coordinate in the xy-plane and `z` is the dependent
+/// value. An x-only `Eq`/`Hash` collapses every column of a surface grid
+/// onto one cell whenever those indices pass through a `HashSet`.
+///
+/// The "one ordinate per abscissa" rule is a property of a *curve*, not of a
+/// point; [`crate::curves::Curve`] enforces it at construction.
 impl PartialEq for Point2D {
     fn eq(&self, other: &Self) -> bool {
-        self.x == other.x
+        self.x == other.x && self.y == other.y
     }
 }
 
@@ -76,8 +90,11 @@ impl Eq for Point2D {}
 
 impl Hash for Point2D {
     fn hash<H: Hasher>(&self, state: &mut H) {
-        self.x.mantissa().hash(state);
-        self.x.scale().hash(state);
+        // Delegate to `Decimal`'s own `Hash`, which normalizes first. Hashing
+        // the raw mantissa and scale would give `dec!(1.0)` and `dec!(1.00)`
+        // different hashes even though they compare equal.
+        self.x.hash(state);
+        self.y.hash(state);
     }
 }
 
@@ -87,6 +104,8 @@ impl PartialOrd for Point2D {
     }
 }
 
+/// Orders on `(x, y)`, lexicographically: the same pair [`PartialEq`] reads,
+/// so `cmp` returns [`Ordering::Equal`] exactly when the points are equal.
 impl Ord for Point2D {
     fn cmp(&self, other: &Self) -> Ordering {
         match self.x.cmp(&other.x) {
@@ -313,6 +332,15 @@ mod tests {
         }
     }
 
+    /// Equality reads both coordinates.
+    ///
+    /// This test used to assert `p1 == p3` and `p1 == p4` — that points
+    /// sharing an abscissa were equal whatever their ordinate. That is the
+    /// contract this type no longer has: `test_comparison_operators` below
+    /// asserts `p1 < p3` on the very same two values, and `Ord` requires
+    /// `a == b` exactly when `a.cmp(&b)` is `Equal`. Only one of the two
+    /// could survive, and the ordering is the one the rest of the crate
+    /// relies on (`BTreeSet<Point2D>`, and `Point2D` as a surface index).
     #[test]
     fn test_equal() {
         let p1 = Point2D::from_f64_tuple(1.0, 2.0).unwrap();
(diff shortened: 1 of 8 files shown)


src/curves/types.rs
```diff
@@ -66,18 +66,35 @@ impl Display for Point2D {
...
+/// points hash alike. `Ord` requires that agreement.
+///
+/// Comparing on `x` alone would also break `Point2D` in its second role: it
+/// is the *index* type of a surface (`AxisOperations<Point3D, Point2D>`),
+/// where it names a coordinate in the xy-plane and `z` is the dependent
+/// value. An x-only `Eq`/`Hash` collapses every column of a surface grid
+/// onto one cell whenever those indices pass through a `HashSet`.
+///
+/// The "one ordinate per abscissa" rule is a property of a *curve*, not of a
+/// point; [`crate::curves::Curve`] enforces it at construction.
```
joaquinbejar: 💡 **Suggestion**: This contradicts the new `Curve::new` documentation immediately above, which says duplicate abscissae are stored unchanged and the invariant is unresolved in #466. `Surface::new` likewise retains duplicate `(x, y)` coordinates, so the analogous sentence in `src/surfaces/types.rs` is also inaccurate. Please describe these as container-level invariants that are currently unresolved/deferred, rather than saying either constructor enforces them.

joaquinbejar (commented):
Decision: COMMENT. The Eq/Ord/Hash implementations now agree on every coordinate, the scale-normalized Decimal hashing is pinned by property tests, and the changed projection/merge behavior is covered; the 15 contract properties and the focused unit tests pass, while build, format, lint, public-API, and semver checks are green. One public-documentation inconsistency remains: the point docs say the constructors enforce coordinate uniqueness even though those constructors explicitly retain duplicates and defer that invariant to #466.

src/curves/types.rs
joaquinbejar: Reworded in 71962bc6, both sentences now say the rule is a container-level invariant that nothing enforces, with Curve::new storing a repeated abscissa and Surface::new retaining a repeated coordinate, and #466 tracking the decision. Fixed the analogous sentence in src/surfaces/types.rs too.

joaquinbejar:
Documentation inconsistency fixed in 71962bc6, and the branch is rebased on main. Both point docs now describe the uniqueness rule as a deferred container-level invariant rather than something the constructors enforce.

src/curves/types.rs
joaquinbejar: Fixed. Both sentences now say the rule is a container-level invariant that is currently unresolved, with the constructors storing duplicates as given and #466 tracking what it should be. You are right that the previous wording contradicted the constructor docs one screen away — and it was contradicting the code as well, since normalising at construction is exactly what this PR backed out of doing.

joaquinbejar (commented):
The follow up corrects both public documentation blocks to match the constructors and the deferred decision in issue 466, and all 15 contract properties still pass. The build, format, lint, public API, and semver checks are green. APPROVE. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

joaquinbejar (commented):
The rebase preserves both reviewed patches unchanged, and all 15 contract properties still pass on the new head. The branch is mergeable and the completed checks are green. APPROVE. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

Issue #463, opened 2026-08-29
Panic sweep: break-even indexing in the multi-leg strategies


joaquinbejar:
## Overview

Second half of the `strategies/` sweep, split off from #460 because the whole
group is past what a reviewer can check numerically in one pass.

This half is the mechanical one: **65 unguarded `break_even_points[i]` accesses
across 14 multi-leg strategy files**, in `get_profit_ranges`,
`get_loss_ranges`, `get_profit_area` and `get_profit_ratio`. Roughly 330 of the
1,131 panics the probe found.

They fail with `index out of bounds: the len is 0 but the index is 0` whenever
the break-even vector is empty — which happens when a constructor did not
populate it, or when the strategy has no break-even at all.

## Scope

The 14 multi-leg files, including `short_butterfly_spread.rs`,
`call_butterfly.rs`, `long_butterfly_spread.rs`, `iron_condor.rs`,
`iron_butterfly.rs`, the spreads and the straddles/strangles.

Explicitly **not** here — they are #460, because they fire on well-formed input
and need judgement rather than a pattern: `base.rs`, `utils.rs`, `custom.rs`,
`probabilities/`, `delta_neutral/`, the four single-leg strategies, the three
strategies missing `one_option`, and `risk/` `pnl/` `backtesting/`.

## Why it is worth its own PR

It reads as one pattern repeated 65 times, which is the kind of diff a reviewer
can check in bulk rather than line by line — the opposite of the material that
produced the two silently-wrong values in #452, which hid in a large diff among
changes that preserved behaviour.

## Method

As #444, #446, #456 and #460: probe first, fix in the order reject → limit →
checked access, then a proptest. Use `.get()` / `.first()` with a typed error,
and reuse `price_gap` and `lower_break_even` from `src/strategies/base.rs`
rather than adding another helper.

## Tasks

- [ ] Replace every `break_even_points[i]` with a checked access returning a
      typed error
- [ ] Decide, per strategy, whether an empty break-even vector is a
      constructor bug to fix or a legitimate state to report
- [ ] Extend the #460 proptest to cover the multi-leg strategies
- [ ] State the unchanged-behaviour claim per file

## Acceptance criteria

- [ ] The probe reports zero panics from these files
- [ ] Existing tests pass unedited, except any asserting a removed contract
- [ ] `cargo clippy --all-targets --all-features --workspace -- -D warnings`,
      `cargo fmt --all --check`, `cargo test --all-features --workspace`,
[...]

PR #468, opened 2026-08-29
Make the composition layer panic-free (#460)

### LINKED ISSUE #460: Panic sweep: strategies, P&L, risk, metrics and backtesting

joaquinbejar:
## Summary

A probe over the public API of the strategies, P&L, risk and backtesting layer
found **1,131 panics in 26,719 calls — and five inputs that did not panic at
all, they hung**. It now finds **363 panics in 35,087 calls and no hangs**, and
none of the 363 is in a file this PR touches: they are the multi-leg indexing
split off into #463, and four defects outside this layer.

## The hangs matter more than the count

`calculate_price_range` walked `while current <= end { current += step }`. A
step of `Positive::ZERO` never terminates and grows a `Vec` until the OOM
killer — and zero is a legal step, which `expected_value` reaches by computing
`step = underlying_price / 100` from a zero spot. `CustomStrategy::new` scanned
`0.5·S` to `1.5·S` a cent at a time, which at `S = 1e20` is 1e22 iterations.

Both now walk a bounded range, sized with `Decimal` arithmetic before the first
step and capped. A panic leaves a stack trace and a dead request; a hang leaves
a stuck worker that answers every health check.

## Three defects that needed no extreme input

- **445 hits**: `Collar`, `CoveredCall` and `ProtectivePut` never overrode
  `one_option`, so `BasicAble::get_underlying_price()` — a plain getter —
  aborted the process on all three shipped strategies. They now override it.
- **Every call**: `LongCall`, `ShortCall`, `LongPut` and `ShortPut` never call
  `update_break_even_points` in their constructors, so the break-even vector is
  always empty and `get_profit_area()` aborted on **a default 100/110 long call
  at 20% vol**.
- **168 hits**: `probabilities/utils.rs` took `(target / current).ln()` with raw
  `Positive` division — a zero spot broke the invariant, a 1e-28 spot overflowed
  it. The zero spot is now rejected at the entry point as a typed error.

## The two `panic!` in `base.rs` stay, marked

`one_option` returns `&Options` and `one_option_mut` returns `&mut Options`.
There is no value of those types to return, and fabricating a default `Options`
would answer `get_underlying_price` with a price nobody quoted — the failure
mode that is worse than the abort. They keep a `// scan-banned: allow` with
that argument.

The probe now shows no strategy this crate ships reaches the default: those 445
hits were entirely the three missing overrides, and the count after adding them
is zero. The remaining exposure is a downstream implementor that forgets. The
alternative — `Result<&Options, StrategyError>` — ripples through 13 trait
defaults, 22 call sites in `short_strangle.rs` and 2 in `probabilities/core.rs`,
and is worth its own change if you want it.

## Scope, and what went to #463

25 files, 1,265 insertions. The seam moved once during the work: the 17
constructor `net_cost / quantity` divisions were going to land here, but
rewiring them touches the same 14 multi-leg files that #463 must edit for the
indexing, and two PRs editing the same lines is worse than one bigger PR. Only
`bull_call_spread.rs` is fixed here, because the proptest drives the base trait
defaults through it and its break-even panic shadowed everything behind it.

#463 is now 237 of the 363: 195 `break_even_points[i]` accesses across three
butterfly files, 24 constructor divisions, 18 profit-area divisions. The
pattern for every one of them is already in this PR.

## Unchanged behaviour

The claim rests on two verified facts rather than on testing. Every `Positive`
operator in `positive` 0.6 delegates to its own `checked_*`, so swapping in the
checked form cannot move a value. And `d_div` re-rounds at
`DIV_DEFAULT_SCALE = 28`, which is `Decimal`'s maximum scale, so that rounding
is the identity and `d_div == checked_div == /`.

Two deliberate behaviour changes, both commented at the site: a zero-quantity
leg drops out of the optimizer's candidate list instead of aborting it, and
`AdjustmentPlan::new`'s quality score falls back to `Decimal::MAX` — the worst
rank — when it cannot be computed. That is a ranking key, not a price.

One I nearly made and backed out: `Collar::update_break_even_points` keeps its
`if let Ok(be)` shape, so a break-even below zero still records **no point**
rather than recording zero. Changing it to the `lower_break_even` sentinel
would have changed what `get_break_even_points()` returns.

## Testing

- [x] `tests/property/strategies_panic_freedom_test.rs` — 8 properties driving
      the base trait defaults, **two** of the four single-leg strategies
      (`LongCall` and `ShortPut`), the three spot-leg strategies, the
      probability kernels, SPAN, and the P&L primitives, over the structural
      degeneracies as well as the numeric extremes
- [x] **The single-leg coverage is two, not four, and the gap is real.**
      `LongPut` and `ShortCall` were reached through `StrategyConstructor`,
      whose implementations answer `OperationNotSupported` for both, so those
      branches never executed and the property passed without testing anything.
      They are now removed from it rather than left pretending. `LongPut`'s
      break-even is covered by a direct regression instead; `ShortCall` has
      only its fixed simulation tests and **no** boundary-input panic property
- [x] Three of its generators are **deliberately bounded**, each with the reason
      in the file, because the property cannot hold above those bounds until the
      out-of-scope defects below are fixed. That is a real gap in the test and
      it is stated rather than hidden
- [x] Only `src/strategies/utils.rs` tests were touched, because
      `calculate_price_range` now returns `Result`; the five keep their
      assertions and gained an `.expect`. Four new ones cover the bounded walk,
      the empty range, the rejected zero step and the sample cap
- [x] 3941 lib, 423 integration, 48 property, 207 doc — 0 failures
- [x] `cargo fmt --all --check`, clippy `-D warnings`, `make scan-banned`,
      `make public-api-check` and `cargo doc` clean

## Public API impact

**Two breaking changes, both on `ProtectivePut`, plus additive conversions.**

| before | after |
|---|---|
| `pub fn total_fees(&self) -> Positive` | `pub fn total_fees(&self) -> Result<Positive, PositiveError>` |
| `pub fn protection_level(&self) -> Decimal` | `pub fn protection_level(&self) -> Result<Decimal, StrategyError>` |

Both aborted the process on inputs a caller can supply: a fee at
`Positive::MAX` in the raw fee sum, and a zero spot in `protection_level`'s
raw division. A `try_` twin was considered and rejected, because it would have
left the panicking method in place — the opposite of what this sweep is for,
and the same reason the return type of `SimulationStats::update`,
`Simulator::get_random_walk` and `generate_ou_process` moved in the PRs before
this one.

**Migration.** Add `?` where the value is used inside a fallible function, or
`.expect("…")` at a boundary that cannot propagate. `ProtectivePut::new` now
rejects a zero underlying price, so a position built through the constructor
never sees the error arm of `protection_level`; a deserialized one can.

Additive: `impl From<DecimalError> for StrategyError` and
`impl From<DecimalError> for AdjustmentError`.

## Found outside this layer, not fixed

Filed separately rather than folded in: the `Positive` accumulation in
`model/position.rs` (126 of the residual 363), the upstream calendar overflow
in `expiration_date` reached through `get_date()`, `model/utils.rs::mean_and_std`,
and the greek aggregators in `greeks/equations.rs` — where `alpha_from` returns
`Decimal::MAX` as a documented sentinel, so two legs at a sub-contract quantity
abort on `MAX + MAX`.

Closes #460

### DIFF
--- a/src/strategies/collar.rs
+++ b/src/strategies/collar.rs
@@ -484,7 +484,8 @@ impl Collar {
     /// range.
     fn total_fees(&self) -> Result<Positive, PositiveError> {
         self.spot_leg
-            .fees()
+            .open_fee
+            .checked_add(&self.spot_leg.close_fee)?
             .checked_add(&self.long_put.open_fee)?
             .checked_add(&self.long_put.close_fee)?
             .checked_add(&self.short_call.open_fee)?
--- a/src/strategies/collar.rs
+++ b/src/strategies/collar.rs
@@ -76,18 +76,21 @@ use crate::error::{GreeksError, PositionError, PricingError, StrategyError};
 use crate::greeks::Greeks;
 use crate::model::ExpirationDate;
 use crate::model::ProfitLossRange;
+use crate::model::decimal::{d_add, d_div, d_sub};
 use crate::model::leg::traits::LegAble;
 use crate::model::leg::{Leg, SpotPosition};
 use crate::model::position::Position;
 use crate::model::types::{OptionBasicType, OptionStyle, OptionType, Side};
 use crate::pnl::PnLCalculator;
 use crate::pricing::payoff::Profit;
+use crate::strategies::base::price_gap;
 use crate::strategies::delta_neutral::DeltaNeutrality;
 use crate::strategies::probabilities::core::ProbabilityAnalysis;
 use crate::strategies::probabilities::utils::VolatilityAdjustment;
 use crate::strategies::{BasicAble, Strategies};
 use chrono::Utc;
 use positive::Positive;
+use positive::PositiveError;
 use rust_decimal::Decimal;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
@@ -204,6 +207,17 @@ impl Collar {
         call_close_fee: Positive,
     ) -> Result<Self, StrategyError> {
         // Create the spot position (long underlying)
+        // Every per-share figure this strategy reports — the effective cost
+        // basis, the break-even, the premium per share — divides by the size
+        // of the spot leg. A collar with no shares is not a position, and
+        // rejecting it here keeps those divisors non-zero.
+        if quantity == Positive::ZERO {
+            return Err(StrategyError::invalid_parameters(
+                "Collar::new",
+                "quantity must be strictly positive: a collar with no shares has no per-share basis",
+            ));
+        }
+
         let spot_leg = SpotPosition::new(
             underlying_symbol.clone(),
             quantity,
@@ -331,9 +345,12 @@ impl Collar {
     }
 
     /// Returns the collar width (distance between put and call strikes).
+    ///
+    /// A put struck above the call inverts the collar; the width of that
+    /// region is zero, not a negative price.
     #[must_use]
     pub fn collar_width(&self) -> Positive {
-        self.call_strike() - self.put_strike()
+        price_gap(self.call_strike(), self.put_strike())
     }
 
     /// Calculates the net premium (credit if positive, debit if negative).
@@ -370,7 +387,11 @@ impl Collar {
         let spot_delta = self.spot_leg.delta()?;
         let put_delta = self.long_put.delta()?;
         let call_delta = self.short_call.delta()?;
-        Ok(spot_delta + put_delta + call_delta)
+        Ok(d_add(
+            d_add(spot_delta, put_delta, "Collar::net_delta")?,
+            call_delta,
+            "Collar::net_delta",
+        )?)
     }
 
     /// Calculates the maximum profit potential.
@@ -392,16 +413,24 @@ impl Collar {
         let cost_basis = self.spot_leg.cost_basis;
         let quantity = self.spot_leg.quantity;
         let net_premium = self.net_premium();
-        let total_fees = self.total_fees();
+        let total_fees = self.total_fees()?;
 
         if call_strike >= cost_basis {
-            let capital_gain = (call_strike - cost_basis) * quantity;
-            let total_profit = capital_gain.to_dec() + net_premium - total_fees.to_dec();
+            let capital_gain = price_gap(call_strike, cost_basis).checked_mul(&quantity)?;
+            let total_profit = d_sub(
+                d_add(capital_gain.to_dec(), net_premium, "Collar::max_profit")?,
+                total_fees.to_dec(),
+                "Collar::max_profit",
+            )?;
             Ok(Positive::new_decimal(total_profit.max(Decimal::ZERO)).unwrap_or(Positive::ZERO))
         } else {
             // Call strike below cost basis
-            let capital_loss = (cost_basis - call_strike) * quantity;
-            let total_profit = net_premium - capital_loss.to_dec() - total_fees.to_dec();
+            let capital_loss = price_gap(cost_basis, call_strike).checked_mul(&quantity)?;
+            let total_profit = d_sub(
+                d_sub(net_premium, capital_loss.to_dec(), "Collar::max_profit")?,
+                total_fees.to_dec(),
+                "Collar::max_profit",
+            )?;
             Ok(Positive::new_decimal(total_profit.max(Decimal::ZERO)).unwrap_or(Positive::ZERO))
         }
     }
@@ -425,27 +454,41 @@ impl Collar {
         let cost_basis = self.spot_leg.cost_basis;
         let quantity = self.spot_leg.quantity;
         let net_premium = self.net_premium();
-        let total_fees = self.total_fees();
+        let total_fees = self.total_fees()?;
 
         if cost_basis >= put_strike {
-            let capital_loss = (cost_basis - put_strike) * quantity;
-            let total_loss = capital_loss.to_dec() - net_premium + total_fees.to_dec();
+            let capital_loss = price_gap(cost_basis, put_strike).checked_mul(&quantity)?;
+            let total_loss = d_add(
+                d_sub(capital_loss.to_dec(), net_premium, "Collar::max_loss")?,
+                total_fees.to_dec(),
+                "Collar::max_loss",
+            )?;
             Ok(Positive::new_decimal(total_loss.max(Decimal::ZERO)).unwrap_or(Positive::ZERO))
         } else {
             // Put strike above cost basis (unusual but possible)
-            let capital_gain = (put_strike - cost_basis) * quantity;
-            let total_loss = total_fees.to_dec() - net_premium - capital_gain.to_dec();
+            let capital_gain = price_gap(put_strike, cost_basis).checked_mul(&quantity)?;
+            let total_loss = d_sub(
+                d_sub(total_fees.to_dec(), net_premium, "Collar::max_loss")?,
+                capital_gain.to_dec(),
+                "Collar::max_loss",
+            )?;
             Ok(Positive::new_decimal(total_loss.max(Decimal::ZERO)).unwrap_or(Positive::ZERO))
         }
     }
 
     /// Calculates total fees for all positions.
-    fn total_fees(&self) -> Positive {
-        self.spot_leg.fees()
-            + self.long_put.open_fee
-            + self.long_put.close_fee
-            + self.short_call.open_fee
-            + self.short_call.close_fee
+    ///
+    /// # Errors
+    ///
+    /// Returns [`PositiveError`] when the running total leaves the `Positive`
+    /// range.
+    fn total_fees(&self) -> Result<Positive, PositiveError> {
+        self.spot_leg
+            .fees()
+            .checked_add(&self.long_put.open_fee)?
+            .checked_add(&self.long_put.close_fee)?
+            .checked_add(&self.short_call.open_fee)?
+            .checked_add(&self.short_call.close_fee)
     }
 
     /// Checks if the put is currently in-the-money.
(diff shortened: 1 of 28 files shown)


src/strategies/collar.rs
```diff
@@ -425,27 +454,41 @@ impl Collar {
...
-            + self.short_call.open_fee
-            + self.short_call.close_fee
+    ///
+    /// # Errors
+    ///
+    /// Returns [`PositiveError`] when the running total leaves the `Positive`
+    /// range.
+    fn total_fees(&self) -> Result<Positive, PositiveError> {
+        self.spot_leg
+            .fees()
```
joaquinbejar: 🔴 **Critical**: This still calls `SpotPosition::fees()`, which uses raw `Positive + Positive`. `Collar::new` with `spot_open_fee = Positive::MAX` and a nonzero close fee panics before this `Result` can return. Sum the spot fee fields with `checked_add` here. The same unchecked fee aggregation remains in `ProtectivePut::total_fees` and both `CoveredCall::max_profit_potential` and `max_loss_potential`.

src/strategies/protective_put.rs
```diff
@@ -90,6 +92,17 @@ impl ProtectivePut {
         put_open_fee: Positive,
         put_close_fee: Positive,
     ) -> Result<Self, StrategyError> {
+        // Every per-share figure this strategy reports — the effective cost
+        // basis, the break-even, the premium per share — divides by the size
+        // of the spot leg. A protective put with no shares is not a position, and
+        // rejecting it here keeps those divisors non-zero.
+        if quantity == Positive::ZERO {
```
joaquinbejar, line 100: 🔴 **Critical**: This guard still accepts a zero underlying price, but `protection_level()` divides by `spot_leg.cost_basis` with raw `Decimal / Decimal`. `ProtectivePut::new(... underlying_price = Positive::ZERO ...).protection_level()` will panic. Reject a zero spot here or make `protection_level` fallible and use checked division.

src/strategies/long_put.rs
```diff
@@ -268,10 +270,18 @@ impl BreakEvenable for LongPut {
...
+        let per_contract = d_div(
+            self.get_net_cost()?,
+            self.long_put.option.quantity.to_dec(),
+            "LongPut::update_break_even_points",
+        )?;
         self.break_even_points.push(
-            (self.long_put.option.strike_price
-                + self.get_net_cost()? / self.long_put.option.quantity)
-                .round_to(2),
+            lower_break_even(self.long_put.option.strike_price, -per_contract)
```
joaquinbejar: 🟠 **Warning**: `get_net_cost()` is a positive debit for a long put. Negating `per_contract` makes `lower_break_even` subtract a negative value, so a 100 strike with a 5 debit reports 105 instead of 95. Pass `per_contract` without negating it. The new property does not catch this because both the `LongPut` and `ShortCall` `StrategyConstructor` implementations always return `OperationNotSupported`, so those branches never run.

joaquinbejar (commented):
The panic sweep still leaves reproducible panics in the spot strategy paths, and the rewritten long put break even moves in the wrong direction while its property branch never executes. REQUEST_CHANGES. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

src/strategies/collar.rs
joaquinbejar: Fixed, the four aggregations sum the two spot fee fields with checked_add instead of going through SpotPosition::fees(). That covers Collar::total_fees, ProtectivePut::total_fees and both CoveredCall::max_profit_potential and max_loss_potential.

src/strategies/protective_put.rs
joaquinbejar: Did both in the end. new() rejects a zero underlying price now, and protection_level is fallible with checked division too, since ProtectivePut derives Deserialize and a position can reach the struct without passing through the constructor.

src/strategies/long_put.rs
joaquinbejar: Right on both counts, the negation is gone so a 100 strike bought for 5 breaks even at 95. Since the property cannot reach LongPut through StrategyConstructor, the regression is a direct test that builds the position and checks the break-even.

joaquinbejar:
All three are addressed in 292676fd. The spot fee aggregations go through checked_add at the four call sites rather than through SpotPosition::fees(), ProtectivePut rejects a zero spot and its protection_level is fallible, and the long put break-even no longer negates the debit so a 100 strike bought for 5 reports 95. The public API snapshot carries the two ProtectivePut signature changes.

src/strategies/protective_put.rs
```diff
@@ -202,34 +227,84 @@ impl ProtectivePut {
...
-    #[must_use]
-    pub fn protection_level(&self) -> Decimal {
+    ///
+    /// # Errors
+    ///
+    /// Returns [`StrategyError`] when the spot leg's cost basis is zero, which
+    /// [`ProtectivePut::new`] rejects but a deserialized position can still
+    /// carry, and when the difference or the ratio leaves the representable
+    /// `Decimal` range.
+    pub fn protection_level(&self) -> Result<Decimal, StrategyError> {
```
joaquinbejar, line 289: 🔴 **Critical**: This reshapes a public method from `Decimal` to `Result<Decimal, StrategyError>`, and `total_fees` above also changes from `Positive` to `Result`. Existing downstream code will stop compiling, while the PR still says the public API impact is purely additive and the changelog gives no migration note. Preserve source compatibility with additive `try_` methods, or explicitly declare both breaking changes and document the migration before merging.

joaquinbejar (commented):
The follow up fixes the three reported runtime defects and the targeted tests pass, but it also introduces two breaking ProtectivePut signatures while the PR still describes the public API as purely additive and gives no migration note. REQUEST_CHANGES. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

tests/property/strategies_panic_freedom_test.rs
```diff
@@ -0,0 +1,560 @@
...
+
+        // `ShortCall` and `LongPut` are deliberately absent from this
+        // property. Their `new` is private and their
+        // `StrategyConstructor::get_strategy` always returns
+        // `OperationNotSupported`, so nothing an integration test can write
+        // constructs one: an earlier version of this file wrapped
+        // `get_strategy` in `if let Ok(..)`, and that branch never executed.
+        // Dead code shaped like coverage is worse than none, because it reads
+        // as tested. Their coverage lives in-file, beside the private
+        // constructor — see `tests_long_put_break_even` in `long_put.rs`.
```
joaquinbejar, line 354: 💡 **Suggestion**: This now explicitly removes `LongPut` and `ShortCall` from the property, but the PR Testing section still claims this file drives all four single leg strategies via `StrategyConstructor`. Please update the PR description to record the actual two strategy coverage and the gap, especially because `ShortCall` has only fixed simulation tests rather than an equivalent boundary input panic property.

joaquinbejar (commented):
This follow up correctly removes the two dead branches and the targeted property suite passes, but the PR description now overstates the property coverage and the previously reported breaking ProtectivePut API mismatch remains unresolved. REQUEST_CHANGES. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

src/strategies/protective_put.rs
joaquinbejar: Declared them rather than papering over with try_ twins, since a twin leaves the panicking method in place and this sweep exists to remove it. The changelog now carries both signature changes with the migration, and the PR body has a before and after table plus the note that new() rejects a zero spot so only a deserialized position reaches the error arm.

tests/property/strategies_panic_freedom_test.rs
joaquinbejar: Fair, the description overstated it. The Testing section now says two single leg strategies rather than four, explains that LongPut and ShortCall never executed because StrategyConstructor answers OperationNotSupported for both, and records that ShortCall has no equivalent boundary input property.

joaquinbejar (commented):
The follow up now declares both breaking ProtectivePut signatures in the PR summary and changelog, gives a concrete migration path, and accurately records the single leg property coverage gap. The targeted tests passed and the completed build, format, lint, public API and semver checks are green. APPROVE. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

Issue #471, opened 2026-08-29
Infallible signatures that can still abort


joaquinbejar:
## Overview

A group of public methods whose return type has no error channel, and whose
bodies can abort. They are what is left after #460 closed everything reachable
through the public constructors: each of these is now reachable only by
mutating a `pub` field directly, or through a trait whose signature forbids
failure.

None of them can be fixed without a public signature change, which is why #460
reported them rather than fixing them.

## The trait one, which needs a design decision

`src/model/leg/spot.rs:251` — `LegAble::pnl_at_price(&self, Positive) -> Decimal`
computes `(price - cost_basis) * quantity` and aborts with `Subtraction
overflowed`. It is infallible **by trait signature**, so fixing it means
changing the trait method to return `Result` and updating every implementor.
That is a design change, not a patch.

## The `Add`/`Sum` ones, which cannot be made fallible at all

`impl Add for PnL` and `impl Sum for PnL` add `initial_costs: Positive +
Positive`. The std traits return `Self`, so there is no fallible form of them.

#460 added a `pub(crate) PnL::try_add` and routed every accumulation inside the
library through it, so the operator is only reachable from caller code now. To
close it properly: make `try_add` public and deprecate the operator.

## The getters

Each is a `-> Decimal` or `-> Positive` over arithmetic on `pub` fields:

- `Collar::net_premium` — `premium * quantity` overflow, also reached from
  `is_zero_cost` and `is_credit`, both `-> bool`
- `CoveredCall::effective_cost_basis`, `ProtectivePut::effective_cost_basis` —
  divide by the spot leg size
- `ProtectivePut::total_fees`, `ProtectivePut::protection_level` — the latter
  divides by the cost basis
- `PortfolioGreeks::delta_gap`, `gamma_gap`, `add`, `combined` — unchecked
  `Decimal` on `pub` fields a caller sets

## The decision to make

Three options, and the right one may differ per item:

- Return `Result` and accept the public break. The crate is pre-1.0, and
  `main` now has a `public-api` snapshot job so the break lands in the diff.
- Make the fields private with checked constructors, so the arithmetic cannot
  be handed impossible inputs in the first place.
- Document the precondition and leave the panic, with a
  `// scan-banned: allow` and the argument — the position #460 took for
  `one_option`, where no value of the return type exists.

## Context

Found by `panic-hunter` while sweeping the composition layer for #460. Every
one of them was unreachable through the public constructors after that PR's
[...]

Issue #470, opened 2026-08-29
Panic sweep: the model layer reached from strategies


joaquinbejar:
## Overview

Three defects in the model layer that the composition sweep (#460) reached but
could not fix in scope. Together they are 126 of the 363 panics that sweep left
behind, and they cap what its property test can generate.

## 1. `Position` cost arithmetic — `src/model/position.rs:258` and `:779`

`Positive` operators inside functions that already return
`Result<Positive, PositionError>`:

```rust
Side::Long => (self.premium + self.open_fee + self.close_fee) * self.option.quantity,
```

becomes

```rust
Side::Long => self.premium
    .checked_add(&self.open_fee)?
    .checked_add(&self.close_fee)?
    .checked_mul(&self.option.quantity)?,
```

and likewise `Ok((self.open_fee + self.close_fee) * self.option.quantity)`.
`PositionError` has no `From<PositiveError>`; either add one or map through
`PositionError::invalid_position`.

## 2. Calendar overflow through `get_date()` — `src/model/position.rs:1134`

`` `DateTime + TimeDelta` overflowed `` for `ExpirationDate::Days(1e9)`. The
panic is upstream, in `expiration_date` 0.3.0 `convert.rs:134`:

```rust
Ok(base + Duration::days(i64::try_from(*days)?))
```

The proper fix is there — `base.checked_add_signed(Duration::try_days(..)?)` —
or a range guard in this crate before the call. **This is the single largest
residual**, and every strategy reaches it through
`calculate_pnl_at_expiration`. #441 already guards the one call site in
`OptionChain::build_chain`; this is the general case.

## 3. `mean_and_std` — `src/model/utils.rs:436`

`vec.iter().sum::<Positive>() / vec.len() as f64` overflows on the sum and
divides by zero on an empty vector. Every `get_profit_ranges` implementation
averages its leg volatilities through it. It returns `(Positive, Positive)`, so
fixing it properly is a signature change to `Result` — which is why #460's
property test stops its volatility generator at 1e6.

## Acceptance criteria

- [ ] None of the three aborts for any input reachable through the public API
- [ ] #460's proptest generators can be unbounded for premium, fee and
      volatility, and the bounds removed from that file
- [ ] Existing tests pass unedited

joaquinbejar:
The calendar overflow is worse than the P&L call suggests, and the fix belongs in a repo we own.

It takes `Display` down as well. Same input, four accessors:

```
[...]

Issue #469, opened 2026-08-29
Greek aggregation aborts on its own alpha sentinel


joaquinbejar:
## Overview

The greek aggregators sum eleven greeks across a strategy's legs with unchecked
`Decimal` addition. The plain overflow is the lesser problem: `alpha_from`
returns `Decimal::MAX` as a **documented sentinel** when theta vanishes, so a
strategy with two legs whose theta rounds to zero aborts on
`Decimal::MAX + Decimal::MAX`.

That is not an extreme input. It needs two legs at a sub-contract quantity,
which is an ordinary position size for a scaled or fractional book.

## Context

- `src/greeks/equations.rs:235-246` — `Greeks::greeks`, the twelve-field
  aggregation
- the per-greek aggregators from `:276` — `delta += single.delta` and eleven
  siblings
- `alpha_from` documents `Decimal::MAX` as its "theta vanished" sentinel, so
  the value is expected, not an anomaly

Both functions already return `Result<_, GreeksError>` and
`GreeksError: From<DecimalError>` already exists, so the fix is twelve `d_add`
calls in `greeks` and one each in the eleven per-greek loops.

## Why it is not in #460

`Greeks::greeks` is the one call the strategies panic-freedom proptest
deliberately does not make, precisely because of this. Fixing it here lets that
exclusion be removed.

Found by `panic-hunter` while sweeping the composition layer.

## Tasks

- [ ] Convert the aggregation in `Greeks::greeks` and the eleven per-greek
      aggregators to `d_add`
- [ ] Decide whether `Decimal::MAX` is the right sentinel for a vanished theta
      now that summing two of them is an error rather than an abort
- [ ] Remove the `Greeks::greeks` exclusion from the strategies proptest

## Acceptance criteria

- [ ] A two-leg strategy whose legs both return the alpha sentinel reports an
      error rather than aborting
- [ ] Aggregated greeks are unchanged for every strategy whose legs sum inside
      the `Decimal` range
- [ ] The strategies proptest covers `Greeks::greeks`

Issue #462, opened 2026-08-29
Arithmetic Asian loses all precision at short maturities


joaquinbejar:
## Overview

The general branch of the arithmetic Asian — the one that prices every
non-degenerate contract — loses all its precision as maturity shortens, and it
returns confident wrong values rather than errors. At 86 seconds to expiry it
overprices by a factor of **612**.

Measured against mpmath quadrature on the defining double integral at 50
decimal places. `S = K = 100`, `σ = 20%`, `r = 8%`, `q = 4%`:

| maturity | crate | reference | relative error |
|---|---|---|---|
| 0.5 days | 0.171855815757 | 0.171855815411 | 2.0e-9 |
| 0.2 days | 0.108377149270 | 0.108377120846 | 2.6e-7 |
| 0.1 days (2.4 h) | 0.076522709098 | 0.076521744333 | 1.3e-5 |
| 0.05 days (72 min) | 0.054005346244 | 0.054052666108 | −8.8e-4 |
| 0.02 days (29 min) | 0.034378185828 | 0.034154202089 | +6.6e-3 |
| 0.01 days (14 min) | 0.041790507589 | 0.024139351974 | **+73%** |
| 0.005 days (7 min) | 0.234996014908 | 0.017063436757 | **+1277%** |
| 0.002 days (3 min) | 0.875324687084 | 0.010788684989 | **80×** |
| 0.001 days (86 s) | 4.678919422800 | 0.007627618506 | **612×** |

## Mechanism

The same catastrophic cancellation that #461 fixed in the `σ` direction, in the
`T` direction instead. `term1` and `term2` each grow as `2S²/(a·c·T²)` and
cancel down to about `S²`, while the quantity extracted from their difference
is `σ²_asian·T`, which *shrinks* with `T`. At 14 minutes the two terms are
around `2.8e15` and the signal is `3.65e-7`, so the branch runs out of digits.
Below `1e-5` days it collapses the other way, to approximately zero.

The two removable-limit branches added in #445 degrade the same way, one to two
decades later: `b = -σ²` and `b = -σ²/2` hold to about `1e-4` days, then return
the spot-volatility price — a factor of `√3` too high — at `1e-8` days.

## Why it is not covered by #461

#461 fixed the driftless branch and left the general branch untouched by
design, so that prices away from the boundaries were unchanged *by
construction*. That claim is still true. This issue is that the untouched
branch was already wrong in this regime — the PR neither introduced nor
[...]

PR #472, opened 2026-08-29
Widen a locked quote to one tick instead of leaving it locked (#459)

### LINKED ISSUE #459: Widened quotes can come out locked at one tick
## Overview

`widen_around` in `src/chains/optiondata.rs` does not guarantee `ask > bid`.
When half a spread rounds to exactly one tick — `spread = 0.02` at
`decimal_places = 2` — a sub-tick mid produces a locked quote: bid and ask both
land on `0.01`.

That is the honest consequence of applying the spread once (#456). It was
hidden before, because the spread was applied three times per row and the
second pass widened a mid the first had already lifted to the tick, so the
deep wings came out `0.01 / 0.02` by accident rather than by rule.

## The decision to make

A locked market is a real thing on an exchange, but a quote *generator* that
produces one is arguably wrong: a chain consumer marking at the touch sees a
zero-width spread on the cheapest strikes, which is the opposite of what a
thin market means.

The alternatives, none of which belongs in a panic sweep:

- Floor the ask one tick above the bid whenever they collide, so the thinnest
  market is one tick wide.
- Require `spread >= 2 ticks` at the build params and reject a spread that
  cannot produce a two-sided quote at the configured `decimal_places`.
- Accept the lock and document it, on the grounds that a contract worth less
  than half a tick has no meaningful two-sided market.

## Context

- `src/chains/optiondata.rs` — `widen_around`, and `tick_size` beside it
- `src/chains/chain.rs::test_new_option_chain_build_chain_long` asserts the
  current `0.01 / 0.01 / 0.01` for the deep wings
- #439 introduced the tick floor; #456 made the single application visible

## Tasks

- [ ] Decide which of the three the chain builder should do
- [ ] Implement it and state the rule in `apply_spread`'s doc comment
- [ ] Update the golden with the resulting numbers
- [ ] Add a property test asserting whatever invariant is chosen — `ask > bid`,
      or `ask >= bid` with the lock documented

## Acceptance criteria

- [ ] The quote invariant is stated in the documentation and enforced by a test
- [ ] A chain built with a spread narrower than two ticks behaves according to
      the decision rather than by accident
- [ ] No well-formed row moves

joaquinbejar:
## Summary

`widen_around` floored both sides of a generated quote at one tick and then
clamped the bid with `.min(ask)`. When half a spread rounds to exactly one tick
— `spread = 0.02` at `decimal_places = 2` — a sub-tick mid put both sides on
`0.01` and the chain came out **locked**: a zero-width market on the cheapest
strikes, which is the opposite of what a thin market means.

That was the honest consequence of applying the spread once (#456). Before it,
the spread was applied three times per row and the second pass widened a mid the
first had already lifted, so the deep wings came out `0.01 / 0.02` by accident
rather than by rule.

## The rule, now stated rather than accidental

A new private helper closes both widening paths:

```rust
fn unlock(bid: Positive, ask: Positive, tick: Positive) -> Option<(Positive, Positive)> {
    if bid < ask {
        return Some((bid, ask));
    }
    Some((bid, bid.checked_add(&tick).ok()?))
}
```

`widen_around` and `widen_book` end with `unlock(bid, ask, tick)` instead of
`Some((bid, ask))`. Every generated quote satisfies `tick <= bid < ask`. The
**bid never moves** — only the ask is lifted, by exactly one tick — and the lift
is `checked_add`, so an ask leaving the representable `Decimal` range still
returns `None` and `apply_spread` leaves that side untouched with its existing
`warn!`. The rule is written up under a `# The quote invariant` heading in
`apply_spread`'s doc comment.

The `if bid < ask { return … }` early exit is the mechanical guarantee behind
the "no well-formed row moves" criterion: a quote whose sides already differ is
returned by identity, so nothing but a collision can move.

## No well-formed row moved — measured, not asserted

Seven chains built across different `(spread, decimal_places)` configurations
and diffed before and after: **277 rows, 554 sides**.

- 138 sides moved, and **all 138 had `bid == ask` before**. Not one had a
  two-sided market.
- 0 bids moved. Every movement is the ask, raised by exactly one tick (`0.01`
  at `dp = 2`, `1` at `dp = 0`).
- The strike key sets before and after are identical — generation is unaffected.
- After the change, 0 rows violate `ask > bid`.

| config (spread, decimal_places) | sides moved |
|---|---|
| `tiny_0.001_dp2` | 75 |
| `coarse_0.5_dp0` | 51 |
| `golden_long` (0.02, dp 2) | 11 |
| `narrow_0.02_dp2` | 1 |
| `wide_0.10_dp2` | 0 |
| `narrow_0.002_dp3` | 0 |
| `short_21` (0.02, dp 2, 21 strikes) | 0 |

The three configs at zero are the control: their half-spread exceeds a tick, so
no row was ever locked and nothing moved. `coarse_0.5_dp0` is the same defect at
a different scale — half a spread of `0.25` against a tick of `1`.

## The golden

`test_new_option_chain_build_chain_long` asserts the two extreme strikes, so
exactly two assertions changed: `first.put_ask` and `last.call_ask`, `0.01` →
`0.02`. The mids stay `0.01`. Eleven deep-wing rows moved in total, all of them
locked at `0.01 / 0.01` beforehand. No other golden in the suite moved.

## Tests

`tests/property/quote_invariants_test.rs` now asserts `ask > bid` rather than
`ask >= bid`, plus a new property covering the sub-tick regime. Two unit tests
in `optiondata.rs` pin the two sides of the rule:
`test_locked_quote_is_lifted_one_tick` and
`test_two_sided_quote_is_left_where_the_widening_put_it`.

4675 passed, 0 failed, 79 ignored. `make scan-banned` clean.
`make public-api-check` clean — `unlock` is private beside the already-private
`widen_around`, `widen_book` and `tick_size`, so the public API is untouched.

## Stack

Chain D of four independent chains off `main`; this is its only PR, so the diff
reads against `main` directly.

- Stack: **#459** (this one) — independent of chains A (#470 → #469 → #463 →
  #471), B (#466 → #451 → #453) and C (#462), and of the #442 capstone.
- Base branch: `main`.

Closes #459

### DIFF
--- a/src/chains/optiondata.rs
+++ b/src/chains/optiondata.rs
@@ -239,16 +239,20 @@ fn tick_size(decimal_places: u32) -> Option<Positive> {
 /// A quote that is already two-sided is returned untouched: only a collision
 /// moves anything.
 ///
-/// Returns `None` when the lifted ask leaves the representable `Decimal`
-/// range; the lift is range-checked like every other step, never assumed to
-/// fit.
+/// Returns `None` when the lift cannot be made: either the sum leaves the
+/// representable `Decimal` range, or it lands back on the bid because the tick
+/// is finer than the last representable digit at that magnitude —
+/// `Decimal::MAX + 0.01` rounds to `Decimal::MAX`. Both cases would hand back
+/// the locked quote this function exists to remove, so neither is assumed to
+/// fit; the caller leaves the quote untouched instead.
 #[inline]
 #[must_use]
 fn unlock(bid: Positive, ask: Positive, tick: Positive) -> Option<(Positive, Positive)> {
     if bid < ask {
         return Some((bid, ask));
     }
-    Some((bid, bid.checked_add(&tick).ok()?))
+    let lifted = bid.checked_add(&tick).ok()?;
+    (lifted > bid).then_some((bid, lifted))
 }
 
 /// Widens a quote around `anchor` by half a spread on each side, flooring both
@@ -1620,6 +1624,46 @@ mod optiondata_coverage_tests {
 
         assert_eq!(option_data.call_bid, spos!(0.01));
         assert!(option_data.call_ask.unwrap() >= option_data.call_bid.unwrap());
+    }
+
+    /// The lift is range-checked rather than assumed to fit. An ask already at
+    /// the top of the `Decimal` range has no tick above it, and that is the
+    /// branch keeping `apply_spread` from emitting a quote it cannot form.
+    #[test]
+    fn test_unlock_reports_a_lift_that_cannot_fit() {
+        let tick = tick_size(2).expect("two decimal places is a representable scale");
+
+        // Locked at the top of the range: there is no tick above the ask.
+        assert_eq!(unlock(Positive::MAX, Positive::MAX, tick), None);
+
+        // An ordinary locked quote is lifted by exactly one tick, and the bid
+        // — the side carrying the information — does not move.
+        assert_eq!(
+            unlock(pos_or_panic!(0.01), pos_or_panic!(0.01), tick),
+            Some((pos_or_panic!(0.01), pos_or_panic!(0.02)))
+        );
+
+        // A quote that is already two-sided is returned untouched.
+        assert_eq!(
+            unlock(pos_or_panic!(0.01), pos_or_panic!(0.50), tick),
+            Some((pos_or_panic!(0.01), pos_or_panic!(0.50)))
+        );
+    }
+
+    /// And the same case reached through the public entry point: a quote whose
+    /// widened ask cannot be represented is left exactly as it was, not
+    /// half-written.
+    #[test]
+    fn test_apply_spread_leaves_an_unrepresentable_quote_untouched() {
+        let mut option_data = create_test_option_data();
+        option_data.call_bid = Some(Positive::MAX);
+        option_data.call_ask = Some(Positive::MAX);
+        let before = option_data.clone();
+
+        option_data.apply_spread(pos_or_panic!(0.6), 2);
+
+        assert_eq!(option_data.call_bid, before.call_bid);
+        assert_eq!(option_data.call_ask, before.call_ask);
     }
 
     #[test]
--- a/src/chains/optiondata.rs
+++ b/src/chains/optiondata.rs
@@ -225,12 +225,40 @@ fn tick_size(decimal_places: u32) -> Option<Positive> {
         .and_then(|tick| Positive::new_decimal(tick).ok())
 }
 
+/// Lifts `ask` one tick above `bid` when the two collide, so the thinnest
+/// market this generator emits is one tick wide.
+///
+/// Both sides are floored at the tick, so a row whose whole quote sits below
+/// the tick lands on it twice and comes out locked. A locked market is a real
+/// thing on an exchange, but a quote *generator* that produces one is wrong:
+/// a consumer marking at the touch would read a zero-width spread on the
+/// cheapest strikes, which is the opposite of what a thin market means. The
+/// bid is the side that carries the information — it is the floor itself —
+/// so the ask is the side that moves.
+///
+/// A quote that is already two-sided is returned untouched: only a collision
+/// moves anything.
+///
+/// Returns `None` when the lifted ask leaves the representable `Decimal`
+/// range; the lift is range-checked like every other step, never assumed to
+/// fit.
+#[inline]
+#[must_use]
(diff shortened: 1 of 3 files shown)


tests/property/quote_invariants_test.rs
```diff
@@ -82,7 +87,11 @@ proptest! {
         let call_bid = option_data.call_bid.expect("call bid is quoted");
         let call_ask = option_data.call_ask.expect("call ask is quoted");
         prop_assert!(call_bid >= floor, "bid {call_bid} below the tick {floor}");
-        prop_assert!(call_bid <= call_ask, "crossed quote {call_bid}/{call_ask}");
+        prop_assert!(call_bid < call_ask, "locked or crossed quote {call_bid}/{call_ask}");
+        prop_assert!(
+            call_ask.to_dec() - call_bid.to_dec() >= floor.to_dec(),
```
joaquinbejar: 🟠 **Warning**: This new assertion uses raw `Decimal` subtraction, with the same pattern at line 138 and raw multiplication at lines 159 and 162. `global_rules.md` requires checked financial arithmetic, so use `checked_sub` and `checked_mul` and make success part of the property setup.

src/chains/optiondata.rs
```diff
@@ -225,12 +225,40 @@ fn tick_size(decimal_places: u32) -> Option<Positive> {
...
+/// Returns `None` when the lifted ask leaves the representable `Decimal`
+/// range; the lift is range-checked like every other step, never assumed to
+/// fit.
+#[inline]
+#[must_use]
+fn unlock(bid: Positive, ask: Positive, tick: Positive) -> Option<(Positive, Positive)> {
+    if bid < ask {
+        return Some((bid, ask));
+    }
+    Some((bid, bid.checked_add(&tick).ok()?))
```
joaquinbejar: 🟠 **Warning**: This is the new failure path that keeps an unrepresentable ask from panicking, but none of the added tests reaches it. Add a direct `unlock(Positive::MAX, Positive::MAX, tick)` regression or an `apply_spread` case that proves the quote remains unchanged when the lift cannot fit.

joaquinbejar (commented):
The quote widening rule is sound and the targeted tests pass, but the new property code bypasses the mandatory checked Decimal arithmetic and the new overflow branch is not covered. REQUEST_CHANGES. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

tests/property/quote_invariants_test.rs
joaquinbejar: Fixed, the two width assertions use checked_sub and the two tick scalings use checked_mul, with the check succeeding folded into the assertion rather than assumed. Same for the pattern at the other line you flagged.

src/chains/optiondata.rs
joaquinbejar: Covered, and writing the regression found a second way into that branch worth mentioning. A tick finer than the last representable digit at the bid's magnitude is absorbed by the rounding, so Decimal::MAX + 0.01 comes back as Decimal::MAX and checked_add reported success while unlock returned the locked quote it exists to remove. The lift now has to actually separate the two sides, covered directly and through apply_spread.

joaquinbejar:
Both addressed in e167ded9. The property arithmetic is checked throughout, and the failed lift is covered from both directions, which surfaced a rounding case where the lift silently did nothing and left the quote locked.

joaquinbejar:
Both review points addressed in `e167ded9`, and the second one was a real defect rather than a coverage gap.

**`unlock`'s failure path did not work.** Writing the test asked for revealed that `Positive::MAX + 0.01` does not overflow: the sum needs 29 significant digits at scale 2, so `Decimal` drops the scale and rounds, and the result is `Positive::MAX` again. `checked_add(..).ok()?` reported success and handed back an ask level with the bid — the helper emitted the very locked quote it exists to prevent, at the top of the range.

The lift is now verified, not merely computed:

```rust
let lifted = bid.checked_add(&tick).ok()?;
(lifted > bid).then_some((bid, lifted))
```

so `bid < ask` holds for every value `unlock` returns, and a bid with no room above it leaves the quote untouched like every other unrepresentable step in the module. `test_unlock_reports_a_lift_that_cannot_fit` and `test_apply_spread_leaves_an_unrepresentable_quote_untouched` pin it at both levels; they fail against the previous commit.

**Checked arithmetic in the properties.** Both width assertions form their difference with `checked_sub` and make the check part of the assertion, and the two tick scalings go through `checked_mul` in a `ticks_of` helper instead of the raw operator.

4676 passed, 0 failed. `cargo clippy --all-targets --all-features --workspace -- -D warnings`, `cargo fmt --all --check` and `make scan-banned` clean.

Issue #453, opened 2026-08-29
Multiply merge is non-deterministic across runs


joaquinbejar:
## Overview

`Surface::merge` and `Curve::merge` implement the Multiply operation with a
rayon `reduce`. Rayon's chunking varies from run to run, and `Decimal`
multiplication rounds once a product needs more than 28 decimal places, so
merging three or more curves or surfaces can produce different last digits on
different runs of the same binary on the same inputs.

Non-deterministic financial output is its own class of problem: a value that
changes between runs cannot be reconciled, cached, or regression-tested at full
precision, and a downstream consumer diffing two runs sees phantom changes.

## Context

- `Surface::merge` and `Curve::merge`, the `Multiply` arm
- Pre-existing: the original code used `par_iter().product()`, which has the
  same property. #452 did not introduce it and deliberately did not change it.
- Found by `panic-hunter` while fixing the `Divide` reducer in #452, which had a
  worse defect in the same function — it discarded every operand but the
  leftmost, and swallowed checked-arithmetic errors with them.
- The Add and Subtract arms are already sequential (`d_sum_iter`), and Divide
  became sequential in #452, so Multiply is the last parallel reducer left.

## The decision to make

Making it sequential buys determinism and costs whatever the parallelism is
worth. Nobody has measured that, which is why it was not folded into a panic
sweep. The alternative — keeping rayon but making the fold associative-safe by
rounding at a fixed scale after each step — costs precision instead.

## Tasks

- [ ] Measure the current parallel Multiply with Criterion across realistic
      merge sizes
- [ ] Implement the sequential fold and measure it against that baseline
- [ ] Decide on the numbers: sequential, or parallel with a documented
      rounding rule that makes the result order-independent
- [ ] Add a determinism test: merging the same three objects repeatedly must
      produce bit-identical output

## Acceptance criteria

- [ ] Merging three or more curves or surfaces is bit-reproducible across runs
- [ ] The Criterion delta is on record in the PR, whichever way it goes
- [ ] Existing merge tests unchanged

PR #473, opened 2026-08-30
Give the surface projection a name and a type that fit it (#466)

### LINKED ISSUE #466: Duplicate abscissae in Curve and Surface have no defined rule

joaquinbejar:
## Summary

`Curve` and `Surface` store points in a `BTreeSet`, and duplicate abscissae get
there legitimately — projecting a grid onto one axis produces several points
sharing an `x` by construction. The crate assumed two contradictory things
about that, and #450 fixing the `Ord`/`Eq` contract exposed the disagreement
rather than resolving it.

The decision this PR makes: **`Curve` stays a function of its abscissa, and the
projection gets a name and a type that admit it is not one.**

`Surface::get_curve` → `Surface::project_onto`, returning `Vec<Point2D>`
instead of `Curve`. Name and type were saying the same wrong thing, so they
change together.

## Why a `Vec`, not a newtype and not the point set

- It preserves cardinality exactly: `project_onto(axis).len() ==
  self.points.len()`, always. A set does not — two surface points that agree on
  the two surviving coordinates project onto the *identical* `Point2D`, and a
  `BTreeSet` keeps one. With the `Ord`/`Eq` contract fixed by #450 that
  coincident-image case is the only remaining silent loss, and it is precisely
  the one a set cannot express.
- A `Vec` already says "sequence of points" rather than "function of `x`". A
  newtype would add a public type, its docs, its `Serialize`/`ToSchema`
  question and its accessors, and would carry no invariant a `Vec` does not.
- Contents stay sorted ascending by the projected `(x, y)`, which is exactly
  what the `BTreeSet` gave. Only its deduplication is gone, so the ordering
  semantics do not change silently along with the type.

`"greatest ordinate wins"` was not adopted: #450 rejected it precisely because
it would bless `collect`'s accidental behaviour as the projection's semantics.

## The rule is documented as a contract, not as an enforced invariant

`points` is a `pub` field on both types, so a struct literal reaches a
duplicate-carrying state without passing through any constructor.
Normalization there could only ever be a convenience, never an invariant — and
the docs say that, instead of claiming a guarantee `new` does not make. The
rule is stated on `Curve::new`, `Curve::from_vector`, `Curve::merge`,
`get_point`, `contains_point`, `get_values`, `merge_axis_interpolate`,
`Surface::new`, the four interpolation traits and `find_bracket_points`, each
with what that consumer actually does when the rule is broken.

## A behaviour change beyond the return type

"No consumer may silently pick a survivor" was violated in the interpolators,
not only in the prose. `cubic_interpolate` and `spline_interpolate`
short-circuited an exact `x` with `points.iter().find(|p| p.x == x)`, returning
the **lowest** ordinate of a stack, and `linear_interpolate` reached the same
value through a bracket ending there. All three now go through

```rust
fn exact_point_at(&self, x: Decimal) -> Result<Option<Point2D>, InterpolationError>
```

which returns `InterpolationError::DegenerateInterval` when several points
share `x` — the same answer the arithmetic already gives for a zero-width
bracket (#452). No new error variant and no new rule. A curve with one point
per abscissa is unaffected.

Two consumers keep reading the first match, now documented as "the lowest
ordinate" with `get_values` named as the multi-valued reader: `get_point` and
`merge_axis_interpolate`. The exact-match branch of `bilinear_interpolate`
keeps its old behaviour and is documented as the exception, because that method
is reworked in the next PR of this chain (#451) and moving it twice would make
both diffs harder to check.

## Tests

The three projection tests assert exact contents instead of membership probes —
the test surface projects five points and `collect` used to return three — and
a fourth covers two surface points projecting onto the same 2D point. A new
`tests_duplicate_abscissa` module pins the interpolator behaviour on both
sides: a stacked abscissa errors, a single-ordinate curve is untouched.

4681 passed, 0 failed. `cargo fmt --all --check`, `cargo clippy --all-targets
--all-features --workspace -- -D warnings`, `make scan-banned` and
`make public-api-check` all clean. `cargo doc` emits one warning fewer than
`main` — the redundant explicit link in `surface.rs` went with the rewrite.

## Public API

```diff
-pub fn Surface::get_curve(&self, Axis) -> Curve
+pub fn Surface::project_onto(&self, Axis) -> Vec<Point2D>
```

`public-api/optionstratlib.txt` and `CHANGELOG.md` updated in this commit.

## Stack

Chain B of four independent chains off `main`.

- Stack: **#466** (this one) → #451 → #453. Independent of chain A (#470 →
  #469 → #463 → #471), chain C (#462), #459, and the #442 capstone.
- Base branch: `main`.
- The next two PRs in this chain branch off this one, so read their diffs
  against this branch rather than against `main`.

Closes #466

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,45 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **`Surface::get_curve` is renamed to `Surface::project_onto` and returns
+  `Vec<Point2D>` instead of `Curve`, which is a breaking change** (#466).
+  One change with one reason: the method is a projection, and both its name
+  and its type said it was a curve. Projecting a surface onto one axis is
+  multi-valued by construction: every row of a grid contributes a different
+  height above the same projected abscissa, and two rows that agree on the
+  two surviving coordinates project onto the very same point. A `Curve` is a
+  function of its abscissa and stores its points in a `BTreeSet`, so it
+  silently dropped the collisions. The vector keeps every point:
+  `surface.project_onto(axis).len() == surface.points.len()`, always.
+  Contents are sorted ascending by the projected `(x, y)` pair, which is the
+  order the `BTreeSet` gave; only its deduplication is gone.
+
+  The method has no production callers and its `Axis` parameter type is not
+  re-exported, so it is uncallable from outside the crate. To migrate inside
+  it, aggregate the ordinates sharing an abscissa to one value, then build the
+  `Curve`; `Curve::from_vector(surface.project_onto(axis))` reproduces the old
+  behaviour, losses included. The note on `Surface::get_curve` under the #450
+  entry below describes the method by its former name.
+
+- **Interpolating a curve at a repeated abscissa now fails instead of
+  returning the lowest ordinate stacked there** (#466). `Curve` is documented
+  as a function of its abscissa, but nothing enforces it (`points` is a `pub`
+  field, so no constructor could). `linear_interpolate`, `cubic_interpolate`
+  and `spline_interpolate` return `InterpolationError::DegenerateInterval`
+  when several points share the requested `x`, rather than picking one of
+  them: the value there is not defined and no choice among them is less
+  arbitrary than another. A curve with one point per abscissa is unaffected,
+  and `AxisOperations::get_values` still reads every ordinate at an abscissa.
+  The exact-match branch of `Curve::bilinear_interpolate` still returns the
+  lowest ordinate; it is being reworked under #451.
+
+  The one-point-per-abscissa rule, and what each consumer does when it is
+  broken, are now stated on `Curve::new`, `Curve::from_vector`,
+  `Curve::merge`, `get_point`, `contains_point`, `get_values`,
+  `merge_axis_interpolate`, the four interpolation traits and
+  `find_bracket_points`, with the matching one-height-per-xy-coordinate rule
+  on `Surface::new`.
+
 - **Two `ProtectivePut` methods are now fallible, which is a breaking change**
   (#460). `total_fees` returns `Result<Positive, PositiveError>` instead of
   `Positive`, and `protection_level` returns
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **The crate moves to 0.21.0, which is the breaking bump for a `0.x`
+  version.** `cargo-semver-checks` compares against the published 0.20.0 and
+  rejected the rename below under `inherent_method_missing`, since a version
+  unchanged from its baseline is read as a minor bump and a minor bump may not
+  remove a public method. Every earlier change in this cycle was a return type
+  moving to `Result`, which no lint covers, so this is the first one the check
+  could see.
+
 - **`Surface::get_curve` is renamed to `Surface::project_onto` and returns
   `Vec<Point2D>` instead of `Curve`, which is a breaking change** (#466).
   One change with one reason: the method is a projection, and both its name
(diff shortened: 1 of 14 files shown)


joaquinbejar (commented):
APPROVE because the renamed projection preserves cardinality and ordering, the breaking migration is documented with the required 0.21.0 bump, and the duplicate abscissa behavior is covered without changing valid single valued curves. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

joaquinbejar:
Thanks. The 0.21.0 bump is in e5694e7e, with the four version strings in the lib.rs doc header and README.md regenerated from them; cargo semver-checks reads it locally as 0.20.0 -> 0.21.0 (major change) and passes. Merging.

PR #474, opened 2026-08-30
Price the arithmetic Asian on a divided difference, not a cancellation (#462)

### LINKED ISSUE #462: Arithmetic Asian loses all precision at short maturities

joaquinbejar:
## Summary

The general branch of the arithmetic Asian priced every non-degenerate
contract, and it lost all its precision as maturity shortened — returning
confident wrong values rather than errors. At 86 seconds to expiry it
overpriced by a factor of **612**.

`term1` and `term2` each grow as `2S²/(a·c·T²)` and cancel down to about `S²`,
while the quantity extracted from their difference is `σ²_asian·T`, which
*shrinks* with `T`. At 14 minutes the terms are around `2.8e15` and the signal
is `3.7e-8`. Below `1e-5` days it collapsed the other way, to approximately
zero.

## The rewrite

Writing `x = bT`, `z = cT` and `y = aT = z − x`, the entire second moment is
carried by one function:

```text
φ(w) = (e^w − 1) / w = ∫₀¹ e^{ws} ds,   φ(0) = 1

M2 = S² g(x, z),   g(x, z) = 2 (φ(z) − φ(x)) / (z − x)
```

twice the first divided difference of `φ`. It is symmetric, entire in both
arguments, and **each of the three apparent poles of the closed form is an
ordinary point of it**: `b = 0` is `x = 0` and `c = 0` is `z = 0`, both covered
by `φ(0) = 1`; `b = −σ²` is `z = x`, where a divided difference is the
derivative `2φ'(x)`.

So the four rate-keyed branches — the general one plus the three removable
limits, two of them added by #445 — collapse into one expression with two
numerical fallbacks. Nothing is recovered from a difference of two terms larger
than the answer, which is the property the old form lacked.

## Both thresholds are measured, not chosen

Against 50-digit `mpmath` quadrature on
`E[A²] = (2S²/T²) ∫₀^T e^{bu} ∫₀^u e^{at} dt du`, twelve maturities per decade.

**Series below `max(|x|,|z|) = 1e-3`** — relative error of the second moment:

| `u` | 6-term series | divided difference |
|---|---|---|
| `3.1e-3` | `7.0e-20` | `2.8e-24` |
| `2.1e-3` | `7.0e-21` | `5.4e-23` |
| `1.5e-3` | `7.0e-22` | `2.5e-22` |
| `9.9e-4` | `7.0e-23` | `6.2e-22` |
| `8.2e-4` | `2.2e-23` | `6.9e-22` |

The series falls as `u⁶`; the closed form rises as `u⁻²`, the `1/y`
amplification of the rounding `e^w` leaves inside `φ(z) − φ(x)`. They cross at
`u ≈ 1.2e-3`, so the threshold is `1e-3`.

**Midpoint derivative below `|y| = 1e-8`** — sweeping `a` through zero at
`b = a − σ²`, `σ = 20%`, `T = 1`, which is the `b = −σ²` boundary contract:

| `\|y\|` | midpoint `2φ'(m)` | divided difference |
|---|---|---|
| `1e-6` | `2.1e-14` | `6.4e-16` |
| `1e-7` | `2.1e-16` | `6.4e-17` |
| `1e-8` | `2.1e-18` | `7.0e-18` |
| `1e-9` | `2.1e-20` | `4.5e-18` |
| `1e-10` | `2.1e-22` | `1.6e-17` |

The midpoint truncation falls as `y²/48`; the divided difference falls with `y`
too, until the `φ(z) − φ(x)` rounding floor turns it round near `1e-9`. They
cross just under `3e-8`, so the threshold is `1e-8`.

Both tables are in the doc comments of `M2_SERIES_THRESHOLD` and
`M2_MIDPOINT_THRESHOLD`, so the next person to move one can see what it costs.

## The table, regenerated

Asserted bound `1e-9`; worst measured **`6.9e-12`**, and it comes from the
`f64` normal CDF behind `big_n` rather than from the moment.

| maturity | crate now | reference (50 dp) | relative | was |
|---|---|---|---|---|
| 0.5 days | 0.171855815411 | 0.171855815411 | 1.6e-13 | 2.0e-9 |
| 0.2 days | 0.108377120846 | 0.108377120846 | 2.9e-13 | 2.6e-7 |
| 0.1 days | 0.076521744333 | 0.076521744333 | 3.2e-13 | 1.3e-5 |
| 0.05 days | 0.054052666108 | 0.054052666108 | 4.0e-13 | −8.8e-4 |
| 0.02 days | 0.034154202089 | 0.034154202089 | 5.1e-13 | +6.6e-3 |
| 0.01 days | 0.024139351974 | 0.024139351974 | 2.1e-12 | **+73%** |
| 0.005 days | 0.017063436757 | 0.017063436757 | 1.7e-12 | **+1277%** |
| 0.002 days | 0.010788684989 | 0.010788684989 | 3.5e-12 | **80×** |
| 0.001 days | 0.007627618506 | 0.007627618506 | 6.9e-12 | **612×** |

The reference was rebuilt independently of the implementation, and the
quadrature agrees with the closed form to `1e-38` at every row, so it assumes
neither. Put-call parity is asserted separately at 86 seconds, where
`e^{−rT}(M1 − K) = 5.4794510539814525e-6` is small enough that the identity
constrains the last digits of both legs rather than restating the formula.

## Ordinary maturities

Bit-for-bit identical at **7, 30, 90, 182.5 and 365 days**, call and put,
verified by diffing the printed prices against `main`.

One does move, and it is worth stating rather than burying: **at 1 day the
price changes by a relative `2.5e-10`**. It is a correction, not a regression —
the old value sat `2.5e-10` from the 50-digit reference and the new one sits
`8.5e-14` from it:

| | old | new | reference |
|---|---|---|---|
| 1 day call | 0.2438268020005283 | 0.2438268019395450 | 0.2438268019395657 |
| relative error | 2.5e-10 | 8.5e-14 | |

## Tests

No existing assertion, tolerance or expected value changed, and no test was
removed — the diff over the test module adds only.

Three test **doc comments** did change. Two described the cancellation between
two large terms and the `1e-6` straddle of the old series threshold; neither
mechanism exists any more, and prose asserting a mechanism the code no longer
has reads as a claim about the current implementation. The old and new
sentences are in the commit body.

New: `test_arithmetic_asian_short_maturity_matches_quadrature` (the nine rows),
`..._holds_put_call_parity`, `..._series_threshold_is_continuous`,
`..._midpoint_threshold_is_continuous` (both switchovers are a change of method
and not a discontinuity in the price surface), and
`..._removable_boundaries_hold_at_short_maturity`.

4677 passed, 0 failed. `cargo fmt --all --check`, `cargo clippy --all-targets
--all-features --workspace -- -D warnings`, `make scan-banned` and
`make public-api-check` clean. Public API unchanged.

One incidental fix: the series doc block is now fenced as `text`. Indented,
rustdoc took it for Rust and tried to compile `Σ_{n ≥ 0}` as a doctest.

## Stack

Chain C of four independent chains off `main`; this is its only PR, so the diff
reads against `main` directly.

- Stack: **#462** (this one) — independent of chain A (#470 → #469 → #463 →
  #471), chain B (#466 → #451 → #453), #459, and the #442 capstone.
- Base branch: `main`.

Closes #462

### DIFF
--- a/src/pricing/asian.rs
+++ b/src/pricing/asian.rs
@@ -469,16 +469,33 @@ fn arithmetic_asian_price(option: &Options) -> Result<Decimal, PricingError> {
 ///
 /// The building block of both Turnbull-Wakeman moments: `M1 = S φ(bT)` and
 /// `M2 = 2 S² (φ(cT) - φ(bT)) / (aT)`. The `w = 0` singularity is removable
-/// and the limit is `∫₀^1 ds = 1`, which the guard returns exactly rather
-/// than dividing by the rate that vanished.
+/// and the limit is `∫₀^1 ds = 1`, which the series below returns exactly
+/// rather than dividing by the rate that vanished.
+///
+/// Small `w` is summed rather than evaluated, and at the *same*
+/// [`M2_SERIES_THRESHOLD`] the second moment uses. Sharing the constant is
+/// the point, not a convenience: the two moments have to change regime
+/// together. When they did not, `M1` collapsed to exactly `S` below a cutoff
+/// of its own while `M2` went on evaluating the real carry, and the price
+/// jumped twofold across that cutoff — `S = K = 100`, `σ = 1e-5`, `T = 1`
+/// priced at `4.6066e-4` for `b = 9.99999e-11` and at `2.3033e-4` for
+/// `b = 1e-10`, for a `1e-16` change in the rate.
+///
+/// The closed form loses the leading digits of `e^w` to the `1 /`
+/// amplification exactly as it does inside `φ(z) - φ(x)`: `e^w` rounds at
+/// `Decimal`'s scale, so `(e^w - 1) / w` carries a relative error of about
+/// `u / |w|` — `1e-18` at `w = 1e-10`, but `1e-8` by `w = 1e-20`, past the
+/// `1e-9` bound the issue sets. The series has no such subtraction and
+/// truncates at `w⁶ / 7!`, a relative `2e-4 w⁶`, which is `2e-22` at the
+/// threshold.
 ///
 /// # Errors
 ///
 /// Returns [`PricingError::Decimal`] when `e^w` leaves the representable
 /// `Decimal` range.
 fn growth_average(w: Decimal) -> Result<Decimal, PricingError> {
-    if w.is_zero() {
-        return Ok(Decimal::ONE);
+    if w.abs() < M2_SERIES_THRESHOLD {
+        return growth_average_series(w);
     }
     Ok(d_div(
         d_sub(
@@ -488,6 +505,42 @@ fn growth_average(w: Decimal) -> Result<Decimal, PricingError> {
         )?,
         w,
         "pricing::asian::growth_average",
+    )?)
+}
+
+/// `φ(w)` summed as `Σ_{n ≥ 0} wⁿ / (n+1)! = 1 + w/2 + w²/6 + …`, which
+/// carries the leading `1` explicitly instead of recovering it from
+/// `e^w - 1`.
+///
+/// `w = 0` gives exactly `1` here, every term after the first vanishing, so
+/// the removable singularity needs no guard of its own.
+///
+/// # Errors
+///
+/// Returns [`PricingError::Decimal`] when a power, a factorial or the running
+/// sum leaves the representable `Decimal` range.
+fn growth_average_series(w: Decimal) -> Result<Decimal, PricingError> {
+    // The `n = 0` term is `w⁰ / 1! = 1`, the `1` the correction is added to.
+    let mut w_power = Decimal::ONE;
+    let mut factorial = Decimal::ONE;
+    let mut correction = Decimal::ZERO;
+    for n in 1..M2_SERIES_TERMS {
+        w_power = d_mul(w_power, w, "pricing::asian::phi::series::power")?;
+        factorial = d_mul(
+            factorial,
+            Decimal::from(n + 1),
+            "pricing::asian::phi::series::factorial",
+        )?;
+        correction = d_add(
+            correction,
+            d_div(w_power, factorial, "pricing::asian::phi::series::term")?,
+            "pricing::asian::phi::series::correction",
+        )?;
+    }
+    Ok(d_add(
+        Decimal::ONE,
+        correction,
+        "pricing::asian::phi::series",
     )?)
 }
 
@@ -649,9 +702,11 @@ fn second_moment_series(x: Decimal, z: Decimal) -> Result<Decimal, PricingError>
 /// Returns [`PricingError::Decimal`] when `e^{bT}` or the product with `S`
 /// leaves the representable `Decimal` range.
 fn arithmetic_average_forward(s: Decimal, b: Decimal, t: Decimal) -> Result<Decimal, PricingError> {
-    if b.abs() < dec!(1e-10) {
-        return Ok(s);
-    }
+    // No cutoff of its own. A `|b| < 1e-10` shortcut to exactly `S` used to
+    // live here, which put this moment in a different limit regime from the
+    // second one and made the price discontinuous across the boundary.
+    // `growth_average` handles a vanishing `bT` by summing its series, and
+    // returns exactly `1` at zero.
     let bt = d_mul(b, t, "pricing::asian::average_forward::bt")?;
     Ok(d_mul(
         s,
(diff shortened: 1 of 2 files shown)


src/pricing/asian.rs
```diff
@@ -276,230 +323,42 @@ fn arithmetic_asian_price(option: &Options) -> Result<Decimal, PricingError> {
...
+        "pricing::asian::arithmetic::m2::total_rate",
+    )?;
+    let y = d_mul(
+        b_plus_var,
+        t_dec,
+        "pricing::asian::arithmetic::m2::inner_rate",
+    )?;
+    let m2 = d_mul(
+        s_sq,
+        scaled_second_moment(x, z, y)?,
```
joaquinbejar, line 359: 🟠 **Warning**: The new second moment path evaluates the actual carry for every nonzero `b`, but `arithmetic_average_forward` still returns exactly `S` whenever `|b| < 1e-10`. That splits the two moments at the old branch boundary and creates a real price discontinuity. With `S = K = 100`, `sigma = 1e-5`, `T = 1` year and `q = 0`, rates `9.99999e-11` and `1e-10` price at `0.0004606586931539341767481726` and `0.0002303319329999997499938484` respectively, a twofold jump for a `1e-16` rate change. Make the first moment use a stable small argument treatment in `bT`, or otherwise keep both moments on the same limit regime, and add a regression across this boundary.

joaquinbejar (commented):
REQUEST_CHANGES because the rewritten second moment no longer shares the first moment small carry branch, producing a twofold discontinuity at the existing 1e-10 boundary for a valid low volatility contract. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

src/pricing/asian.rs
joaquinbejar: Confirmed and fixed, and your repro was exact. phi now has its own series regime at the same M2_SERIES_THRESHOLD the second moment switches on, and the first moment's 1e-10 cutoff is gone, so the two change regime together by construction rather than by coincidence. Your two rates now price at 0.00023033193299749981 and 0.00023033193299999983, a relative 1.1e-11 apart instead of a factor of two, with a regression across the boundary.

joaquinbejar:
Fixed in ddaf051a. The seam was the first moment keeping a cutoff the second one did not have; phi is now summed as a series below the same threshold and the cutoff is gone, which also holds phi where the closed form cannot since (e^w - 1)/w carries a relative u/|w| and passes the 1e-9 bound by w = 1e-20. Rebased on main, which now carries the 0.21.0 bump and the coverage fix.

PR #475, opened 2026-08-30
Make the model layer the strategies run through panic-free (#470)

### LINKED ISSUE #470: Panic sweep: the model layer reached from strategies

joaquinbejar:
## Summary

Three defects in the model layer that the composition sweep (#468) reached but
could not fix in scope. Together they were **126 of the 363 panics that sweep
left behind**, and they are why its property test had to bound three of its
generators.

## 1. `Position` cost arithmetic

`total_cost` and `fees` accumulated the premium and the two fees and scaled the
result by the contract quantity with the raw `Positive` operators — inside
functions that already returned `Result<Positive, PositionError>` and so had
somewhere to put the failure all along.

Every step is checked now. `PositionError` grows a `PositiveError` variant so
the cause is carried rather than flattened into a message; the alternative the
issue offered, mapping through `PositionError::invalid_position`, would have
thrown the typed cause away at the only place it is known.

Both doc comments claimed the function was "currently infallible". That was
true when written and stops being true here, so the paragraphs are gone rather
than left to mislead.

## 2. Calendar overflow through `get_date()`

`ExpirationDate::Days(1e9)` aborted with `` `DateTime + TimeDelta` overflowed ``
inside `expiration_date` 0.3.0, and **every strategy reaches it** through
`calculate_pnl_at_expiration`. This was the single largest residual.

#441 guarded the one call site in `OptionChain::build_chain`. This is the
general case, so the guard moves to `src/model/expiration.rs` as
`resolve_expiration_date` and `reject_unrepresentable_expiration`, and
`build_chain` calls the shared one instead of keeping its own copy.

The resolution mirrors the upstream expression step by step — the same
`Positive` → `i64` conversion, the same day-to-span conversion, the same
addition — each through its checked counterpart. The instant returned is
therefore identical for every input the panicking form accepted; only the
inputs it used to abort on behave differently, and those now report.

## 3. `mean_and_std`

```rust
vec.iter().sum::<Positive>() / vec.len() as f64
```

Overflowed on the sum and divided by zero on an empty vector. It returns
`Result<(Positive, Positive), PositiveError>` now, and the sixteen
`get_profit_ranges` implementations that average their leg volatilities through
it propagate.

The squared deviations are still taken in `f64`, exactly as before, so the
returned figures are unchanged for every sample the panicking form accepted.
What changes is the reporting: a deviation that overflows `f64` used to
collapse to zero through `unwrap_or(Positive::ZERO)` and now surfaces.

## What this unlocks in the property test

The strategies sweep can now drive what it had to avoid:

- **premium and fees** — unbounded, `Positive::MAX` included
- **volatility** — unbounded, `Positive::MAX` included
- **expirations** — a billion days, `1e15` days and `Positive::MAX` days: past
  the calendar, past `TimeDelta`, and past the `i64` day count respectively

The three doc comments explaining why each bound existed are gone with the
bounds.

**One bound survives**, narrowed and renamed to `spot_leg_money` so it says
what it actually is: the spot-leg strategies run through
`LegAble::pnl_at_price`, `LegAble::total_cost`, `LegAble::fees` and
`Collar::net_premium`, all of which return a bare `Decimal` or `Positive` and
have nowhere to report an overflow. Lifting it needs those signatures to gain
an error channel — which is #471, the last PR of this chain. The comment on the
generator names the four functions and their line numbers rather than saying
"outside this sweep".

## Tests

4684 passed, 0 failed. 66 property tests, up from 63 — a new
`tests/property/model_panic_freedom_test.rs` drives the three fixed sites
directly. Existing tests pass unedited.

`cargo fmt --all --check`, `cargo clippy --all-targets --all-features
--workspace -- -D warnings`, `make scan-banned` and `make public-api-check`
clean.

## Public API

```diff
-pub fn model::utils::mean_and_std(Vec<Positive>) -> (Positive, Positive)
+pub fn model::utils::mean_and_std(Vec<Positive>) -> Result<(Positive, Positive), PositiveError>
+pub enum variant PositionError::PositiveError(PositiveError)
+pub fn model::resolve_expiration_date(&ExpirationDate) -> Result<DateTime<Utc>, ExpirationDateError>
+pub fn model::reject_unrepresentable_expiration(&ExpirationDate) -> Result<(), ExpirationDateError>
```

Snapshot regenerated and committed.

**Expect `semver` to fail on this PR until the 0.21.0 bump lands.** `PositionError`
is not `#[non_exhaustive]`, so adding a variant trips `enum_variant_added`
against a `Cargo.toml` still reading 0.20.0 — the same reason #473 went red,
diagnosed there. Rebasing onto a `main` that carries the bump clears it; no
change to this diff is needed. The `mean_and_std` retyping has no lint of its
own, which is the gap #455 exists for, so it is declared here instead.

## Stack

Chain A of four independent chains off `main`, and the bottom of its chain.

- Stack: **#470** (this one) → #469 → #463 → #471. Independent of chain B
- Base branch: `main`.
- The next three PRs in this chain branch off this one, so read their diffs
  against this branch rather than against `main`.

Closes #470

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `merge_axis_interpolate`, the four interpolation traits and
   `find_bracket_points`, with the matching one-height-per-xy-coordinate rule
   on `Surface::new`.
+- **`model::utils::mean_and_std` is now fallible, which is a breaking change**
+  (#470). It returns `Result<(Positive, Positive), PositiveError>` instead of
+  `(Positive, Positive)`. It summed with the raw `Positive` operator, which
+  panics on overflow, and divided by `vec.len() as f64`, which divides by zero
+  on an empty sample; every `get_profit_ranges` implementation averages its leg
+  volatilities through it. The returned figures are unchanged for every sample
+  the panicking form accepted — the squared deviations are still taken in
+  `f64` — but a deviation that overflows is now reported instead of collapsing
+  to zero through `unwrap_or(Positive::ZERO)`. To migrate, add `?` where the
+  value feeds a fallible function.
+
+  `PositionError` gains a `PositiveError` variant in the same change, so
+  `Position::total_cost` and `Position::fees` can carry the cause of a
+  `Positive` overflow instead of flattening it into a message. Both already
+  returned `Result`, so their signatures are unchanged. Matching exhaustively
+  on `PositionError` needs a new arm.
+
+  `model::resolve_expiration_date` and `model::reject_unrepresentable_expiration`
+  are new. `ExpirationDate::get_date()` aborts with
+  `` `DateTime + TimeDelta` overflowed `` for a day count no calendar can
+  represent, and every strategy reaches it through
+  `calculate_pnl_at_expiration`; these resolve or reject it instead. The guard
+  #441 added inside `OptionChain::build_chain` is replaced by a call to the
+  shared one. The instant returned is identical for every input the panicking
+  form accepted.
 
 - **Two `ProtectivePut` methods are now fallible, which is a breaking change**
   (#460). `total_fees` returns `Result<Positive, PositiveError>` instead of
--- a/src/model/expiration.rs
+++ b/src/model/expiration.rs
@@ -1,7 +1,191 @@
-/// Re-export of `ExpirationDate` from the standalone `expiration_date` crate.
-///
-/// This module re-exports the `ExpirationDate` enum and its error type from the
-/// external `expiration_date` crate, which provides all the core functionality
-/// for handling financial instrument expiration dates.
+//! Re-export of `ExpirationDate` from the standalone `expiration_date` crate.
+//!
+//! This module re-exports the `ExpirationDate` enum and its error type from the
+//! external `expiration_date` crate, which provides all the core functionality
+//! for handling financial instrument expiration dates.
+//!
+//! It also owns the single calendar-overflow guard the rest of the crate uses.
+//! The `expiration_date` crate resolves a relative `Days(n)` with the `+`
+//! operator on `DateTime<Utc>` and with `TimeDelta::days`, both of which abort
+//! the process on overflow instead of reporting it, so every call site in this
+//! library goes through [`resolve_expiration_date`] or
+//! [`reject_unrepresentable_expiration`] rather than calling `get_date` and
+//! friends directly.
+
+use chrono::{DateTime, TimeDelta, Utc};
 pub use expiration_date::ExpirationDate;
 pub use expiration_date::error::ExpirationDateError;
+use positive::Positive;
+
+/// Hour of the fixed resolution base used by
+/// `ExpirationDate::get_date_with_options(true)` and therefore by
+/// `ExpirationDate::get_date_string`.
+const FIXED_BASE_HOUR: u32 = 18;
+
+/// Minute of the fixed resolution base. See [`FIXED_BASE_HOUR`].
+const FIXED_BASE_MINUTE: u32 = 30;
+
+/// Reports a day count that no calendar instant can represent.
+#[cold]
+#[inline(never)]
+fn unrepresentable(days: Positive) -> ExpirationDateError {
+    ExpirationDateError::ArithmeticOverflow(format!(
+        "expiration of {days} days is outside the representable calendar range"
+    ))
+}
+
+/// Adds a whole number of days to `base` without aborting on overflow.
+///
+/// Mirrors `base + Duration::days(i64::try_from(days)?)`, the expression the
+/// `expiration_date` crate uses, step by step: the same `Positive` to `i64`
+/// conversion, the same day-to-span conversion and the same addition, each
+/// through its checked counterpart. The instant returned is therefore
+/// identical for every input the panicking form would have accepted.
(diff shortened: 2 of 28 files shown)


src/model/expiration.rs
```diff
@@ -1,7 +1,191 @@
...
+//! external `expiration_date` crate, which provides all the core functionality
+//! for handling financial instrument expiration dates.
+//!
+//! It also owns the single calendar-overflow guard the rest of the crate uses.
+//! The `expiration_date` crate resolves a relative `Days(n)` with the `+`
+//! operator on `DateTime<Utc>` and with `TimeDelta::days`, both of which abort
+//! the process on overflow instead of reporting it, so every call site in this
+//! library goes through [`resolve_expiration_date`] or
+//! [`reject_unrepresentable_expiration`] rather than calling `get_date` and
+//! friends directly.
```
joaquinbejar, line 13: 🟠 **Warning**: The new guard is not yet the single route used by the model layer. `Options::calculate_pnl` and `Options::calculate_pnl_at_expiration` still call `ExpirationDate::get_date()` directly in `src/model/option.rs`, and `FuturePosition::expiration_timestamp` and `FuturePosition::is_expired` do the same. A public `Options` value with `Days(1_000_000_000)` therefore still aborts in `calculate_pnl_at_expiration`, so the no public panic acceptance criterion and this module level guarantee are not met. Route those production call sites through `resolve_expiration_date`, or explicitly narrow the scope and guarantee.

CHANGELOG.md
```diff
@@ -9,6 +9,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **`model::utils::mean_and_std` is now fallible, which is a breaking change**
```
joaquinbejar, line 74: 🟠 **Warning**: This head changes the public `mean_and_std` signature and adds an exhaustive `PositionError` variant while `Cargo.toml` remains at 0.20.0, so both semver jobs correctly fail. Rebase onto the 0.21.0 bump before this revision can be merge ready, or carry the appropriate version update here. Depending on a future merge of #473 does not make the current head satisfy the required checks.

joaquinbejar (commented):
REQUEST_CHANGES because public model APIs still reach the panicking expiration resolver and this head introduces breaking public API changes while the package version remains 0.20.0, leaving the semver checks red. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

src/model/expiration.rs
joaquinbejar: Fixed in 54294679, all four go through resolve_expiration_date now. Options::calculate_pnl and calculate_pnl_at_expiration propagate the typed error, and the two Expirable methods keep their infallible signatures and their fallbacks, which were never reached for that input because the process was already gone. A regression drives both P&L entry points on Days(1_000_000_000).

CHANGELOG.md
joaquinbejar: Right that a future merge does not make this head green. The 0.21.0 bump landed in #473 and this branch is rebased onto it, so the version is in the head under review rather than promised, and both semver jobs pass.

joaquinbejar:
Both addressed in 54294679, rebased onto the 0.21.0 main. The four remaining call sites resolve through resolve_expiration_date, so the module guarantee and the no public panic criterion hold, and the version is in this head rather than depending on another PR.

PR #477, opened 2026-08-30
Define bilinear interpolation near the upper boundary (#451)

### LINKED ISSUE #451: Define bilinear interpolation near the upper boundary

joaquinbejar:
## Summary

`Curve::bilinear_interpolate` read a four-point window starting at the bracket
index, but `find_bracket_points` only guarantees `i + 1`, so the window ran off
the end for any `x` in the **last three intervals**. #446 made that return
`InterpolationError::Bilinear` instead of aborting — the conservative call,
which changed no working result but left the method unusable across most of its
domain.

It now answers everywhere, and the rule was validated against a reference
rather than asserted.

## The rule: clamp the far edge, never the near one

The cell is built from two segments. The **near edge** is the segment that
brackets `x` and is never clamped. The **far edge** is the segment two
positions on, clamped to the curve's last segment where it would run off.

## Why not the cubic precedent

The issue proposed following `cubic_interpolate`, which clamps its whole window
to `[len-4, len-3, len-2, len-1]`. **That is wrong here, and the diff documents
why under its own heading**, because the next person will reach for it exactly
as the issue did.

The window start also supplies the denominator of `t = (x − x[i]) / (x[i+1] −
x[i])`. Clamping the start pushes `t` past `1`, so both edges become linear
*extrapolations* of a cell that does not contain `x`. The weight identity

```text
y = (1−t)/2·y[i] + t/2·y[i+1] + (1−t)/2·y[f] + t/2·y[f+1]
```

sums to one for any `t`, but is non-negative only for `t ∈ [0, 1]`. Bilinear
interpolation is by definition a convex combination of its cell's corners, and
the start-clamp gives that up.

Concretely, on `(0,0), (1,1), (2,4), (3,9)` the start-clamp returns **`9.5` at
`x = 2.5`** — above every ordinate on the curve. Clamping the far edge keeps
`t` in `[0, 1]`, keeps the segment holding `x` in the cell, and keeps the
answer inside the range the four ordinates span, which
`test_bilinear_interpolate_stays_within_the_ordinate_range` pins.

## What the boundary behaviour is, stated plainly

- On the **final** segment the two edges coincide, so the result is the linear
  interpolant there.
- On the **second-to-last** segment the far edge is the immediately following
  segment rather than the one after it.
- The clamp reaches only those two segments. Every abscissa the method already
  answered **keeps its answer digit for digit**, verified segment by segment
  against the pre-fix rule and pinned by
  `test_bilinear_interpolate_interior_is_unchanged_by_the_clamp`.
- The result jumps at every sample: the exact-match branch returns the sample
  and neither one-sided limit matches it. **That jump is as old as the method**
  and owes nothing to this change — `test_bilinear_interpolate_jump_at_a_sample_predates_the_clamp`
  demonstrates it at an interior sample the clamp never touches. At the last
  two samples the two one-sided limits also disagree with each other, because
  the clamp gives the segments either side different far edges; that part is
  new and `test_bilinear_interpolate_jumps_where_the_clamp_starts` pins it.

`InterpolationError::Bilinear`'s doc comment now points at this resolution.

## The exact-match branch

It goes through `Curve::exact_point_at` now, so a stack of ordinates at the
queried abscissa gives `InterpolationError::DegenerateInterval` instead of the
lowest of them.

This closes the exception #466 deliberately left open one PR ago: after that
change this was the **last silent survivor-pick among the four interpolators**,
and #466's own docs named it as deferred to this issue. `Curve::new` and the
`BiLinearInterpolation` trait now say all four behave alike.

A stack *elsewhere* on the curve still does not stop the method answering,
matching the precedent `linear_interpolate` set in #466: only the near edge's
width is divided by, and a stack on the far edge is read for its two ordinates
alone.

## Reference

Built in the scratchpad, never in the repo: the rule restated from scratch and
evaluated in **exact rational arithmetic**, cross-checked against the same rule
in `mpmath` at 60 dps. `C4 = (0,0),(1,1),(2,4),(3,9)`; `C5` adds `(4,16)`;
`C6 = (0,1),(0.5,2.25),(2,−1.5),(3.25,4),(5,0.75)` — non-uniform spacing and a
non-monotone ordinate.

| curve | x | where | reference (exact) | computed | error |
|---|---|---|---|---|---|
| C4 | 0 | left end (sample) | 0 | 0 | exact |
| C4 | 0.5 | interior seg 0 | 3.5 | 3.5 | exact |
| C4 | 1.5 | interior seg 1 — *was an error* | 4.5 | 4.5 | exact |
| C4 | 2.5 | last segment — *was an error* | 6.5 | 6.5 | exact |
| C4 | 3 | right end (sample) | 9 | 9 | exact |
| C5 | 0.5 | interior seg 0 | 3.5 | 3.5 | exact |
| C5 | 1.5 | interior seg 1 | 7.5 | 7.5 | exact |
| C5 | 2.5 | interior seg 2 — *was an error* | 9.5 | 9.5 | exact |
| C5 | 3.5 | last segment — *was an error* | 12.5 | 12.5 | exact |
| C5 | 4 | right end (sample) | 16 | 16 | exact |
| C6 | 0.25 | interior seg 0 | 1.4375 | 1.43750 | exact |
| C6 | 1 | interior seg 1 | 1.958333…33 | 1.958333…35 | 1.7e-28 |
| C6 | 2.5 | interior seg 2 — *was an error* | 1.7 | 1.700 | exact |
| C6 | 4 | last segment — *was an error* | 2.607142857…14 | 2.607142857…70 | −1.4e-28 |
| C6 | 5 | right end (sample) | 0.75 | 0.75 | exact |

The two residuals are `47/24` and `73/28`: `d_div` rounds the fraction at scale
28 with banker's rounding and the result carries it forward, so the error is one
unit in the 28th place scaled by the edge rises. Everything with a terminating
expansion is exact.

## No existing interpolation test changed

None did, and none exercised the exact-match-on-a-stack path. The deletions
across the whole diff are confined to the body of `bilinear_interpolate` and to
doc comments; not one test line was removed or edited.

Three pre-existing tests do build a curve with stacked abscissae —
`(0,0), (1,1), (0,1), (1,2)` sorts to two ordinates at `x = 0` and two at
`x = 1` — but every query in them is outside the curve's x-range, so the
exact-match branch (old `find` and new `exact_point_at` alike) finds nothing and
the query fails at `find_bracket_points` with `OutOfRange`, the same variant
before and after.

The path that actually changed — a query *at* `x = 0` or `x = 1` on such a
curve — had no test at all. It has one now:
`test_bilinear_interpolate_stacked_abscissa_is_degenerate`.

No production code selects `InterpolationType::Bilinear`; every call site
outside the trait dispatch is a test.

## Tests

3999 lib + 63 property + 423 integration + 207 doctests, 0 failed. `cargo fmt
--all --check`, `cargo clippy --all-targets --all-features --workspace -- -D
warnings`, `make scan-banned` and `make public-api-check` all clean, verified in
the worktree rather than left to CI.

## Stack

Chain B, second of three.

- Stack: #466 → **#451** (this one) → #453. Independent of chain A (#470 → #469
  → #463 → #471), chain C (#462), #459, and the #442 capstone.
- Base branch: `main`. #466 merged as #473, so this branch has been rebased
  onto `main` and the diff is now this PR's own work alone — one commit, no
  cumulative parent diff to read past. The 0.21.0 bump this chain needed came
  in with #473 and is already in `main`.
- Re-verified after the rebase rather than assuming it: 4021 lib + 67 property
  + 423 integration + 207 doctests, 0 failed, on top of the merged #470, #462,
  #466 and #459.

Closes #451

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **`Curve::bilinear_interpolate` now answers across the whole interior
+  domain, and its exact-match branch fails on a repeated abscissa** (#451).
+  The method read a four-sample window starting at the bracket index, which
+  `find_bracket_points` only guarantees to `i + 1`, so every `x` in the last
+  two segments left the window past the end and returned
+  `InterpolationError::Bilinear`. The cell is now built from the segment
+  bracketing `x` and the segment two positions on, the far one clamped to the
+  curve's last segment where it would run off the end. On the final segment
+  the two edges coincide and the answer is the linear interpolant there; on
+  the second-to-last the far edge is the immediately following segment.
+
+  The clamp reaches only those last two segments, so every abscissa the
+  method already answered keeps its answer digit for digit, and no existing
+  test changed. Values were validated against a reference implementation of
+  the rule in exact rational arithmetic, cross-checked against `mpmath` at 60
+  digits, over three curves including a non-uniform one.
+
+  The near edge is deliberately not clamped, unlike `cubic_interpolate` at its
+  own boundary: the window start is the denominator of the fraction along the
+  cell, so moving it back off the segment holding `x` pushes that fraction
+  past `1`, turns two of the four weights negative and stops the answer being
+  a convex combination of the cell's corners. On `(0,0), (1,1), (2,4), (3,9)`
+  that variant returns `9.5` at `x = 2.5`, above every ordinate on the curve.
+
+  The exact-match branch now goes through `Curve::exact_point_at` like the
+  three other interpolators, so several ordinates at the queried abscissa
+  return `InterpolationError::DegenerateInterval` instead of the lowest of
+  them. This is a behaviour change on a curve that breaks the
+  one-point-per-abscissa rule, and it closes the exception the #466 entry
+  below left open. No signature changed and the public API is unchanged.
+
 - **The crate moves to 0.21.0, which is the breaking bump for a `0.x`
   version.** `cargo-semver-checks` compares against the published 0.20.0 and
   rejected the rename below under `inherent_method_missing`, since a version
@@ -82,8 +113,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   them: the value there is not defined and no choice among them is less
   arbitrary than another. A curve with one point per abscissa is unaffected,
   and `AxisOperations::get_values` still reads every ordinate at an abscissa.
-  The exact-match branch of `Curve::bilinear_interpolate` still returns the
-  lowest ordinate; it is being reworked under #451.
+  The exact-match branch of `Curve::bilinear_interpolate` was left out of that
+  pass and is covered by the #451 entry above, which brings it into line.
 
   The one-point-per-abscissa rule, and what each consumer does when it is
   broken, are now stated on `Curve::new`, `Curve::from_vector`,
--- a/src/curves/curve.rs
+++ b/src/curves/curve.rs
@@ -146,14 +146,10 @@ impl Curve {
     /// `new` stores them all; a curve built that way is outside the contract
     /// and the consumers behave as follows:
     ///
-    /// - the interpolators return
+    /// - all four interpolators return
     ///   [`InterpolationError::DegenerateInterval`] when asked about a
     ///   repeated abscissa, and pick no survivor: the bracket around it has
-    ///   zero width and the slope across it is undefined. The one exception
-    ///   is the exact-match branch of
-    ///   [`BiLinearInterpolation::bilinear_interpolate`],
-    ///   which still returns the lowest ordinate and is being reworked under
-    ///   issue #451;
+    ///   zero width and the slope across it is undefined;
     /// - `get_point` and `merge_axis_interpolate` read the first match in
     ///   `(x, y)` order, which is the lowest ordinate of the stack, and the
     ///   rest are invisible to the caller. Use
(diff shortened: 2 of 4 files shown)


joaquinbejar (commented):
The upper boundary rule keeps the bracketing segment local, the checked arithmetic is preserved, and the reference and boundary coverage exercise the changed behavior. APPROVE. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

joaquinbejar:
Thanks, merging. Agreed on the clamp being the wrong shape for bilinear: the window start is also the denominator of t, so clamping it pushes t past 1 and the weights stop being a convex combination of the cell's corners.

PR #478, opened 2026-08-30
Refuse the alpha sentinel in an aggregate instead of aborting or dropping a leg (#469)

### LINKED ISSUE #469: Greek aggregation aborts on its own alpha sentinel

joaquinbejar:
## Summary

The greek aggregators summed eleven greeks across a strategy's legs with
unchecked `Decimal` addition. The plain overflow was the lesser problem:
`alpha_from` returns `Decimal::MAX` as a **documented sentinel** when theta
vanishes but gamma does not, so a strategy with two legs at a sub-contract
quantity aborted on `Decimal::MAX + Decimal::MAX`.

Fixing that surfaced a second defect in the same place, and it is the one worth
reading this PR for.

## 24 accumulations, not 23

12 in `Greeks::greeks` (the twelve-field aggregate) and 12 in the per-greek
aggregators — the issue says eleven of the latter; there are twelve. All go
through `d_add` now. No new plumbing: `GreeksError: From<DecimalError>` already
existed and both entry points already returned `Result`.

## The second defect: `checked_add` cannot see this one

**One sentinel leg beside an ordinary one never aborted — it silently dropped
the ordinary leg.** `Decimal::MAX` has scale 0 with a mantissa filling all 96
bits, so a fractional addend would need a 29th significant digit; `rust_decimal`
rescales the addend to scale 0 instead of failing, and anything below half a
unit rounds to zero. The addition *succeeds*, the running total stands still,
and the other leg's alpha is gone without a trace.

Converting to `d_add` does not fix that, because nothing overflows. So the two
alpha accumulation paths — and **only** those two, no other greek has a
sentinel — carry an explicit guard that refuses a sentinel operand and names the
leg:

```
leg {index} ({symbol} {strike} {style} {side}) has a vanished theta, so its
alpha is the Decimal::MAX sentinel and cannot be summed into an aggregate;
read the leg's own alpha instead
```

### The guard refuses exactly when a contribution would be lost

This is narrower than "a sentinel can never be aggregated", deliberately:

- **Refused** — a sentinel meeting a total that already carries a contribution,
  or a contribution meeting a total that is already the sentinel. Covers
  sentinel + ordinary in either order, and two sentinels, caught before the
  range limit.
- **Allowed** — a lone sentinel, and a sentinel beside legs whose own alpha is
  zero (`alpha` returns zero for a vanished gamma, so nothing is dropped).

The looser rule would have broken `OptionData::greeks_snapshot`
(`src/chains/optiondata.rs:1229`), which calls `greeks()` on a **single** option
and maps `alpha == Decimal::MAX` to `None`. Refusing a lone sentinel there
would discard eleven good greeks to report one bad one, and falsify that
module's doc comment. **A lone sentinel still returns the sentinel; chain
snapshots behave exactly as before.**

No silent drop survives either way.

## The mechanism is now pinned once, where the next person will look

`Decimal::MAX.checked_add(x)` returning `Ok(Decimal::MAX)` is a property of
`rust_decimal`, not of this sentinel, and it has cost **three** separate
diagnoses in this programme — it also made `chains::optiondata::unlock` re-emit
the locked quote it existed to remove (#459, merged as #472).

`checked_add_near_max_reports_ok_without_moving_the_value` now sits beside the
`d_add` family in `src/model/decimal.rs`, names both call sites it has broken,
and states the rule: **when you nudge a value and need to know it moved, compare
the result against the input; `checked_*` returning `Ok` only says the result
was representable.**

It also pins the boundary, which is narrower than first diagnosed: the silent
window is `|x| < 0.5`, not `|x| < 1`. At half a unit the addend rounds away from
zero to 1 and the sum genuinely overflows. Both sides are asserted so the
boundary is visible rather than implied.

## The sentinel stays `Decimal::MAX`

The issue asks whether it is still the right sentinel. It is: it was a hazard
only because the aggregators added it with the raw operator, and reporting that
sum was the whole of the danger. Replacing it would change the documented return
of the public `alpha` and ripple into `OptionData::calculate_greeks` for no
added safety. `alpha_from`'s doc comment carries that reasoning, states that
nothing prevents a caller reaching the sentinel — with the concrete reachable
case, rather than a claim of impossibility — and points at the pinned test.

The error variant reuses `CalculationErrorKind::ThetaError`: a vanished theta is
the cause, `numerical.rs:147` already uses it that way, and inventing an
`AlphaError` would be a public-API change to buy a better label.

## Aggregated greeks are unchanged — two independent checks

1. **Source-level.** `Decimal::checked_add` and `impl Add for Decimal` both
   dispatch to the same `ops::add_impl`; they differ only in whether a non-`Ok`
   result panics or yields `None`. For any sum the old code did not abort on,
   `d_add` returns the identical value — bit-identically, not approximately.
2. **Differential test.** `test_checked_aggregation_matches_the_unchecked_sum`
   builds a 24-leg matrix (both styles × both sides × three strikes × two
   maturity/vol/quantity regimes), accumulates a reference with the raw `+`
   operator greek by greek, and asserts the whole `Greek` struct equals it
   across all twelve fields, plus each of the twelve `Greeks::*` methods against
   the corresponding reference field.

The fixture for the sentinel tests was verified to be the real defect rather
than a construction: reverting only the two alpha accumulations to `+=` makes
both two-sentinel tests fail with `Addition overflowed` while the other five
still pass.

## The proptest exclusion is gone

`Greeks::greeks` was deliberately absent from the strategies panic-freedom
property, with a comment explaining why. Both are removed and the call is
driven with the rest. 67 property tests pass.

## Tests

4019 lib + 67 property + 423 integration + 207 doctests, 0 failed. `cargo fmt
--all --check`, `cargo clippy --all-targets --all-features --workspace -- -D
warnings`, `make scan-banned` and `make public-api-check` clean — re-run after
the rebase onto `main` rather than trusting the pre-rebase run.

Public API unchanged: both entry points already returned `Result<_,
GreeksError>` and no error variant was added. `CHANGELOG.md` carries the
behaviour change under `### Fixed`, including the note that a lone sentinel is
unaffected.

## Stack

Chain A, second of four.

- Stack: #470 → **#469** (this one) → #463 → #471. Independent of chain B
- Base branch: `main`. #470 merged as #475, so this branch is rebased onto
  `main` and the diff is its own work alone.

Closes #469

### DIFF
--- a/src/greeks/equations.rs
+++ b/src/greeks/equations.rs
@@ -232,7 +232,7 @@ pub trait Greeks {
         let mut vega = Decimal::ZERO;
         let mut rho = Decimal::ZERO;
         let mut rho_d = Decimal::ZERO;
-        let mut alpha = Decimal::ZERO;
+        let mut alpha = AlphaSum::default();
         let mut vanna = Decimal::ZERO;
         let mut vomma = Decimal::ZERO;
         let mut veta = Decimal::ZERO;
@@ -246,13 +246,7 @@ pub trait Greeks {
             vega = d_add(vega, single.vega, "greeks::aggregate::vega")?;
             rho = d_add(rho, single.rho, "greeks::aggregate::rho")?;
             rho_d = d_add(rho_d, single.rho_d, "greeks::aggregate::rho_d")?;
-            alpha = add_alpha(
-                alpha,
-                single.alpha,
-                option,
-                index,
-                "greeks::aggregate::alpha",
-            )?;
+            alpha = alpha.push(single.alpha, option, index, "greeks::aggregate::alpha")?;
             vanna = d_add(vanna, single.vanna, "greeks::aggregate::vanna")?;
             vomma = d_add(vomma, single.vomma, "greeks::aggregate::vomma")?;
             veta = d_add(veta, single.veta, "greeks::aggregate::veta")?;
@@ -266,7 +260,7 @@ pub trait Greeks {
             vega,
             rho,
             rho_d,
-            alpha,
+            alpha: alpha.total,
             vanna,
             vomma,
             veta,
@@ -417,17 +411,12 @@ pub trait Greeks {
     /// the sentinel and another leg contributes to the same sum.
     fn alpha(&self) -> Result<Decimal, GreeksError> {
         let options = self.get_options()?;
-        let mut alpha_value = Decimal::ZERO;
+        let mut alpha_value = AlphaSum::default();
         for (index, option) in options.into_iter().enumerate() {
-            alpha_value = add_alpha(
-                alpha_value,
-                alpha(option)?,
-                option,
-                index,
-                "greeks::alpha::aggregate",
-            )?;
+            alpha_value =
+                alpha_value.push(alpha(option)?, option, index, "greeks::alpha::aggregate")?;
         }
-        Ok(alpha_value)
+        Ok(alpha_value.total)
     }
 
     /// Calculates the aggregate vanna value for all options.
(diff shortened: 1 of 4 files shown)


joaquinbejar (commented):
The checked Greek aggregation is the right direction, but the alpha sentinel state currently reports the wrong leg in one ordering and the new public documentation links to a removed helper, which makes cargo doc fail. REQUEST_CHANGES. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

src/greeks/equations.rs
```diff
@@ -1731,6 +1806,123 @@ fn alpha_from(gamma: Decimal, theta: Decimal) -> Result<Decimal, GreeksError> {
...
+            self.total = Decimal::MAX;
+            return Ok(self);
+        }
+        // A leg whose own alpha is zero adds nothing and loses nothing, so it
+        // neither counts as a contribution nor conflicts with the sentinel.
+        if value.is_zero() {
+            return Ok(self);
+        }
+        if self.sentinel {
+            return Err(alpha_sentinel_error(option, index));
```
joaquinbejar: 🟠 **Warning**: When the sentinel arrives before an ordinary leg, this branch passes the ordinary option to `alpha_sentinel_error`, so the returned `ThetaError` falsely says that leg’s theta vanished. Store the sentinel leg’s identity in `AlphaSum` and report it here, or use an order-neutral error message, then assert the reason for the sentinel-first ordering.

src/greeks/equations.rs
```diff
@@ -211,7 +211,15 @@ pub trait Greeks {
...
     ///
-    /// Returns a `GreeksError` if any individual Greek calculation fails.
+    /// Returns a `GreeksError` if any individual Greek calculation fails, or
+    /// [`crate::error::DecimalError::Overflow`] (wrapped by `GreeksError`) when
+    /// one of the twelve running sums leaves the representable `Decimal` range.
+    ///
+    /// The `alpha` sum has one further failure of its own, and it is reachable
+    /// on ordinary input. [`alpha`] returns `Decimal::MAX` for a leg whose theta
+    /// has vanished, and that value cannot be added to a real one; a leg
+    /// carrying it beside any other contribution is refused by [`add_alpha`],
```
joaquinbejar: 🟠 **Warning**: This link and the same link in `Greeks::alpha` still point to `add_alpha`, which no longer exists after the helper became `AlphaSum::push`. The crate denies broken intra-doc links, so `cargo doc` fails on both lines; use plain code text or point the documentation at the current helper.

src/greeks/equations.rs
joaquinbejar: Good catch, that one was mine and it was reporting a vanished theta for a leg whose theta is fine. AlphaSum keeps the sentinel leg's index and description from the moment it arrives and reports that instead of the arriving leg, so the same two legs give the same attribution in either order, asserted for both orderings.

src/greeks/equations.rs
joaquinbejar: Also mine, and cargo doc did fail on all three. They are plain code text now rather than links, since AlphaSum is private and a link would trade a broken-link error for a private-link warning. cargo doc is clean apart from two private-link warnings that predate this branch.

joaquinbejar:
Both fixed in 811807f1, and both were introduced by my own order-invariance commit rather than by the original branch. The refusal now names the sentinel leg in either order, and the three stale doc links are gone so cargo doc passes.

PR #479, opened 2026-08-30
Give the chain file tests a directory of their own

joaquinbejar:
## Summary

Second sighting of the shape #476 fixed, in different clothes. Two chain tests failed together on a full-suite run and each passed on its own:

```
chains::chain::tests_chain_base::test_load_from_json
chains::chain::chain_coverage_tests_bis::test_deserializer_field_handling
```

Both end with `fs::remove_file(...)` asserted with `is_ok()` / `.unwrap()`.

## What is actually being asserted

`cargo` runs test binaries concurrently against a single working directory. A test that asserts its own cleanup succeeded is asserting that **nothing else touched the file between the write and the removal**, and that is not a property of the test — it is a property of everything else running at the same time. It is the same argument #476 made for `prepare_file_path`: `remove_file` reporting `NotFound` means the postcondition already holds.

These two do not collide on a shared path — the four file-writing tests in that module use four distinct names — so the fix is not deduplication. The fix is that none of them should be writing into the working tree at all.

## What changed

Six tests now write into a directory of their own under the system temp directory, and assert the round trip rather than the deletion:

| test | was | asserts now |
|---|---|---|
| `test_save_to_csv` | `./SP500-…csv` | the artifact exists |
| `test_save_to_json` | `./SP500-…json` | the artifact exists |
| `test_load_from_csv` | `./SP500-…csv` | it loads back with its fields |
| `test_load_from_json` | **`tests/`**`/SP500-…json` | it loads back with its fields |
| `test_deserializer_field_handling` | `./AAPL-…json` | it deserializes to the same chain |

The deletion was never the behaviour under test; `save_to_csv` / `save_to_json` / `load_from_*` are.

`test_load_from_json` was writing into `tests/`, which is a source directory — a failed run left an artifact next to the integration suite.

## Testing

- [x] 4010 lib tests, 0 failures; `cargo fmt --all --check` and clippy `-D warnings` clean
- [x] `tests/` no longer collects a `.json` artifact after a run

## Public API impact

None. Test-only.

### DIFF
--- a/src/chains/chain.rs
+++ b/src/chains/chain.rs
@@ -5333,15 +5333,31 @@ impl From<&Vec<OptionData>> for OptionChain {
 
 #[cfg(test)]
 mod tests_chain_base {
+
     #![allow(clippy::indexing_slicing)]
     use super::*;
+
+    /// A directory of this test's own, outside the working tree.
+    ///
+    /// Writing chain artifacts into `.` or `tests/` puts every file-writing
+    /// test in one shared directory, and `cargo` runs test binaries
+    /// concurrently. Asserting that the cleanup succeeded then asserts that
+    /// nothing else touched the file, which is not a property of the test:
+    /// two of these failed together on a full-suite run, each passing on its
+    /// own. `tests/` is a source directory besides, so the round trip was
+    /// leaving artifacts in the tree.
+    fn scratch_dir(name: &str) -> String {
+        let dir = std::env::temp_dir().join(format!("optionstratlib_chain_{name}"));
+        let _ = std::fs::remove_dir_all(&dir);
+        std::fs::create_dir_all(&dir).expect("the temp directory is writable");
+        dir.to_string_lossy().into_owned()
+    }
 
     use crate::model::ExpirationDate;
 
     use rust_decimal_macros::dec;
 
     use positive::spos;
-    use std::fs;
     use tracing::info;
 
     #[test]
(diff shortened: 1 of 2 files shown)


src/chains/chain.rs
```diff
@@ -5333,15 +5333,31 @@ impl From<&Vec<OptionData>> for OptionChain {
...
+    ///
+    /// Writing chain artifacts into `.` or `tests/` puts every file-writing
+    /// test in one shared directory, and `cargo` runs test binaries
+    /// concurrently. Asserting that the cleanup succeeded then asserts that
+    /// nothing else touched the file, which is not a property of the test:
+    /// two of these failed together on a full-suite run, each passing on its
+    /// own. `tests/` is a source directory besides, so the round trip was
+    /// leaving artifacts in the tree.
+    fn scratch_dir(name: &str) -> String {
+        let dir = std::env::temp_dir().join(format!("optionstratlib_chain_{name}"));
```
joaquinbejar: 🟠 **Warning**: This path is deterministic across the whole machine, so two checkout or test processes running the same test share it and one can remove the other’s directory on the next line. Return a unique `tempfile::TempDir` and keep it alive for the test, applying the same fix to the duplicate helper near the end of this file.

joaquinbejar (commented):
The tests are moving out of the working tree, but their new temp directories are deterministic across processes, so concurrent runs can still delete each other’s files and preserve the flake this change is meant to remove. REQUEST_CHANGES. GitHub cannot record that formal state from the PR author, so this review is submitted as COMMENT.

src/chains/chain.rs
joaquinbejar: Right, and that was the same flake moved rather than fixed: a path built from the test's own name is shared by every process running that test, so the remove_dir_all on the next line is the collision again. Both helpers use tempfile::tempdir now, which picks a fresh directory per call and removes it when the handle drops, and each test holds the handle for as long as it needs the files. tempfile was already a dev-dependency.

joaquinbejar:
Fixed in d2dea1a4. You were right that the fix reproduced its own bug one level down: unique per test is not unique per process, and the directory is picked by tempfile now.

PR #480, opened 2026-08-30
Read the multi-leg break-even points without panicking (#463)

### LINKED ISSUE #463: Panic sweep: break-even indexing in the multi-leg strategies

joaquinbejar:
## Summary

The mechanical half of the strategies sweep: **unguarded `break_even_points`
indexing across the fourteen multi-leg strategies**, in `get_profit_area`,
`get_profit_ratio`, `get_profit_ranges`, `get_loss_ranges` and
`get_best_range_to_show`. An empty vector turned each into `index out of
bounds: the len is 0 but the index is 0`.

A probe over these files' public API, run against the parent commit and against
this branch:

| | panics | calls |
|---|---|---|
| before | **43,857** | 254,070 |
| after | **0** | 244,962 |

Breakdown of the 43,857: 36,234 index-out-of-bounds (31,806 `len 0 index 0`,
4,014 `len 0 index 1`, 414 `len 1 index 1`), 7,281 `Positive` invariant or
overflow, 342 division by zero. The call count drops because `get_strategy` now
rejects a zero-quantity leg set instead of building an unusable strategy.

## 75 accesses, not 65

The issue counted `break_even_points[i]`. A second form,
`self.get_break_even_points()?[0]`, sits in the same functions in five of the
files and the issue's grep missed it: **10 more**.

| file | converted | file | converted |
|---|---|---|---|
| `long_strangle.rs` | 10 | `iron_condor.rs` | 4 |
| `short_strangle.rs` | 10 | `iron_butterfly.rs` | 4 |
| `long_straddle.rs` | 8 | `call_butterfly.rs` | 4 |
| `short_straddle.rs` | 8 | `short_butterfly_spread.rs` | 4 |
| `long_butterfly_spread.rs` | 7 | `bear_call_spread.rs` | 3 |
| `bull_put_spread.rs` | 5 | `bear_put_spread.rs` | 3 |
| | | `bull_call_spread.rs` | 3 |
| | | `poor_mans_covered_call.rs` | 2 |

`base.rs` and `custom.rs` keep theirs — they are #460's and already
length-guarded.

## Per strategy: constructor bug, or legitimate state?

The issue asks for this decision per strategy, and the two answers are
different fixes. Conflating them is how a silent wrong value gets in.

| strategy | empty vector is | why |
|---|---|---|
| `LongButterflySpread` | **constructor bug** (+ legitimate short vector) | `get_strategy` never called `update_break_even_points`; separately a wing that never crosses zero yields 1 point or none |
| `ShortButterflySpread` | **constructor bug** (+ legitimate short vector) | same missing call, same conditional push |
| `CallButterfly` | legitimate state | its push is conditional on `Positive::new_decimal(candidate)`; the probe hit `len 1 index 1` from a plain `new` |
| `BearCallSpread`, `BearPutSpread`, `BullCallSpread`, `BullPutSpread`, `PoorMansCoveredCall` | legitimate state | the constructor always pushes exactly 1 |
| `LongStraddle`, `ShortStraddle`, `LongStrangle`, `ShortStrangle`, `IronCondor`, `IronButterfly` | legitimate state | the constructor always pushes exactly 2 |

Where it is a constructor bug it is fixed at the constructor (`1c835106`);
everywhere else the readers report the shortfall and name which point is
missing — `StrategyError` on the `Strategies` methods,
`ProbabilityError::RangeError` on the probability ranges.

**Two public routes reach an empty vector on every strategy in the table**, so
"legitimate state" is not hypothetical: `break_even_points` is a `pub` field
and every strategy derives `Deserialize`, so `"break_even_points": []` in a
JSON document reached all five readers.

## Unchanged behaviour, and the five places it is not

For every vector long enough to answer the read, **every value is identical in
all fourteen files**. The substitutions are value-preserving by construction:
`price_gap(hi, lo)` is the same subtraction for an ordered pair; `d_div` rounds
banker's at scale 28 and `positive`'s `DIV_SCALE` is also 28 with the same
strategy, so quotients are bit-identical; `d_add` / `d_sub` / `d_mul` and the
`checked_*` twins differ from the operators only in the overflow arm.

Five sites change what happens where the old code **panicked**, never where it
returned: `BearCallSpread::get_max_loss` on inverted strikes, now
`MaxLossError`; `PoorMansCoveredCall::get_profit_area` for a profit above the
strike; the two straddle/strangle ratios on a zero divisor; and
`CallButterfly::get_profit_ratio` on an overflowing product.

**One behaviour change beyond that**, called out because it is a constructor
that used to succeed: `LongButterflySpread::get_strategy` and
`ShortButterflySpread::get_strategy` now return `Err` for a zero-quantity leg
set, where they returned `Ok` with an empty vector.

It reads as a new refusal and it is not — it is the two outliers rejoining the
convention. Counting call sites of `strategy.update_break_even_points()?` on
`889004b7`, across the fourteen multi-leg strategies:

| call sites | files |
|---|---|
| 2 — from `new` **and** from `get_strategy` | the other twelve |
| 1 — from `new` only | `long_butterfly_spread.rs`, `short_butterfly_spread.rs` |

So twelve constructors already propagate that failure rather than hand back a
strategy whose break-even vector they could not fill. **The butterflies' `Ok`
was the absence of the call, not a decision to tolerate a degenerate leg set.**
Restoring it restores the convention; the `Err` is a consequence of that, not a
new policy.

The substance agrees: `update_break_even_points` divides the net cost by the
leg quantity, and a leg set with no contracts has no per-contract break-even to
compute. Returning `Ok` moved that failure to whichever reader touched it
first, which is this sweep's own thesis inverted — report at the boundary, not
at the abort site.

## Tests

The strategies panic-freedom property now covers the multi-leg strategies
(+228 lines). 4032 lib + 71 property + 423 integration + 207 doctests, 0
failed.

Gates, all run on the rebased head:

```
```

The last three are in `rules/global_rules.md`'s pre-submission checklist and
are not covered by `make pre-push`: `create-doc` has no `--all-features`, and
`--all-features` is not the default feature set.

## Stack

Chain A, third of four.

- Stack: #470 → #469 → **#463** (this one) → #471. Independent of chain B
- Base branch: `main`. #470 and #469 are merged, so this branch is rebased onto
  `main` and the diff is its own work alone.

Closes #463

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,43 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Every multi-leg strategy aborted the process when its break-even vector
+  was shorter than the point it read** (#463). `get_profit_area`,
+  `get_profit_ratio`, `get_profit_ranges`, `get_loss_ranges` and
+  `get_best_range_to_show` indexed `break_even_points` directly across the
+  fourteen multi-leg strategies, 75 unguarded accesses, and an empty vector
+  turned each into `index out of bounds: the len is 0 but the index is 0`.
+  The vector is a `pub` field and every strategy derives `Deserialize`, so a
+  JSON document carrying `"break_even_points": []` reached all five. Two ways
+  in needed no extreme input at all: `LongButterflySpread::get_strategy` and
+  `ShortButterflySpread::get_strategy` never called
+  `update_break_even_points`, so a butterfly assembled from a leg set always
+  carried an empty vector; and a butterfly whose wing never crosses zero
+  profit legitimately has one break-even point or none, which made
+  `the len is 1 but the index is 1` reachable from an ordinary constructor.
+
+  The two constructors now populate the vector like every other constructor
+  in the crate. The readers report the shortfall, `StrategyError` on the
+  `Strategies` methods and `ProbabilityError::RangeError` on the probability
+  ranges, and name which of the two points is missing. Every value a
+  populated vector produced is unchanged: the widths taken between two
+  break-even points go through `price_gap`, which is the same subtraction for
+  an ordered pair, and a zero-width region instead of a panic for a crossed
+  one.
+
+  The arithmetic around those reads moved to the checked helpers in the same
+  files, since a `Result` that reports a missing point and then aborts on the
+  next line is not panic-free. `update_break_even_points` divides the premium
+  by a quantity that can be zero; `get_max_loss` on the bear call spread
+  subtracted strikes in an order nothing enforces; the strangles' default
+  strikes multiply the spot by 1.1 and 0.9, which overflows at
+  `Positive::MAX`; and the fourteen `PnLCalculator` impls summed their legs
+  with `impl Add for PnL`, which returns `Self` and so has nowhere to report
+  an overflowing four-leg total. They now use the checked `PnL::try_add` that
+  #460 added for exactly this. A probe over the public API of these files
+  went from 43,857 panics in 254,070 calls to none, and the property suite
+  gained four cases covering the fourteen strategies, both butterfly leg-set
+  constructors and the emptied vector.
 - **Six chain tests wrote their artifacts into the working tree and asserted
   their own cleanup succeeded.** `cargo` runs test binaries concurrently
   against one working directory, so `assert!(fs::remove_file(..).is_ok())`
--- a/src/strategies/bear_call_spread.rs
+++ b/src/strategies/bear_call_spread.rs
@@ -32,13 +32,13 @@ use crate::{
     error::{
         GreeksError, OperationErrorKind, PricingError,
         position::{PositionError, PositionValidationErrorKind},
-        probability::ProbabilityError,
+        probability::{ProbabilityError, ProfitLossRangeErrorKind},
         strategies::{ProfitLossErrorKind, StrategyError},
     },
     greeks::Greeks,
     model::{
         ProfitLossRange,
-        decimal::d_sum,
+        decimal::{d_add, d_div, d_mul, d_sub, d_sum},
         position::Position,
         types::{OptionBasicType, OptionStyle, OptionType, Side},
         utils::mean_and_std,
@@ -340,11 +340,21 @@ impl BreakEvenable for BearCallSpread {
     fn update_break_even_points(&mut self) -> Result<(), StrategyError> {
         self.break_even_points = Vec::new();
 
-        self.break_even_points.push(
-            (self.short_call.option.strike_price
-                + self.get_net_premium_received()? / self.short_call.option.quantity)
-                .round_to(2),
-        );
+        // The credit per contract: a spread with no contracts has none, and
+        // the strike it is added to can sit at the top of the `Decimal`
+        // range. Both are reported rather than aborted.
+        let per_contract = d_div(
+            self.get_net_premium_received()?.to_dec(),
+            self.short_call.option.quantity.to_dec(),
+            "BearCallSpread::update_break_even_points",
+        )?;
+        let break_even = d_add(
+            self.short_call.option.strike_price.to_dec(),
+            per_contract,
+            "BearCallSpread::update_break_even_points",
+        )?;
+        self.break_even_points
+            .push(Positive::new_decimal(break_even)?.checked_round_to(2)?);
 
         Ok(())
     }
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,43 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  The two constructors now populate the vector like every other constructor
+  in the crate. The readers report the shortfall, `StrategyError` on the
+  `Strategies` methods and `ProbabilityError::RangeError` on the probability
+  ranges, and name which of the two points is missing. Every value a
+  populated vector produced is unchanged: the widths taken between two
+  break-even points go through `price_gap`, which is the same subtraction for
+  an ordered pair, and a zero-width region instead of a panic for a crossed
+  one.
+
+  The arithmetic around those reads moved to the checked helpers in the same
```
joaquinbejar, line 36: 💡 **Suggestion**: This change is over 1,400 added lines and combines safe break-even reads, arithmetic hardening, PnL aggregation, and constructor behavior across fourteen strategies. Consider moving the checked arithmetic and PnL work into a follow-up so the indexing fix and its behavior changes are easier to validate and backport.

joaquinbejar (commented):
The break-even access changes preserve valid strategy results, surface malformed and extreme inputs as typed errors, and pass both the new panic-freedom properties and the 4,032-test all-feature library run. I left one non-blocking suggestion about splitting the broader arithmetic and PnL hardening from this already large indexing fix. COMMENT. GitHub cannot record a formal review state from the PR author, so this review is submitted as COMMENT.

CHANGELOG.md
joaquinbejar: Fair concern, but the split is not clean here so I would rather keep it whole. The arithmetic sits inside the same impls that consume the break-even values, ProbabilityAnalysis and Profit especially, so a split would leave the first half reading points it cannot compute yet. The measurement is also whole-file: 43,857 panics to 0 over this set, which would need rerunning twice against two partial states to mean anything. Splitting also does not help backporting, since the crate keeps no release branches.

joaquinbejar:
Thanks. Keeping it whole for the reason in the thread, and one thing worth recording for whoever reads this later: the two butterfly constructors returning Ok with an empty break-even vector was the absence of an update_break_even_points call rather than a decision to tolerate a degenerate leg set. Eighteen other constructors already call it with ? before returning, so the Err this PR introduces aligns them with the existing convention rather than adding a new refusal.

PR #481, opened 2026-08-30
Fold the merge product in order instead of in parallel (#453)

### LINKED ISSUE #453: Multiply merge is non-deterministic across runs

joaquinbejar:
## Summary

`Curve::merge` and `Surface::merge` reduced their `Multiply` arm with a rayon
`reduce`. `Decimal` multiplication rounds past 28 significant digits, so it is
**not associative**, and rayon picks the grouping from its length splitter —
which is keyed to the input length *and the ambient pool size*.

The premise is confirmed and it is worse than "different last digits". Merging
40 curves, same binary, same input:

| rayon threads | distinct results in 2000 runs |
|---|---|
| 1 | 1 |
| 2 | **52** |
| 4 | **809** |
| 8 | **1996** |
| 16 | 1 |

Two machines with different core counts disagree on the same input while each
looks perfectly stable locally. Over a sweep of 205,379 reciprocal triples,
**628 (0.31%) regroup to a different last digit**.

## `Curve`'s Divide was still parallel

The issue and my own brief said Divide went sequential in #452. **That was
`Surface` only.** `Curve`'s Divide was still a parallel product of reciprocals,
taking its bracketing from the same splitter, so "merging three or more curves
is bit-reproducible" would have been false without fixing it too. Semantics are
unchanged, including the `Decimal::MAX` zero-divisor sentinel.

## Criterion

The first attempt — a 12-configuration single-shot A/B — was thrown away rather
than reported: the machine was at load 79 on 16 cores, and running the
*identical* code pair three times gave deltas from −32% to +114%. Replaced with
five alternating old/new rounds, 500 samples a side, comparing minima as the
best estimator of uncontended cost.

**End to end, `Multiply` (min over 5 paired rounds):**

| benchmark | parallel | sequential | ratio |
|---|---|---|---|
| curve size_32 / 2 operands | 93.68 µs | 89.24 µs | **0.953** |
| curve size_32 / 3 | 100.78 µs | 93.09 µs | **0.924** |
| curve size_32 / 5 | 126.91 µs | 119.98 µs | **0.945** |
| curve size_32 / 10 | 185.06 µs | 177.14 µs | **0.957** |
| curve size_512 / 10 | 692.19 µs | 608.75 µs | **0.879** |
| surface side_8 / 3 | 10.986 ms | 10.733 ms | **0.977** |

Sequential is 2–12% faster at the uncontended end and never slower; at the
median it is inside the noise floor.

**The two folds in isolation:**

| operands | rayon `reduce` | sequential fold | factor |
|---|---|---|---|
| 2 | 16.695–17.282 µs | 22.254–22.363 ns | **761×** |
| 3 | 16.282–16.801 µs | 69.709–70.576 ns | **236×** |
| 5 | 27.690–32.905 µs | 158.98–160.76 ns | **189×** |
| 10 | 26.996–29.055 µs | 702.28–888.80 ns | **35×** |

Stated plainly: that isolated figure is a *cold* dispatch from the main thread,
and in-merge the reduce is nested inside the outer grid `par_iter`, which is why
end-to-end is a few percent rather than 100×.

**Decision: sequential**, and the number that drove it is the isolated fold. The
code was dispatching a rayon job per grid point to multiply 2–10 numbers. There
was no speed to trade away for the determinism.

Reproduce: `cargo bench --bench benches -- 'decimal_product_fold'`.

## The determinism tests pin what actually broke

Repeating a merge on one pool reaches only the **work-stealing** component. The
pool size is itself an input to the grouping — rayon sets a reduction's
splitting threshold to the input length over eight times the thread count — and
the old reducer was perfectly stable on 1 and on 16 threads while flapping on
1996 of 2000 runs on 8. A test that never varied the pool would have called the
worst case stable.

So there are two guards, and their doc comments say which is which:

- `..._is_reproducible_across_pool_sizes` merges on 1, 2, 4, 8 and 16 threads.
  This is the property that actually broke.
- `..._is_reproducible` repeats 512 times on one pool, sized from the weakest
  rate that flapped (~2.6%/run) to leave about one chance in a million of
  missing a reintroduced parallel fold.

Both compare ordinates as `(mantissa, scale)` pairs: `Decimal` compares equal
across scales, so `dec!(2.0) == dec!(2.00)` and a plain equality check would not
notice a fold that moved the scale.

At the three operands the issue names, the old reducer was stable on every pool
size probed — 5000 runs, one result — but returned the `a*(b*c)` grouping. So
the tests also pin the exact product; that is what fails the old code, on the
first iteration.

## Numerical change

`Multiply` and `Divide` move in their last decimal places: old `a*(b*c)`, new
`(a*b)*c`. Any fix repins the bracketing, so a last-digit move is unavoidable in
either direction; the sequential fold at least pins the value a reader would
predict.

Measured rather than estimated — `Decimal`'s 28 is decimal *places*, not
significant digits, so the relative size depends on the value's exponent:

| arm | operands | before | after | move |
|---|---|---|---|---|
| Multiply | `0.010989010989…`, `0.010752688172…`, `0.009433962264…` | `0.0000011147302687168785768907` | `…908` | 1 unit in the last place, **9e-23** relative |
| Divide (`Curve`) | 97, 91, 93 | `0.0114616566229469455275906915` | `…888` | 27 units in the last place, **2.4e-25** relative |

`Curve`'s Divide moves further because each reciprocal is separately rounded to
28 places by `d_div` before being multiplied in. `Surface`'s Divide is untouched
and does not move.

Signed off on the grounds that a move twenty orders of magnitude below a tick is
below any price, greek or premium this crate represents, while a value that
changes with the machine's core count
cannot be reconciled, cached or regression-tested at all. Every existing merge
test compares within 1e-3/1e-4, so none moved — which protects us, not a
downstream consumer holding a golden at full precision, and the `CHANGELOG.md`
entry gives the concrete triple and both values so they can check.

## Tests and gates

4038 lib + 67 property + 423 integration + 207 doctests, 0 failed.

```
```

`d_product_iter` lands beside `d_sum_iter` in `src/model/decimal.rs` with
left-to-right order documented as its contract. `Surface::merge` also gained a
doc comment; it had none.

## Stack

Chain B, third of three — the last of that chain.

- Stack: #466 → #451 → **#453** (this one). Independent of chain A (#470 →
  #469 → #463 → #471), chain C (#462), #459, and the #442 capstone.
- Base branch: `main`. #466 and #451 are merged, so this is rebased onto `main`
  and the diff is its own work alone.

Closes #453

### DIFF
--- a/benches/geometrics/merge.rs
+++ b/benches/geometrics/merge.rs
@@ -1,0 +1,192 @@
+use criterion::{BenchmarkId, Criterion, Throughput};
+use optionstratlib::curves::{Curve, Point2D};
+use optionstratlib::geometrics::{Arithmetic, GeometricObject, MergeOperation};
+use optionstratlib::surfaces::{Point3D, Surface};
+use rayon::prelude::*;
+use rust_decimal::Decimal;
+use rust_decimal_macros::dec;
+use std::hint::black_box;
+
+/// Operand counts the merge is measured at.
+///
+/// Two operands is the `merge_with` case and cannot expose an ordering
+/// difference; three is the smallest count that can. Five and ten stand for
+/// the multi-leg strategy surfaces this crate actually multiplies together.
+const OPERAND_COUNTS: [usize; 4] = [2, 3, 5, 10];
+
+/// Point counts per curve, in the range an option chain produces: a few tens
+/// of strikes for a weekly, a few hundred for a full-term underlying.
+const CURVE_SIZES: [usize; 3] = [32, 128, 512];
+
+/// Side length of the square grid backing each surface fixture, so the
+/// surfaces carry 64 and 256 points respectively.
+///
+/// `Surface::merge` resamples onto a fixed 51 x 51 grid whatever the operands
+/// hold, so a denser fixture buys more interpolation cost per call and not one
+/// extra invocation of the fold under test. Two sizes are enough to show how
+/// far interpolation dilutes it.
+const SURFACE_SIDES: [usize; 2] = [8, 16];
+
+/// Builds a `Decimal` with a long mantissa from two small integers.
+///
+/// The quotient by 970 is a repeating decimal, so every fixture ordinate
+/// fills the 28-digit scale. That is what makes a product of three or more of
+/// them round, which is the whole subject of the measurement: on short
+/// mantissas the fold would be exact and the ordering of the reduction would
+/// not matter.
+fn long_mantissa(n: u64, offset: u64) -> Decimal {
+    let numerator = Decimal::from(n.wrapping_mul(7).wrapping_add(offset) % 89 + 11);
+    // Kept near 1 so a ten-operand product neither overflows nor collapses
+    // to zero, which would make the benchmark measure a degenerate case.
+    dec!(0.5) + numerator / dec!(970)
+}
+
+/// Builds a curve of `points` samples spanning `[0, 10]`.
+///
+/// `index` shifts the ordinates so the operands of a merge are distinct;
+/// merging a curve with itself would let the compiler share interpolation
+/// work that a real merge cannot.
+fn curve_fixture(points: usize, index: u64) -> Curve {
+    let step = dec!(10) / Decimal::from(points.max(2) - 1);
+    let samples: Vec<Point2D> = (0..points)
+        .map(|i| {
+            let x = step * Decimal::from(i);
+            Point2D::new(x, long_mantissa(i as u64, index * 31))
+        })
+        .collect();
+    Curve::from_vector(samples)
+}
+
+/// Builds a `side * side` surface spanning `[0, 10] x [0, 10]`.
+fn surface_fixture(side: usize, index: u64) -> Surface {
+    let step = dec!(10) / Decimal::from(side.max(2) - 1);
+    let samples: Vec<Point3D> = (0..side)
+        .flat_map(|i| {
+            (0..side).map(move |j| {
+                let x = step * Decimal::from(i);
+                let y = step * Decimal::from(j);
+                Point3D::new(x, y, long_mantissa((i * side + j) as u64, index * 31))
+            })
+        })
+        .collect();
+    Surface::from_vector(samples)
+}
+
+/// End-to-end cost of `Curve::merge` under `Multiply`, per operand count and
+/// curve size.
+///
+/// This is the number a caller sees. It includes the cubic interpolation of
+/// every operand onto the shared grid, which dominates; the fold under test
+/// is a small share of it, so the isolated fold is measured separately by
+/// [`benchmark_decimal_product_fold`].
+pub fn benchmark_curve_merge_multiply(c: &mut Criterion) {
+    let mut group = c.benchmark_group("curve_merge_multiply");
+
+    for &size in CURVE_SIZES.iter() {
+        for &operands in OPERAND_COUNTS.iter() {
+            let curves: Vec<Curve> = (0..operands)
+                .map(|i| curve_fixture(size, i as u64))
+                .collect();
+            let refs: Vec<&Curve> = curves.iter().collect();
+
+            group.throughput(Throughput::Elements(operands as u64));
+            group.bench_with_input(
+                BenchmarkId::new(format!("size_{size}"), operands),
+                &refs,
+                |b, refs| {
+                    b.iter(|| {
+                        let merged = Curve::merge(black_box(refs), MergeOperation::Multiply);
+                        black_box(merged.is_ok())
+                    })
+                },
+            );
+        }
+    }
+
+    group.finish();
+}
+
+/// End-to-end cost of `Surface::merge` under `Multiply`.
+///
+/// `Surface::merge` samples a 51 x 51 grid, so it runs the fold 2601 times
+/// per call against `Curve::merge`'s 101: if the reduction strategy were
+/// ever going to show up end to end, it would show up here first.
+pub fn benchmark_surface_merge_multiply(c: &mut Criterion) {
+    let mut group = c.benchmark_group("surface_merge_multiply");
+    group.sample_size(10);
+
+    for &side in SURFACE_SIDES.iter() {
+        for &operands in OPERAND_COUNTS.iter() {
+            let surfaces: Vec<Surface> = (0..operands)
+                .map(|i| surface_fixture(side, i as u64))
+                .collect();
+            let refs: Vec<&Surface> = surfaces.iter().collect();
+
+            group.throughput(Throughput::Elements(operands as u64));
+            group.bench_with_input(
+                BenchmarkId::new(format!("side_{side}"), operands),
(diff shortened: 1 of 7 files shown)


src/model/decimal.rs
```diff
@@ -396,6 +396,45 @@ pub(crate) fn d_mul(lhs: Decimal, rhs: Decimal, op: &'static str) -> Result<Deci
...
+/// Multiplicative counterpart to [`d_sum_iter`], with the same
+/// overflow-tagging semantics and the same zero-allocation shape. Returns
+/// [`Decimal::ONE`] on an empty iterator, so an empty product is the
+/// multiplicative identity just as an empty sum is [`Decimal::ZERO`].
+///
+/// # Ordering
+///
+/// The fold runs strictly left to right, and that order is part of the
+/// contract rather than an implementation detail. `Decimal` multiplication
+/// rounds as soon as a product needs more than the 28 significant digits the
```
joaquinbejar: 💡 **Suggestion**: This should say 28 decimal places, not 28 significant digits. The mantissa holds about 28.9 digits while `MAX_SCALE` is 28. These examples round because their scale exceeds that limit even though the result has fewer than 28 significant digits. The same wording appears in the merge docs and determinism test comments.

benches/geometrics/merge.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/// End-to-end cost of `Surface::merge` under `Multiply`.
+///
+/// `Surface::merge` samples a 51 x 51 grid, so it runs the fold 2601 times
+/// per call against `Curve::merge`'s 101: if the reduction strategy were
+/// ever going to show up end to end, it would show up here first.
+pub fn benchmark_surface_merge_multiply(c: &mut Criterion) {
+    let mut group = c.benchmark_group("surface_merge_multiply");
+    group.sample_size(10);
```
joaquinbejar, line 116: 💡 **Suggestion**: The reported 500 samples per side does not apply to this group. A sample size of 10 produces only 50 Criterion samples across the five runs. Please distinguish the surface count in the PR and changelog so this result is reproducible.

benches/geometrics/merge.rs
```diff
@@ -0,0 +1,192 @@
+use criterion::{BenchmarkId, Criterion, Throughput};
```
joaquinbejar, line 1: 💡 **Suggestion**: This PR adds 669 lines. Consider moving the benchmark harness into a follow up PR so the ordered fold and regression tests stay easier to review and backport.

joaquinbejar (commented):
The ordered folds remove the pool dependent bracketing, preserve checked error handling, and the focused determinism tests plus all 4,037 all feature library tests pass. I left three nonblocking suggestions about decimal precision wording, the surface sample count, and splitting the 669 line change. COMMENT.

src/model/decimal.rs
joaquinbejar: You are right and the check is one line: Decimal::MAX is 79228162514264337593543950335, which is 29 significant digits at scale 0 and perfectly representable, so significant digits cannot be the limit these products hit. Reworded to decimal places across seven sites, the fold docs, both merge docs and the determinism test comments.

benches/geometrics/merge.rs
joaquinbejar: Correct, the 500 was the curve group only. The surface group sets sample_size(10) to keep the 51 by 51 grid affordable, so it pools 50 over the same five runs. The changelog now names both rather than one figure covering two groups.

benches/geometrics/merge.rs
joaquinbejar: Keeping the harness here. The benchmark is the evidence for the claim the PR makes about removing the parallel reducer, so splitting it would land the change and its justification separately, and whichever went first would be unreviewable on its own terms. Same reasoning I gave on the split suggestion in #480.

joaquinbejar:
Two of the three applied in 9d6dd29c, the harness stays. The digits versus decimal places one is worth carrying beyond this PR: the same wording sits in two doc comments in src/greeks/equations.rs that #478 already merged, so main currently states the wrong limit there too. Left out of this diff to keep it in the files this PR touches, and flagged for a follow-up.

src/model/decimal.rs
joaquinbejar: 💡 **Suggestion**: This correction is not complete yet. The two merge docs now say 28 decimal digits, which still reads as significant digits, the curve determinism comment still says significant digits, and the PR summary still uses the old wording. Please use decimal places or scale in all four spots.

joaquinbejar (commented):
The follow up fixes the sample count and most precision wording without touching the ordered folds. One documentation correction is still incomplete across the merge docs, the curve determinism comment, and the PR summary. COMMENT.

joaquinbejar:
Correcting my own note above: the two doc comments in src/greeks/equations.rs are fine and need no follow-up. They say a daily theta "rounds to zero at Decimal's twenty-eight-digit scale", which is a statement about the smallest representable value at scale 28 and is correct. The error Codex found was a different claim — 28 significant digits as a ceiling on a product — and that one is gone. I read the two as the same wording without checking what each was asserting, which is the mistake this PR's review was about.

PR #482, opened 2026-08-30
Give the infallible signatures that could abort an error channel (#471)

### LINKED ISSUE #471: Infallible signatures that can still abort

joaquinbejar:
## Summary

The last group from the panic-freedom programme: **public methods whose return
type has no error channel and whose bodies can abort.** After #460 closed
everything reachable through the public constructors, each of these is reachable
only by mutating a `pub` field directly, or through a trait whose signature
forbids failure. None can be fixed without a public signature change, which is
why #460 reported them instead of fixing them.

`Cargo.toml` stays at `0.21.0`: `main` has not published since that bump, so
this cycle's break is already covered. `cargo semver-checks --all-features`
reads `0.20.0 -> 0.21.0 (major change)`, `no semver update required`.

## Decisions, one per item

| item | decision | error type |
|---|---|---|
| `LegAble::pnl_at_price` | `Result<Decimal, _>` — the trait signature was the thing forbidding the report | `PricingError` (matches `Position::pnl_at_expiration`) |
| `LegAble::total_cost`, `LegAble::fees` | `Result<_, _>` — **added, see scope below** | `PositionError` (matches `Position::total_cost` / `fees`) |
| `FuturePosition::unrealized_pnl`, `PerpetualPosition::unrealized_pnl` | `Result` — **added**; they *are* the body of `pnl_at_price` on those types | `PricingError` |
| `PerpetualPosition::roe_percentage`, `margin_ratio`, `effective_leverage` | `Result` — **added**, mechanical consequence | `PricingError` |
| `PnL::try_add` | now `pub`; operator impls kept and documented as superseded | `PricingError` (unchanged) |
| `Collar::net_premium` | `Result<Decimal, _>`; fields stay `pub` | `PricingError` |
| `Collar::is_zero_cost`, `is_credit` | `Result<bool, _>` — `false` would read as a definite classification | `PricingError` |
| `CoveredCall::effective_cost_basis` | `Result<Positive, _>` | `PositiveError` |
| `ProtectivePut::effective_cost_basis` | `Result<Positive, _>` | `PositiveError` |
| `ProtectivePut::total_fees`, `protection_level` | **already fallible** — #460 fixed them; the issue's list was stale | — |
| `PortfolioGreeks::delta_gap`, `gamma_gap` | `Result<Decimal, _>` | `GreeksError` |
| `PortfolioGreeks::add`, `combined` | `Result<(), _>` / `Result<Self, _>`; `add` writes through `combined`, so a failure on the fourth of five sums leaves the receiver untouched | `GreeksError` |

Each decision is recorded in the doc comment, which is what the issue asks for —
not just the fix.

**No error enum and no variant was added**, deliberately: a variant on an
exhaustive public enum trips `enum_variant_added` and would cost a second
version bump this cycle. Every error type here already existed with the `From`
impls needed.

## Scope: six items beyond the issue's list

`LegAble` has three infallible methods with the same defect, all named in the
same property-test doc comment. Fixing one would have forced a **second break of
the same public trait** in a later PR. A trait break should happen once, so all
three go together, and the two `unrealized_pnl` implementations go with them
because they are the body `pnl_at_price` calls — leaving them infallible moves
the abort one frame down rather than removing it. The three `PerpetualPosition`
ratios follow mechanically from `unrealized_pnl`; their degenerate answers are
preserved (`ZERO` for a zero margin or notional, `MAX` for wiped equity), and
only the overflow is new.

## `#[deprecated]` on `Add`/`Sum` is not expressible

The plan called for deprecating the operator impls in favour of `try_add`.
**rustc rejects that outright**, verified with a standalone `--edition 2024`
reproduction before deciding:

```
error: `#[deprecated]` attribute cannot be used on trait impl blocks
error: `#[deprecated]` attribute cannot be used on trait methods in impl blocks
  = note: `#[deny(useless_deprecated)]` on by default
```

So all four impls (`Sum for PnL`, `Sum<&PnL> for PnL`, `Add for PnL`,
`Add for &PnL`) carry a `# Deprecated in favour of PnL::try_add` doc section
instead, stating the defect, why no fallible form of the std traits exists, the
`try_fold(PnL::default(), |acc, x| acc.try_add(x))` replacement, and the
compiler limitation itself so the next person does not retry it.

It warns nowhere in the crate, and would not have: #460 already routed every
in-library accumulation through `try_add`, which is why the operator is reachable
only from caller code.

## Public API

`public-api/optionstratlib.txt`: 81 insertions, 78 deletions, regenerated with
`make public-api-update` and committed here. Deduplicated across re-export
paths, the distinct changes are the signature moves in the table above.
`CHANGELOG.md` carries the breaking-change entry with the migration.

## Tests and gates

4049 lib + 71 property + 423 integration + 210 doctests, 0 failed.

```
cargo semver-checks --all-features                               no update required
```

## Stack

Chain A, last of four — and the last implementation PR of the programme.

- Stack: #470 → #469 → #463 → **#471** (this one). #442 is the capstone that
  follows once every chain has landed.
- Base branch: `main`. The three below it are merged, so this diff is its own
  work alone.

Closes #471

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -157,6 +157,95 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Twenty public methods whose return type had no error channel are now
+  fallible, which is a breaking change** (#471). Each aborted the process on
+  arithmetic over `pub` fields. #460 closed everything reachable through the
+  public constructors, and what remained was reachable only by writing a
+  field directly, by deserializing a document that carries one, or through a
+  trait whose signature forbade failure — so none of them could be fixed
+  without changing the signature, which is why they were reported rather than
+  fixed at the time. The values returned are unchanged for every input the
+  panicking forms accepted. To migrate, add `?` where the value feeds a
+  fallible function, or handle the `Result` at a boundary that cannot
+  propagate.
+
+  **`LegAble` gains an error channel on three methods.** `pnl_at_price`
+  returns `Result<Decimal, PricingError>` instead of `Decimal`, and
+  `total_cost` and `fees` return `Result<Positive, PositionError>` instead of
+  `Positive`. `pnl_at_price` computes `(price - cost_basis) * quantity` and
+  aborted with `Subtraction overflowed`; `total_cost` multiplies size by cost
+  basis before adding two fees, and aborted with `Positive arithmetic overflow
+  in add`. The four implementors — `SpotPosition`, `FuturePosition`,
+  `PerpetualPosition` and the `Leg` enum — are updated, and every one of the
+  three now uses the checked helpers. `Leg::pnl_at_price` and
+  `Leg::total_cost` propagate the option leg's failure instead of swallowing
+  it as `unwrap_or(Decimal::ZERO)` / `unwrap_or(Positive::ZERO)`; both
+  fallbacks were values a legitimate leg can also return, so neither could be
+  told apart from an answer. `Position::total_cost` and `Position::fees` have
+  returned `Result` since #470, so the leg trait was the last infallible
+  wrapper over the same sums.
+
+  **Five `PerpetualPosition` and `FuturePosition` methods follow.**
+  `FuturePosition::unrealized_pnl` and `PerpetualPosition::unrealized_pnl`
+  return `Result<Decimal, PricingError>`, and with them
+  `PerpetualPosition::roe_percentage`, `margin_ratio` and
+  `effective_leverage`. `unrealized_pnl` is the entire body of
+  `pnl_at_price` on those two types, so leaving it infallible would have
+  moved the abort one frame down rather than removed it. The three ratios
+  keep their existing degenerate answers — `Decimal::ZERO` for a zero margin
+  or a zero notional, `Decimal::MAX` for wiped-out equity — rather than
+  turning those into errors; only the overflow is new.
+
+  **`PnL::try_add` is now public, and the operator impls are documented as
+  superseded.** `impl Add for PnL` and `impl Sum for PnL` add `initial_costs:
+  Positive + Positive` with the raw operator. `Add::add` and `Sum::sum` are
+  fixed by `std` to return `Self`, so no fallible form of either exists and no
+  signature change makes them safe. #460 routed every accumulation inside the
+  library through a `pub(crate) try_add`; it is `pub` now so callers have the
+  same route. The four operator impls are kept, because removing them would
+  break every `a + b` and `.sum()` at a call site with nothing to overflow.
+  They carry a `# Deprecated in favour of PnL::try_add` doc section rather
+  than a `#[deprecated]` attribute: Rust rejects that attribute on a trait
+  `impl` block and on a trait method inside one (`error: #[deprecated]
+  attribute cannot be used on trait impl blocks`), so the compiler cannot
+  emit this warning at all.
+
+  **Three `Collar` methods.** `net_premium` returns `Result<Decimal,
+  PricingError>` instead of `Decimal`, and `is_zero_cost` and `is_credit`
+  return `Result<bool, PricingError>` instead of `bool`. `net_premium`
+  multiplies each leg's `premium` by its `quantity`, both `pub` fields, and
+  aborted on overflow; the two predicates read it. They report the same
+  failure rather than collapsing it into `false`, which would have read as a
+  definite classification.
+
+  **`CoveredCall::effective_cost_basis` and
+  `ProtectivePut::effective_cost_basis`** return `Result<Positive,
+  PositiveError>` instead of `Positive`. Both divide the premium by
+  `spot_leg.quantity`. Both constructors reject a zero share count, but a
+  strategy deserialized from JSON or mutated in place can carry one, and
+  `Positive` has no value meaning "undefined per-share figure".
+
+  **Four `PortfolioGreeks` methods.** `delta_gap` and `gamma_gap` return
+  `Result<Decimal, GreeksError>`, `combined` returns
+  `Result<PortfolioGreeks, GreeksError>`, and `add` returns `Result<(),
+  GreeksError>` instead of nothing. All five Greek fields are `pub` and are
+  written directly by callers aggregating from their own sources, so
+  `from_positions` guarantees nothing about them. `add` now writes through
+  `combined`, so a failure on the fourth of five sums leaves the receiver
+  exactly as it was rather than committing the first three.
+
+  No error enum and no error variant was added, so nothing that matches
+  exhaustively on a public error type needs a new arm. The property suite's
+  `test_spot_leg_strategies_never_panic` drops the bounded `spot_leg_money`
+  generator that existed only for these signatures and runs against the
+  unbounded `extreme_money`, so `Positive::MAX` is back in range for the
+  three spot-leg strategies.
+
+  Two siblings with the same defect are deliberately out of scope and remain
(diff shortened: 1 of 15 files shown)


src/model/leg/future.rs
```diff
@@ -299,16 +325,20 @@ impl LegAble for FuturePosition {
...
+            self.unrealized_pnl(price)?,
+            self.fees.to_dec(),
+            "FuturePosition::pnl_at_price/net",
+        )?)
     }
 
-    fn total_cost(&self) -> Positive {
-        self.total_margin_required() + self.fees
+    fn total_cost(&self) -> Result<Positive, PositionError> {
+        Ok(self.total_margin_required().checked_add(&self.fees)?)
```
joaquinbejar: 🔴 **Critical**: This will still abort before `checked_add` runs when `initial_margin_req` is `Positive::MAX` and `quantity` is greater than one, because `total_margin_required` uses raw `Positive` multiplication. The same leak remains in `PerpetualPosition::margin_ratio` and `effective_leverage`, which call `notional_value_at_price` and its raw multiplication. Move those products onto `checked_mul` and add overflow regressions so every new fallible signature returns its error.

src/strategies/covered_call.rs
```diff
@@ -292,14 +292,34 @@ impl CoveredCall {
...
+    ///
+    /// Returns [`PositiveError::ArithmeticError`] when the spot leg holds
+    /// zero shares, so there is no per-share premium to divide out, and when
+    /// `premium × quantity` overflows.
+    pub fn effective_cost_basis(&self) -> Result<Positive, PositiveError> {
+        let premium_per_share = self
+            .short_call
+            .premium
+            .checked_mul(&self.short_call.option.quantity)?
+            .checked_div(&self.spot_leg.quantity)?;
```
joaquinbejar: 🟠 **Warning**: `checked_div` selects the dependency default, so this call site still has no explicit rounding decision. The binding rules require monetary division to choose and document its strategy. Use `checked_div_with_strategy` here and in `ProtectivePut::effective_cost_basis`, then pin a repeating quotient in a regression test.

CHANGELOG.md
```diff
@@ -157,6 +157,95 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Twenty public methods whose return type had no error channel are now
```
joaquinbejar, line 160: 💡 **Suggestion**: This diff adds 930 lines across the model, PnL and three strategy APIs. Consider splitting the leg trait migration, the PnL alternative and the strategy getters so each public break has a smaller review and migration surface.

joaquinbejar (commented):
The checked arithmetic migration covers most edited paths, and the library plus targeted property tests pass. Two public Result paths can still panic through raw Positive multiplication, and the cost basis divisions leave rounding implicit. I also left a scope split suggestion for this 930 line API change. REQUEST_CHANGES. GitHub cannot record a formal change request from the PR author, so this review is submitted as COMMENT.

src/model/leg/future.rs
joaquinbejar: Confirmed and fixed. total_cost, margin_ratio and effective_leverage form their own checked products now, so the three Result paths report the overflow instead of aborting upstream of their own check, with regressions on each. One thing I did not do and want visible: the infallible helpers keep their raw arithmetic, and so does the Marginable impl beside them, whose initial_margin and maintenance_margin multiply exactly the same way. Closing that class means making Marginable fallible, which is a trait break past this issue rather than something to slip in here.

src/strategies/covered_call.rs
joaquinbejar: Right, and the choice was already made elsewhere in the crate: d_div rounds every Decimal division with MidpointNearestEven at DIV_DEFAULT_SCALE. The four per-share divisions in CoveredCall and ProtectivePut now use checked_div_with_strategy with the same strategy, so a per-share value rounds the same way whichever path computes it. Pinned with one contract of 3.50 over three shares, which repeats.

CHANGELOG.md
joaquinbejar: Keeping it whole, and the reason is specific to this one rather than a general preference. Splitting the leg trait migration from the PnL alternative would break the same public trait twice, since pnl_at_price calls unrealized_pnl and both have to move together for either to stop aborting. Two migrations for one defect is a worse deal for a downstream user than one larger diff, and the version bump covering it is already in main.

joaquinbejar:
Both defects fixed in 5b912030, the split declined for the reason in that thread. The public API snapshot regenerates with no diff, since the fix is caller-side rather than another signature change. What this PR does not close, stated so it is not mistaken for done: FuturePosition::total_margin_required, notional_value_at_entry, PerpetualPosition::notional_value_at_price and the Marginable impl for both types still multiply with the raw operator. They are infallible signatures with no error channel to fix, which is the next trait break rather than this one.

joaquinbejar (commented):
The follow-up closes the raw-multiplication panic paths, makes all four per-share rounding choices explicit, and adds focused overflow and repeating-quotient regressions; the full all-features library suite passes locally (4052 passed, 5 ignored). APPROVE. GitHub cannot record a formal approval from the PR author, so this review is submitted as COMMENT.

PR #483, opened 2026-08-30
Close the panic sweep: volatility, the panicking macros, and the lint that keeps them out (#442)

### LINKED ISSUE #442: Sweep the remaining modules for reachable panics

joaquinbejar:
## Summary

The capstone of the panic-freedom programme. #442's task list is now closed:
nine sibling issues landed the module sweeps, and this PR does the four things
left — the `volatility` sweep, the `unreachable!` / `panic!` inventory, the
silent-zero conversions in `Options`, and the lint that stops any of it
regressing.

| #442 box | closed by |
|---|---|
| `src/pricing/` exotics | #445 (issue #444) |
| `src/curves/` + `src/surfaces/` | #452 (#446) |
| `src/strategies/` | #468 (#460), #480 (#463) |
| `src/simulation/`, `src/chains/` beyond `build_chain` | #458 (#456) |
| **`src/volatility/`** | **here** |
| **the 14 `unreachable!` / 5 `panic!`** | **here** |
| **`make scan-banned` extended to the math operators** | **here** |
| **`Options::intrinsic_value`'s silent zero** | **here** |
| Upstream `expiration_date` 0.3.0 | still open — belongs in that crate |

## 1. `src/volatility/`: 287 panics → 0

Probed with `catch_unwind` over 3305 cases across the `Decimal` extremes, the
same method the rest of the programme used. Eight functions aborted:

| function | input | panic |
|---|---|---|
| `adjust_volatility` | `Microsecond → Custom(1e-27)` | `Positive arithmetic overflow in div` (179 hits) |
| `simulate_heston_volatility` | `dt = Decimal::MAX` | `Multiplication overflowed` (53) |
| `garch_volatility` | `returns = []` | `index out of bounds: the len is 0 but the index is 0` (13) |
| `annualized_volatility` | `vol = MAX`, `Microsecond` | `Positive arithmetic overflow in mul` (12) |
| `historical_volatility` | `window_size = 0` | `window size must be non-zero` (10) |
| `de_annualized_volatility` | `Custom(Positive::ZERO)` | `Positive invariant broken in div` (9) |
| `volatility_for_dt` | delegates to the above | (9) |
| `implied_volatility` | `max_iterations = i64::MAX` | `attempt to multiply with overflow` (2) |

**No signature changed** — all eight already returned `Result`. The zero-periods
guard reuses `adjust_volatility`'s existing message shape so the two entry
points agree on where the domain ends.

`tests/property/volatility_panic_freedom_test.rs` is the sweep, and it bites:
reverting `utils.rs` to `main` fails five of its six properties. The two that
pass either way cover `generate_ou_process` and `uncertain_volatility_bounds`,
which were already clean.

## 2. Every `unreachable!` and `panic!`, disposed of individually

Each is now a typed error **or** carries a `// scan-banned: allow -- <reason>`
with the argument for why it cannot fire. A marker without a real argument is
worse than the panic, because it looks reviewed.

The `src/constants.rs` group is the interesting one: four are in `const` items,
where the `match` is compiler-evaluated, so a failing arm is `evaluation of
constant value failed` at **build** time — there is no run time at which the arm
exists. The other seven are `LazyLock` cells whose closure captures nothing and
whose only input is a `dec!()` literal. No `const` alternative exists:
`positive` keeps `from_decimal_const` crate-private, and a `Positive::ZERO`
stand-in would make every annualisation divide by zero.

## 3. Three silent zeros in `src/model/option.rs`, not one

The issue names `intrinsic_value`. There are three:
`Decimal::from_f64(payoff).unwrap_or_default()` and the same pattern for
`price` and `iv`. A non-finite `f64` became `0` with no error — silently wrong
rather than a panic, which this programme treats as the worse failure. All three
report now.

## 4. `make scan-banned` extended — and it found things

It guarded `.unwrap()` / `.expect()`. It now also rejects `panic!`,
`unreachable!`, `todo!`, `unimplemented!`, `.exp()`, `.ln()`, `.powd(` and
`.sqrt().unwrap()`.

The lint and the fixes it demands land together, deliberately: a lint alone is a
failing lint, and fixes alone leave nothing enforcing them.

Two findings worth naming:

- **Its comment filter skipped every line starting with `*`**, which swallowed
  the continuation lines of multi-line products — **five `.exp()` calls in
  `src/pricing/compound.rs` were invisible to it.** Fixed.
- The `f64` methods share three of those names and do not abort, so those call
  sites carry a marker saying so; grep cannot tell receiver types apart.

## 5. `make doc` now builds documentation

It was `cargo clippy -- -W missing-docs`, which builds no docs. It is now
`cargo doc --all-features --no-deps`.

This matters beyond the name. `create-doc` — the target `pre-push` actually
reaches, via `readme` — has no `--all-features`, so **a doc link behind
`plotly`, `static_export` or `async` was checked by nothing**, locally or in CI.
Demonstrated by injecting `[`this_symbol_does_not_exist`]` into
`src/visualization/plotly.rs` on `main`:

```
make create-doc                     exit 0    link never mentioned
cargo doc --all-features --no-deps  exit 101  3 occurrences
```

**The new gate starts green**, which is the point — a gate that starts red
teaches people to ignore it:

```
$ make doc            # on this branch
warning: public documentation for `new` links to private item `RNDStatistics::new`
warning: public documentation for `BreakEvenable` links to private item `lower_break_even`
exit 0
```

Those two warnings are pre-existing and named in `AGENTS.md` so nobody
re-investigates them.

`AGENTS.md` gains five facts, each with the command that demonstrates it and
nothing else — a reference, not a narrative:

1. what `make doc` did, and the feature-gated hole in `create-doc` above
2. `make public-api-check` builds rustdoc with `--all-features` but
   `cargo public-api` prints rustdoc's errors and exits 0 — `error:` lines
   appear under a green check
3. `cargo test --all-features` never exercises the default feature set;
   `make test` runs default, `plotly` and `static_export,plotly` separately
4. `cargo-semver-checks` compares against the **published** baseline, so a
   rename or a variant on an exhaustive public enum needs a minor bump on a 0.x
   crate — `make public-api-check` compares against the snapshot and never asks
5. `gh pr merge` reports a base-branch policy refusal on **stderr only**, so a
   caller redirecting stderr sees a silent no-op

## Gates

```
cargo test --all-features --workspace       4058 lib + 78 property + 423 + 210 doctests, 0 failed
LOGLEVEL=WARN cargo test  (default set)     4062 + 78 + 396 + 210, 0 failed
cargo fmt --all --check                     clean
cargo build --release                       no warnings
make doc                                    exit 0, the two known warnings
make scan-banned                            OK: no unwrap/expect, no panic/unreachable/todo/
                                            unimplemented, no unchecked exp/ln/powd/sqrt
make public-api-check                       OK
```

## Public API impact

**One breaking change, not none.** An earlier revision of this section said the
API was unchanged; it is not, and the snapshot carries the difference:

| before | after |
|---|---|
| `DecimalStats::mean(&self) -> Decimal` | `-> Result<Decimal, DecimalError>` |
| `DecimalStats::std_dev(&self) -> Decimal` | `-> Result<Decimal, DecimalError>` |

Both aborted on a sample they could not sum — `mean` through `iter().sum()`,
`std_dev` through `.powd(Decimal::TWO)` — and `vec![Decimal::MAX; 2]` reached
each. A trait method cannot report that without a return type that can carry
it.

**Migration.** Add `?` where the value feeds a fallible function, or
`.expect("…")` at a boundary that cannot propagate. An external implementor of
`DecimalStats` changes two signatures and returns `Ok(..)`.

**Versioning.** No further bump. 0.21.0 is already the breaking version for
this cycle, set in #473 for the `Surface::project_onto` rename, and it is
unreleased — the published baseline is still 0.20.0. `cargo semver-checks`
reads `0.20.0 -> 0.21.0 (major change)` and passes; a second bump inside one
unreleased cycle would say something untrue about what shipped.

## Stack

The capstone. Every other chain has merged: #459/#472, #466/#473, #462/#474,
#470/#475, #451/#477, #469/#478, #463/#480, #453/#481, #471/#482.

Closes #442

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,16 +61,38 @@ clean:
 .PHONY: check
 check: test fmt-check lint scan-banned
 
-# Fails when a panicking construct reappears in production code. A bare
-# `#[cfg(test)]` never truncates the scan; only the braced body of the item
-# it actually gates is skipped, brace counted (files may carry several):
+# Fails when a panicking construct reappears in production code.
+#
+# The banned set, and why each entry is there:
+#   * `.unwrap()` / `.expect(` — abort instead of reporting.
+#   * `panic!` / `unreachable!` / `todo!` / `unimplemented!` — the same,
+#     spelled out. The pattern requires a non-word, non-`_` character before
+#     the macro name so `pos_or_panic!` is not swept up with `panic!`.
+#   * `.exp()` / `.ln()` / `.powd(` — `rust_decimal`'s `MathematicalOps`
+#     aborts on all three (`Exp overflowed`, `Unable to calculate ln for
+#     zero`, `Pow overflowed`); `d_exp` / `d_ln` / `d_powd` in
+#     `src/model/decimal.rs` are the checked forms. `f64` has the same three
+#     method names and does *not* abort, so those call sites carry a marker
+#     saying so — grep cannot tell the receiver types apart.
+#   * `.sqrt().unwrap()` — `Decimal::sqrt` is the safe one, returning
+#     `Option`; unwrapping it puts the panic back.
+#
+# A bare `#[cfg(test)]` never truncates the scan; only the braced body of the
+# item it actually gates is skipped, brace counted (files may carry several):
 # `mod`/`fn`/`impl`/`trait`/`struct`/`enum`/`union` (any `pub(..)` visibility,
 # `async`/`unsafe`, generics or return type before the `{`, spanning multiple
 # signature lines if needed), and a bare `#[cfg(test)] { ... }` block. A
 # body-less item behind `#[cfg(test)]` (`use`, `const`, `static`, `type`
 # alias, or a `;`-terminated declaration such as a tuple struct or a trait
 # method signature) skips nothing; scanning resumes right after it as usual.
-# Doc comments are skipped too. A reviewed exception carries a trailing
+#
+# Line comments (`///`, `//!`, `//`) are skipped, and so is the closing line
+# of a `/*** … ***/` banner (`*`-run followed by `/`). The filter used to skip
+# every line starting with `*`, which also skipped the continuation lines of a
+# multi-line product — `        * asr.exp()` in `src/pricing/compound.rs` is
+# five such lines — so a banned construct could hide there.
+#
+# A reviewed exception carries a trailing
 # `// scan-banned: allow -- <reason>` marker on the same line.
 .PHONY: scan-banned
 scan-banned:
@@ -117,15 +139,15 @@ scan-banned:
 			} \
 		' "$$f"; \
 	done \
-		| grep -E '\.unwrap\(\)|\.expect\(' \
-		| grep -v -E ':[0-9]+:[[:space:]]*(///|//!|//|\*)' \
+		| grep -E '\.unwrap\(\)|\.expect\(|\.exp\(\)|\.ln\(\)|\.powd\(|\.sqrt\(\)\.unwrap\(\)|[^_[:alnum:]](panic|unreachable|todo|unimplemented)!' \
+		| grep -v -E ':[0-9]+:[[:space:]]*(///|//!|//|\*+/)' \
 		| grep -v 'scan-banned: allow' || true); \
 	if [ -n "$$found" ]; then \
 		echo "Banned patterns found in production code:"; \
 		echo "$$found"; \
 		exit 1; \
 	fi; \
-	echo "OK: no .unwrap()/.expect() in production code"
+	echo "OK: no unwrap/expect, no panic/unreachable/todo/unimplemented, no unchecked exp/ln/powd/sqrt in production code"
 
 # Pinned producers of public-api/optionstratlib.txt. Both anchors are needed
 # and they only work as a pair: `cargo public-api` does not read the source,
--- a/Makefile
+++ b/Makefile
@@ -237,9 +237,19 @@ fix:
 .PHONY: pre-push
 pre-push: fix fmt lint-fix test readme doc
 
+# Builds the crate's documentation with every feature on. It used to run
+# `cargo clippy -- -W missing-docs`, which builds no documentation at all and
+# so never resolved an intra-doc link. `--all-features` is what makes the
+# difference: `create-doc` below omits it, so a link inside `plotly` /
+# `static_export` / `async` code was checked by nothing.
+#
+# The target starts green: two `private_intra_doc_links` warnings stand
+# (`RNDStatistics::new` in src/chains/rnd.rs and `lower_break_even` in
+# src/strategies/base.rs), and warnings do not fail it. `src/lib.rs` denies
+# `rustdoc::broken_intra_doc_links`, so a broken link is an error and exits 101.
 .PHONY: doc
 doc:
-	cargo clippy -- -W missing-docs
+	cargo doc --all-features --no-deps
 
 .PHONY: doc-open
 doc-open:
--- a/Makefile
+++ b/Makefile
@@ -141,7 +141,14 @@ scan-banned:
 	done \
 		| grep -E '\.unwrap\(\)|\.expect\(|\.exp\(\)|\.ln\(\)|\.powd\(|\.sqrt\(\)\.unwrap\(\)|[^_[:alnum:]](panic|unreachable|todo|unimplemented)!' \
 		| grep -v -E ':[0-9]+:[[:space:]]*(///|//!|//|\*+/)' \
-		| grep -v 'scan-banned: allow' || true); \
+		| grep -v -E 'scan-banned: allow -- [^[:space:]]' || true); \
+	malformed=$$(grep -rn 'scan-banned: allow' src \
+		| grep -v -E 'scan-banned: allow -- [^[:space:]]' || true); \
+	if [ -n "$$malformed" ]; then \
+		echo "Exemption markers without a reason (use 'scan-banned: allow -- <reason>'):"; \
+		echo "$$malformed"; \
+		exit 1; \
+	fi; \
 	if [ -n "$$found" ]; then \
 		echo "Banned patterns found in production code:"; \
 		echo "$$found"; \
--- a/src/model/decimal.rs
+++ b/src/model/decimal.rs
@@ -7,7 +7,7 @@ use crate::error::decimal::DecimalError;
 use crate::geometrics::HasX;
 use num_traits::{FromPrimitive, ToPrimitive};
 use rand::distr::Distribution;
-use rand_distr::Normal;
+use rand_distr::StandardNormal;
 use rust_decimal::{Decimal, MathematicalOps, RoundingStrategy};
 use rust_decimal_macros::dec;
 
@@ -65,24 +65,24 @@ macro_rules! assert_decimal_eq {
 /// ```rust
 /// use rust_decimal::Decimal;
 /// use rust_decimal_macros::dec;
+/// use optionstratlib::error::DecimalError;
 /// use optionstratlib::model::decimal::DecimalStats;
 ///
 /// struct DecimalSeries(Vec<Decimal>);
 ///
 /// impl DecimalStats for DecimalSeries {
-///     fn mean(&self) -> Decimal {
-///         let sum: Decimal = self.0.iter().sum();
+///     fn mean(&self) -> Result<Decimal, DecimalError> {
 ///         if self.0.is_empty() {
-///             dec!(0)
-///         } else {
-///             sum / Decimal::from(self.0.len())
+///             return Ok(dec!(0));
 ///         }
+///         let sum: Decimal = self.0.iter().sum();
+///         Ok(sum / Decimal::from(self.0.len()))
 ///     }
-///     
-///     fn std_dev(&self) -> Decimal {
+///
+///     fn std_dev(&self) -> Result<Decimal, DecimalError> {
 ///         // Implementation of standard deviation calculation
 ///         // ...
-///         dec!(0) // Placeholder return
+///         Ok(dec!(0)) // Placeholder return
 ///     }
 /// }
 /// ```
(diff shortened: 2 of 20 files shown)


src/volatility/utils.rs
```diff
@@ -630,11 +702,13 @@ pub fn adjust_volatility(
...
 
     // Scale factor is square root of (from_periods / to_periods). `Positive`
-    // already implements `sqrt()`, so skip the f64 round-trip and keep the
-    // computation in `Decimal` precision end-to-end.
-    let scale_factor = (from_periods / to_periods).sqrt();
+    // has a checked square root, so skip the f64 round-trip and keep the
+    // computation in `Decimal` precision end-to-end. Both steps are checked:
+    // the ratio overflows for a `Custom` target of 1e-27, and the operator
+    // form aborted there.
+    let scale_factor = from_periods.checked_div(&to_periods)?.checked_sqrt()?;
```
joaquinbejar: 🔴 **Critical**: This will silently return zero for `Year` to `Custom(Positive::MAX)`. `checked_div` rounds `1 / MAX` to 28 decimal places before the square root, so the ratio becomes zero even though `sqrt(1 / MAX)` is about `3.55e-15` and is representable. Take the two square roots first, divide them with an explicit rounding strategy, and pin this case in a regression test. The division in `de_annualized_volatility` needs the same explicit strategy choice.

src/model/decimal.rs
```diff
@@ -91,36 +91,71 @@ pub trait DecimalStats {
     ///
     /// The mean is the sum of all values divided by the count of values.
     /// This method should handle empty collections appropriately.
-    fn mean(&self) -> Decimal;
+    ///
+    /// # Errors
+    ///
+    /// Implementations return [`DecimalError`] when the running sum or the
+    /// division by the count leaves the representable `Decimal` range.
+    fn mean(&self) -> Result<Decimal, DecimalError>;
```
joaquinbejar, line 99: 🔴 **Critical**: This changes the public trait contract from `Decimal` to `Result<Decimal, DecimalError>`, so every external implementation and caller must change. The PR summary currently says the public API is unchanged and the version stays at 0.21.0. Call out the break explicitly and add the required migration and versioning updates before merge.

tests/property/volatility_panic_freedom_test.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// window zero, which `slice::windows` rejects by aborting, and at a
+    /// window wider than the sample.
+    #[test]
+    fn test_sample_estimators_never_panic(
+        returns in degenerate_returns(),
+        lambda in extreme_return(),
+        window in 0usize..6,
+    ) {
+        let _ = constant_volatility(&returns);
+        let _ = historical_volatility(&returns, window);
```
joaquinbejar, line 150: 🟠 **Warning**: These properties only discard the results, so every new error path could regress to a silent `Ok(0)` or the wrong error variant and the suite would still pass. Add focused assertions for the documented zero window, empty GARCH sample, Heston overflow, zero period deannualization, rescaling overflow, and IV grid overflow cases.

Makefile
```diff
@@ -117,15 +139,15 @@ scan-banned:
 			} \
 		' "$$f"; \
 	done \
-		| grep -E '\.unwrap\(\)|\.expect\(' \
-		| grep -v -E ':[0-9]+:[[:space:]]*(///|//!|//|\*)' \
+		| grep -E '\.unwrap\(\)|\.expect\(|\.exp\(\)|\.ln\(\)|\.powd\(|\.sqrt\(\)\.unwrap\(\)|[^_[:alnum:]](panic|unreachable|todo|unimplemented)!' \
+		| grep -v -E ':[0-9]+:[[:space:]]*(///|//!|//|\*+/)' \
 		| grep -v 'scan-banned: allow' || true); \
```
joaquinbejar: 🟠 **Warning**: This accepts any line containing `scan-banned: allow`, although the PR requires `scan-banned: allow -- <reason>`. A bare marker can therefore suppress a new `panic!` without the review argument this gate is meant to enforce. Only exempt markers with a nonempty reason and fail malformed markers separately.

joaquinbejar (commented):
The local format, lint, library and focused property checks pass, but this still has two correctness blockers and two coverage gaps. A representable timeframe conversion collapses to zero, the public DecimalStats contract changes despite the summary saying the API is unchanged, the new error contracts are not asserted, and the allow marker can bypass the lint without a reason. This is also over a thousand changed lines across four concerns, so splitting the volatility, payoff, lint and documentation work would make the fixes easier to validate. REQUEST_CHANGES. GitHub does not allow a formal change request on an author owned PR, so I am submitting the review as COMMENT.

src/volatility/utils.rs
joaquinbejar: Confirmed and fixed. sqrt(from / to) rounds the ratio to 28 decimal places before the root can rescue it, so 1/MAX at about 1.26e-29 became zero while sqrt of it is around 1.1e-15 and representable. It takes the two roots first and divides them with MidpointNearestEven, the strategy d_div applies crate-wide, and de_annualized_volatility carries it too. I checked the regression both ways: it fails on the old ratio-then-root form and passes on this one.

src/model/decimal.rs
joaquinbejar: The claim was wrong, the code was not. The changelog already declared the two DecimalStats signatures and the snapshot carries them; it was the PR summary that said the API was unchanged. Rewritten with a before and after table and the migration, and the changelog entry now carries the migration too. On versioning: no further bump, because 0.21.0 is already the breaking version for this cycle, set in #473, and it is unreleased — the published baseline is still 0.20.0 and cargo semver-checks reads 0.20.0 to 0.21.0 as a major change and passes. A second bump inside one unreleased cycle would say something untrue about what shipped.

tests/property/volatility_panic_freedom_test.rs
joaquinbejar: Right, discarding the result means a silent Ok(0) passes as well as the right answer. Five focused cases added. Worth reporting that three of my first assertions were wrong about the contract rather than finding a defect: an oversized window yields an empty series because slice::windows produces none, a sample under two observations has no variance and answers zero by design, and annualisation multiplies by the period count so a zero one answers zero where de_annualized_volatility and adjust_volatility both reject it. I pinned that asymmetry rather than evening it out, since it is the existing public contract and not something this PR introduced.

Makefile
joaquinbejar: Fixed. The exemption now requires scan-banned: allow -- with a non-empty reason, and a bare marker fails as its own error naming the file and line rather than silently exempting. Verified by planting one in src/volatility/utils.rs: make exits 1 with the marker listed.

joaquinbejar:
All four addressed in 49d5d808. On the fifth point, the split: keeping it whole. The lint change and the sweep are one thing — extending scan-banned to the math operators is what finds the volatility calls the sweep then fixes, and landing either alone gives you a failing lint or fixes with nothing enforcing them. The AGENTS.md entries are the record of what these gates do and do not cover, which is only checkable against the gates as they stand after this PR. The Makefile fix I would have taken separately, but it is two lines inside a Makefile change this PR already makes for the lint, so splitting it means reviewing the same file twice.

Issue #91, opened 2025-01-06
 Modify Black-Scholes Functions and Traits to Use `Result<Decimal, Box<dyn Error>>`


joaquinbejar:
## Description
Update the functions and traits in the `Black-Scholes` module to return `Result<Decimal, Box<dyn Error>>` instead of `f64`. This change enhances error handling and improves the precision of financial calculations. The update aims to provide consistent and reliable error propagation across all functions and traits.

---

### Refactoring:
- [ ] Update all relevant functions to return `Result<Decimal, Box<dyn Error>>`.
- [ ] Replace all internal `f64` calculations with `Decimal` for improved precision.
- [ ] Add error handling to manage scenarios such as:
  - Missing or invalid inputs (e.g., time to expiration or volatility).
  - Edge cases, such as zero volatility or division by zero.

### Trait Updates:
- [ ] Modify the `BlackScholes` trait to reflect the new return type.
- [ ] Ensure all trait implementations comply with the updated signature.

### Testing:
- [ ] Update existing test cases to validate the new return type.
- [ ] Add test cases for error scenarios to confirm appropriate error handling.
- [ ] Validate that results using `Decimal` are consistent with `f64` benchmarks, accounting for precision improvements.

### Documentation:
- [ ] Update function and trait documentation to reflect the new return type.
- [ ] Add usage examples that demonstrate handling `Result` types effectively.
- [ ] Highlight the benefits of `Decimal` for financial calculations in the documentation.

---

## Additional Notes
- Ensure backward compatibility where feasible or provide migration guidance for users.
- Optimize for performance to mitigate potential overhead introduced by the use of `Decimal`.
- Maintain alignment with existing project coding standards and guidelines.

---

## Labels
`refactor`, `error-handling`, `decimal`, `enhancement`, `testing`, `documentation`

PR #488, opened 2026-08-30
Quote a worthless option at zero instead of dropping its price (#487)

### LINKED ISSUE #487: A worthless option is priced as absent, which truncates the strike ladder
## Overview

`OptionData::calculate_prices` turns a non-positive Black-Scholes price into **no price at all**:

```rust
// src/chains/optiondata.rs:1003
self.call_middle = Positive::new_decimal(price).ok();
```

Black-Scholes on `Decimal` undershoots by an epsilon for an option that is worth nothing. Measured for a call 300 points out of the money at seven and a half hours to expiry:

```
call price at strike 5100: -0.0000000000000000000000002992
```

`Positive::new_decimal` rejects it, the side is left `None`, and a contract that is simply worthless becomes a contract that does not quote.

## Why it matters

`OptionChain::build_chain` stops extending the ladder when both wings report `some_price_is_none()`:

```rust
// src/chains/chain.rs:597
if next_upper_option_data.some_price_is_none()
    && next_lower_option_data.some_price_is_none()
{
    break;
}
```

Each wing is missing only its worthless side, so two perfectly quotable strikes count as unquotable and the build stops early. `chain_size` then silently means fewer strikes than it says:

```
spot 5100, strike_interval 25, chain_size 20, spread 0.02, days 0.3125
  → 23 strikes, not 41
    strike 4825  call Some(275.11)/Some(275.13)  put None/None
    strike 5375  call None/None                  put Some(274.85)/Some(274.87)
```

The side that disappears is always the one that is out of the money.

This is the same decision #439 settled one layer down — a thin quote is widened to the tick instead of being withdrawn — applied to the price that feeds it. #439 fixed the withdrawal in `apply_spread`; the withdrawal in `calculate_prices` remains.

## Not to be confused with

- **The tenor-dependent truncation of 0.20.0.** That was the `apply_spread` erasure and #439 fixed it: at spot 100 with `strike_interval` 1.0 the ladder is complete at every tenor on 0.21.0. This issue needs wings closer to the money (about 6 percent out) before the epsilon appears, which is why a spot-100 fixture does not show it.
- **A ladder that reaches strike zero.** With `strike_interval` 5 and `chain_size` 20 at spot 100 the twentieth lower strike lands on 0 and the loop stops before inserting it: 40 strikes, not 41. That is correct — a strike of zero is not a contract — and is a property of the grid, not this bug.

## Proposal

Read a successful but non-positive price as **zero**: the option is worth nothing, which is a price, and the tick floor in `apply_spread` then quotes it as a real market would. A genuine pricing FAILURE stays a failure, on the `Err` arm, and keeps producing no price.

## Acceptance criteria

- [ ] A call worth nothing comes back quoted at zero rather than absent, on both the call and the put side
- [ ] `chain_size` means the same number of strikes at every tenor: a request for 20 per side returns 41 strikes at 30, 1, 0.3125, 0.05 and 0.01 days
- [ ] The wing at 5100 in the configuration above is present with both sides quoted
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` clean

## Consumer context

OptionChain-Simulator issue #91 pins a strike ladder at creation so a client holding a position keeps its contracts across steps. It builds a chain wide enough to reach the pinned strikes and filters; the truncation above removes the far strikes before the filter can see them, so the pinned ladder loses exactly the wings a defined-risk position needs.

joaquinbejar:
Closes #487

## What changed

`OptionData::calculate_prices` mapped a non-positive Black-Scholes price to `None`:

```rust
self.call_middle = Positive::new_decimal(price).ok();
```

Black-Scholes on `Decimal` undershoots by an epsilon for an option worth nothing. Measured for a call 300 points out of the money at seven and a half hours to expiry: `-2.992e-25`. The side was left with no price, so a worthless contract became an unquotable one.

A successful but non-positive price now reads as **zero**, through a small `worthless_or` helper. The tick floor added in #439 then quotes it the way a real market does. A genuine pricing failure is untouched: the `Err` arm still produces no price at all.

## Why it truncated the ladder

`build_chain` stops when both wings report `some_price_is_none()`. Each wing was missing only its out-of-the-money side, so two quotable strikes counted as unquotable:

```
spot 5100, strike_interval 25, chain_size 20, spread 0.02, days 0.3125
  before → 23 strikes
    strike 4825  call Some(275.11)/Some(275.13)  put None/None
    strike 5375  call None/None                  put Some(274.85)/Some(274.87)
  after  → 41 strikes
```

## What this is NOT

- **Not the 0.20.0 truncation.** That one was the `apply_spread` erasure and #439 fixed it: at spot 100 with `strike_interval` 1.0 the ladder is already complete at every tenor on `main`. This defect needs wings around 6 percent out of the money before the epsilon appears, which is why a spot-100 fixture does not reach it. I checked the fix against that fixture too and the numbers do not move.
- **Not the strike-zero case.** With `strike_interval` 5 and `chain_size` 20 at spot 100 the twentieth lower strike lands on zero and the loop stops before inserting it: 40 strikes, not 41. Correct, and untouched here.

## Tests

Three, all failing before the change and passing after:

- `tests_worthless_or` — a real price passes through, and the measured `-2.992e-25` reads as zero
- `test_a_worthless_wing_is_quoted_rather_than_dropped` — the 5100 wing in the configuration above comes back with both sides quoted. Driven with `spread = 0.02`, deliberately, so it cannot be read as an artefact of asking for no spread: the wing came back unpriced either way
- `test_build_chain_returns_the_requested_width_at_every_tenor` — `chain_size` 20 means 41 strikes at 30, 1, 0.3125, 0.05 and 0.01 days. This one also covers the 0.20.0 shape of the bug, so the regression cannot come back through the other path

I verified the first two fail on `main` by reverting only the two assignment lines and keeping the tests.

## A second defect, left for a follow-up

`some_price_is_none()` is true when **any** of the four prices is missing, and `build_chain` uses it to mean "this wing does not quote". A row missing only its worthless side is not a row that fails to quote. This change makes the predicate almost unreachable rather than correct, so I have not touched it here; happy to open it as its own issue if you want the predicate itself narrowed.

## Ships as 0.21.1

The bump is in this branch: `Cargo.toml`, the crate docs in `src/lib.rs`, a CHANGELOG entry, and `README.md` regenerated with `make readme` rather than edited. A patch release, since the change is a fix with no public API movement — `worthless_or` is private.

Releasing it rather than holding it is what lets OptionChain-Simulator take the fix at all: 0.21.0 is already on crates.io, so the consumer would otherwise have to point at a git commit.

## Gate

`cargo test --workspace` 4064 + 83 + 423 + 210 pass, 0 failed; `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean; `cargo doc --no-deps` clean.

## Consumer

OptionChain-Simulator #91 pins a strike ladder at creation so a client holding a position keeps its contracts across steps. It builds wide and filters to the pinned strikes; this truncation removed the far strikes before the filter could see them. With this change its end-to-end test passes.

### DIFF
--- a/src/chains/optiondata.rs
+++ b/src/chains/optiondata.rs
@@ -352,6 +352,26 @@ fn mid_within(anchor: Positive, bid: Positive, ask: Positive) -> Positive {
     anchor.max(bid).min(ask)
 }
 
+/// A theoretical price as a quotable one, with a worthless option read as zero.
+///
+/// Black-Scholes on `Decimal` undershoots by an epsilon for an option that is
+/// worth nothing: a call 300 points out of the money at seven and a half hours
+/// to expiry prices at `-2.992e-25`. That is not a value, it is the numeric
+/// floor of "worth nothing", and rejecting it used to leave the side with no
+/// price at all.
+///
+/// Zero is the honest reading, and it is the same decision issue #439 made one
+/// layer down, where a thin quote is widened to a tick instead of being
+/// withdrawn: a contract nobody would pay for is still a contract, and it is
+/// quoted at nothing rather than removed from the chain.
+///
+/// A genuine pricing FAILURE is still a failure, and is handled by the `Err`
+/// arm of the caller; this only reinterprets a successful, non-positive price.
+#[must_use]
+fn worthless_or(price: Decimal) -> Positive {
+    Positive::new_decimal(price).unwrap_or(Positive::ZERO)
+}
+
 impl OptionData {
     /// Creates a new instance of `OptionData` with the given option market parameters.
     ///
@@ -1000,7 +1020,7 @@ impl OptionData {
         let call_option = self.get_option(Side::Long, OptionStyle::Call)?;
         match call_option.calculate_price_black_scholes() {
             Ok(price) => {
-                self.call_middle = Positive::new_decimal(price).ok();
+                self.call_middle = Some(worthless_or(price));
                 self.call_ask = self.call_middle;
                 self.call_bid = self.call_middle;
             }
@@ -1015,7 +1035,7 @@ impl OptionData {
         let put_option = self.get_option(Side::Long, OptionStyle::Put)?;
         match put_option.calculate_price_black_scholes() {
             Ok(price) => {
-                self.put_middle = Positive::new_decimal(price).ok();
+                self.put_middle = Some(worthless_or(price));
                 self.put_ask = self.put_middle;
                 self.put_bid = self.put_middle;
             }
@@ -2540,6 +2560,34 @@ mod tests_get_option_for_iv {
         assert!(result.is_ok());
         let option = result.unwrap();
         assert_eq!(option.side, Side::Short);
+    }
+}
+
+#[cfg(test)]
+mod tests_worthless_or {
+    use super::*;
+    use positive::pos_or_panic;
+    use rust_decimal_macros::dec;
+
+    /// A real price passes through untouched.
+    #[test]
+    fn test_a_positive_price_is_kept() {
+        assert_eq!(worthless_or(dec!(12.34)), pos_or_panic!(12.34));
+        assert_eq!(worthless_or(Decimal::ZERO), Positive::ZERO);
+    }
+
+    /// The negative epsilon Black-Scholes returns for a worthless option reads
+    /// as zero, not as an absent price.
+    ///
+    /// `-2.992e-25` is the value measured for a call 300 points out of the
+    /// money at seven and a half hours to expiry.
+    #[test]
+    fn test_a_worthless_price_reads_as_zero() {
+        assert_eq!(
+            worthless_or(dec!(-0.0000000000000000000000002992)),
+            Positive::ZERO
+        );
+        assert_eq!(worthless_or(dec!(-1)), Positive::ZERO);
     }
 }
 
--- a/src/chains/optiondata.rs
+++ b/src/chains/optiondata.rs
@@ -15,6 +15,7 @@ use crate::{ExpirationDate, OptionStyle, Options, Side};
 use chrono::{DateTime, Utc};
 use positive::Positive;
 use rust_decimal::Decimal;
+use rust_decimal_macros::dec;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::cmp::Ordering;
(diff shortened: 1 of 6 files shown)


joaquinbejar (commented):
The zero-price treatment fixes the measured epsilon case and the ladder regressions are well targeted, but the helper also turns materially negative prices into valid quotes and hides a pricing invariant failure. REQUEST_CHANGES.

src/chains/optiondata.rs
```diff
@@ -352,6 +352,26 @@ fn mid_within(anchor: Positive, bid: Positive, ask: Positive) -> Positive {
...
+/// Zero is the honest reading, and it is the same decision issue #439 made one
+/// layer down, where a thin quote is widened to a tick instead of being
+/// withdrawn: a contract nobody would pay for is still a contract, and it is
+/// quoted at nothing rather than removed from the chain.
+///
+/// A genuine pricing FAILURE is still a failure, and is handled by the `Err`
+/// arm of the caller; this only reinterprets a successful, non-positive price.
+#[must_use]
+fn worthless_or(price: Decimal) -> Positive {
+    Positive::new_decimal(price).unwrap_or(Positive::ZERO)
```
joaquinbejar: 🟠 **Warning**: This turns every negative pricing result into a valid zero quote, not just the documented rounding epsilon; the new test even accepts `-1`. A materially negative Black-Scholes result signals a broken pricing invariant and will now be silently published as a market price. Clamp only a documented small tolerance around zero and propagate or reject larger negatives.

joaquinbejar: You are right, and the test I wrote froze the wrong behaviour by accepting -1. A price is now read three ways: positive is the price, zero or within a documented 1e-12 of it is an option worth nothing and is quoted as zero, and anything more negative leaves the side unpriced with a warning naming the strike and the value, since that is a broken pricing invariant rather than a free option.

The bound sits thirteen orders of magnitude above the -2.992e-25 artefact this was written for and ten below one tick, and a test now asserts that -1, -0.01 and -0.000001 all refuse to become quotes.

joaquinbejar:
Both points addressed and pushed.

The clamp is bounded now, so only rounding noise becomes a zero quote and a materially negative price leaves the side unpriced with a warning; the test that accepted -1 was the clearest sign the helper was too broad, and it now asserts the opposite. Full suite green at 4065 + 83 + 423 + 210, clippy and fmt clean.

Issue #570, opened 2026-09-19
calculate_price_probability and expected_value floor an inverted CDF difference to zero


joaquinbejar:
## Overview

Two sibling sites reach the same event #569 fixed in `ProfitLossRange::calculate_probability`, and both answer it with a silent zero instead of an error:

- `src/strategies/probabilities/utils.rs:248-252`, `calculate_price_probability`
- `src/strategies/probabilities/core.rs:190-194`, `expected_value`

Each subtracts the probability below the lower bound from the probability below the upper bound through `sub_floor_zero`, with a comment attributing any negative result to "a difference of one ulp" from the `Decimal -> f64 -> Decimal` round trip.

## Why the attribution is wrong

The inversion those sites are flooring is not one ulp. On the #569 input (spot `Positive::MAX`, strikes zero, volatility `1e-28`) the price model returns `0.9987` below the lower bound and `0.5` below the upper one: `(MAX - 4) / MAX` rounds to `0.9999999999999999999999999999` at `Decimal`'s 28 places, and `Decimal::checked_ln` returns `+9e-28` for it where the true value is `-1e-28`, so the lower bound lands three standard deviations above the spot.

Measured on that input, `calculate_price_probability` returns `Ok((0.998650101968426, 0, 0.5))`: below + in + above = **1.4987**, silently. That is the probability-nobody-computed outcome #569 refused one function away.

## Context

- #569 made `ProfitLossRange::calculate_probability` report the inversion as `ProbabilityCalculationErrorKind::InvalidProbability`, on the grounds that a distribution function is monotone and a smaller probability at the upper bound is a result outside the model's precision.
- The two comments now give contradictory explanations of the same phenomenon: one says rounding, the other says precision loss in `checked_ln`.
- Ordinary inputs cannot reach it: a probe over 21,168 combinations (spot 0.5..1e5, vol 0.01..3, T 0.5d..3650d) found 0 inversions and 9,555 exact equalities. The floor is only ever taken on inputs past the model's precision.

Found by `panic-hunter` while reviewing the #569 diff.

## Tasks

- [ ] Decide whether the three sites should share one rule (error on inversion, zero on equality) or whether the two strategy-layer sites have a reason to differ; record it in the comments either way
- [ ] Correct the "one ulp" attribution in both comments to the actual mechanism
- [ ] Pin the #569 input against `calculate_price_probability` and `expected_value` so the three-way sum is asserted to be `1` or the call to be an error, never `1.4987`

[...]

PR #571, opened 2026-09-19
Report an inverted range probability instead of aborting on it (#569), release v0.21.3

### LINKED ISSUE #569: BearPutSpread::get_loss_ranges panics in Positive::sub on zero strikes with fee > 0 (proptest finding)
## Summary

The property test `strategies_panic_freedom_test::test_vertical_spread_strategies_never_panic` (tests/property/strategies_panic_freedom_test.rs) found an input on which `BearPutSpread` panics instead of returning an error. Found by the Code Coverage run of the v0.21.2 release PR (#568), run 35422298895; the parallel run of the same workflow passed, so it only shows up when proptest draws this corner. It is not caused by the dependency update: positive 0.6.0 -> 0.6.1 has no source diff.

## Minimal failing input (from proptest)

```text
underlying  = 79228162514264337593543950335   (Decimal::MAX)
low_strike  = 0
high_strike = 0
volatility  = 0.0000000000000000000000000001
quantity    = 0.0000000000000000000000000001
premium     = 0
fee         = 1
rate        = 0
expiration  = Days(3650)
probe       = 0
```

## Panic

```text
thread 'strategies_panic_freedom_test::test_vertical_spread_strategies_never_panic' panicked at positive-0.6.1/src/positive.rs:280:5:
Positive invariant broken in sub: result would be non-positive
```

Frames inside the crate, innermost first:

- `optionstratlib::model::profit_range::ProfitLossRange::calculate_probability`
- `<BearPutSpread as optionstratlib::strategies::probabilities::core::ProbabilityAnalysis>::get_loss_ranges`

So the loss-range computation subtracts `Positive` values (strike minus per-contract cost, or similar) without checking that the result stays positive; with both strikes at zero and a non-zero fee the subtraction goes non-positive and `Positive::sub` aborts.

## Expected

`get_loss_ranges` / `calculate_probability` should return `Err(...)` (or an empty range) for degenerate inputs instead of panicking, consistent with #463 and the intent of the panic-freedom suite. Likely the same pattern exists in the other vertical spreads (`BullCallSpread`, `BearCallSpread`, `BullPutSpread`) and in `PoorMansCoveredCall`; worth checking all five once the arithmetic is made fallible (`checked_sub` / `try_sub` on `Positive`).

## Reproduce

```bash
PROPTEST_CASES=2000 cargo test --test property_tests test_vertical_spread_strategies_never_panic
```

or pin the input above in a unit test.

joaquinbejar:
Closes #569

## What the panic was

`ProfitLossRange::calculate_probability` (`src/model/profit_range.rs`) subtracted the two bound probabilities with the raw `Positive` operator. On the pinned proptest input it aborted with `Positive invariant broken in sub: result would be non-positive`.

The site is shared: every strategy's `get_profit_ranges` / `get_loss_ranges` delegates to it. Reproduced on the issue's input across the five strategies it names, only `BearPutSpread` builds a range that reaches it, but any strategy whose range bounds end up ordered this way would.

## What the input actually is (not what the issue guessed)

The issue reads it as "strike minus cost goes negative". Instrumenting the two probabilities on the pinned input:

```
lower_bound = MAX-4    upper_bound = None → MAX
p_below_lower = 0.998650101968426
p_below_upper = 0.5
```

The distribution function is being evaluated **higher at the lower point than at the upper one**. Not a zero-width range and not rounding noise; the difference is `-0.4987`. Cause: `(MAX - 4) / MAX` rounds to `0.9999999999999999999999999999` at `Decimal`'s 28 places, and `Decimal::checked_ln` returns `+9e-28` for it where the true value is `-1e-28`, so the lower bound lands three standard deviations above the spot while the upper bound sits at the median.

## The decision: report it, do not floor it

A distribution function is monotone. A smaller probability at the upper bound is a result outside the model's precision, and flooring `-0.4987` to zero would publish a probability nobody computed, which this codebase treats as worse than the abort it replaces. So:

```rust
self.probability = prob_below_upper
    .sub_or_none(&prob_below_lower.to_dec())
    .ok_or_else(|| ProbabilityError::CalculationError(
        ProbabilityCalculationErrorKind::InvalidProbability { value, reason }))?;
```

`sub_or_none` returns zero for equal operands, so a zero-width range still reports probability zero (a value, not an error), and only a genuine inversion errors. `InvalidProbability` is the existing variant documented for "a probability calculation produces a result that violates fundamental probability constraints"; no variant added, no signature changed.

**Ordinary inputs cannot reach the error.** The `ln` error is at the 28th decimal and only surfaces below `vol * sqrt(T) ≈ 1e-11`. A probe over 21,168 ordinary combinations (spot 0.5..1e5, vol 0.01..3, T 0.5d..3650d, three rates, bound gaps down to 1e-28) found 0 inversions and 9,555 exact equalities, all of which map to zero.

## One correction to the plan I ran

I first implemented this as a floor (`sub_floor_zero`) on the assumption the case was a zero-width range. It was not, and a zero result does not abort at all: `positive` is built without `non-zero`, so `1 - 1` succeeds with the raw operator. The floor would have hidden a `-0.5`. Instrumenting the values before committing is what caught it.

## Tests

- `bear_put_spread.rs::test_get_loss_ranges_zero_strikes_with_fee_does_not_panic`: the exact pinned input, asserting the `InvalidProbability` error with a negative reported difference. **Fails against the previous expression with the issue's exact panic message** (verified by reverting only the fix line).
- `profit_range.rs::test_calculate_probability_zero_width_range_is_zero`: a lower bound at `MAX` with an open upper bound is a zero-width range through the public constructor; pins that it reports zero and did not become an error. A non-regression guard, not a test that failed before.
- `test_vertical_spread_strategies_never_panic` at `PROPTEST_CASES=2000`: passes.

`# Errors` on `calculate_probability` now documents the new failure mode.

## Review

`panic-hunter` reviewed the diff and a probe crate against it: no residual panic path on the route (`sub_or_none`, `to_dec`, `Display` on `Positive`, `checked_*` throughout `calculate_single_point_probability`); the error-on-inversion decision confirmed defensible by the reachability probe above; the pinned test confirmed to abort on the old code.

It also found the same event **floored silently one function away**: `calculate_price_probability` (`utils.rs:252`) and `expected_value` (`core.rs:194`) use `sub_floor_zero` on the identical subtraction, and on this input return `(0.9987, 0, 0.5)`, a triple summing to `1.4987`. Out of scope for a one-issue fix; filed as **#570** with the measurement.

## Gates

```
cargo test --all-features --workspace       4067 + 83 + 423 + 210, 0 failed
LOGLEVEL=WARN cargo test  (default set)     4071 + 83 + 396 + 210, 0 failed
cargo fmt --all --check                     clean
cargo build --release                       warning-free
cargo doc --all-features --no-deps          clean
make pre-push                               zero warnings, README unchanged
make scan-banned                            OK
make public-api-check                       OK
cargo semver-checks --all-features          196 pass, no update required
```


joaquinbejar:
Second commit `7fa4691d` bumps the crate to **0.21.3** (`Cargo.toml`, the four `lib.rs` doc-header strings, `README.md` via `make readme`) and cuts the dated changelog section, so this PR is the release. `cargo publish --dry-run` packages and compiles the packaged crate clean. Merge, tag `v0.21.3`, publish and GitHub release follow once CI settles on this head.

Issue #545, opened 2026-08-30
M6-04: Move logger initialization to application code


joaquinbejar:
## 0.22.0 implementation policy

- Develop the new 0.22.0 architecture and API. Compatibility with 0.21.3 or earlier releases is not an acceptance requirement. Necessary API removals, renames, signature and serialization changes do not require individual historical-compatibility approvals; explain their purpose and update consumers and documentation.
- Keep one defining owner for each type and capability. Remove wrappers, aliases and adapters maintained only for old paths. Retain explicit facade/prelude exports where they improve the new API without duplicating types or introducing reverse dependencies.
- Preserve useful capabilities and numerical correctness. A source-breaking change does not justify silently dropping behavior. Explain any capability change and provide an executable replacement workflow where applicable.
- Review API changes against the exact effective base of each PR across the six relevant feature surfaces: none, default, plotly, static_export, async and all. Do not use old-release compatibility gates, AB registers or per-break approvals. Tool/build/parser failures remain failures, even when API differences are informational.
- Use 0.22.0 for the new workspace packages. No tag, release or crate publication during implementation. Publication issues execute only in the final release phase, after implementation and release preparation are complete and the owner initiates release. Preserve existing history, tags and releases.

## Mandatory engineering acceptance

- [ ] Monetary values use `Decimal`/validated domain newtypes. Use checked Decimal and arithmetic counter operations; never saturating/wrapping financial arithmetic. Document rounding. Restrict `f64` to bounded numeric internals with finite-value validation and controlled boundary conversion.
- [ ] Public boundaries use domain types and validated constructors. Errors are concrete `thiserror` types owned by the failing layer; no `anyhow`, production `unwrap`/`expect`, unchecked indexing, or submitted placeholder implementations.
- [ ] Use structured `tracing`; no print/debug logging or library-installed global subscriber. Gate async/I/O explicitly, keep defaults synchronous, and never block async execution without `spawn_blocking`. No unsafe concurrency or secret logging.
- [ ] Apply `#[must_use]` to pure functions and builders, `#[inline]` to small hot helpers and `#[cold]` to error paths; follow derive order `Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default` where meaningful. Use one concern per file, named constants, minimal intentional preludes, validated private fields and appropriate serde contracts.
[...]

Issue #506, opened 2026-08-30
M1-09: Relocate cross-layer constants and utilities


joaquinbejar:
## 0.22.0 implementation policy

- Develop the new 0.22.0 architecture and API. Compatibility with 0.21.3 or earlier releases is not an acceptance requirement. Necessary API removals, renames, signature and serialization changes do not require individual historical-compatibility approvals; explain their purpose and update consumers and documentation.
- Keep one defining owner for each type and capability. Remove wrappers, aliases and adapters maintained only for old paths. Retain explicit facade/prelude exports where they improve the new API without duplicating types or introducing reverse dependencies.
- Preserve useful capabilities and numerical correctness. A source-breaking change does not justify silently dropping behavior. Explain any capability change and provide an executable replacement workflow where applicable.
- Review API changes against the exact effective base of each PR across the six relevant feature surfaces: none, default, plotly, static_export, async and all. Do not use old-release compatibility gates, AB registers or per-break approvals. Tool/build/parser failures remain failures, even when API differences are informational.
- Use 0.22.0 for the new workspace packages. No tag, release or crate publication during implementation. Publication issues execute only in the final release phase, after implementation and release preparation are complete and the owner initiates release. Preserve existing history, tags and releases.

## Mandatory engineering acceptance

- [ ] Monetary values use `Decimal`/validated domain newtypes. Use checked Decimal and arithmetic counter operations; never saturating/wrapping financial arithmetic. Document rounding. Restrict `f64` to bounded numeric internals with finite-value validation and controlled boundary conversion.
- [ ] Public boundaries use domain types and validated constructors. Errors are concrete `thiserror` types owned by the failing layer; no `anyhow`, production `unwrap`/`expect`, unchecked indexing, or submitted placeholder implementations.
- [ ] Use structured `tracing`; no print/debug logging or library-installed global subscriber. Gate async/I/O explicitly, keep defaults synchronous, and never block async execution without `spawn_blocking`. No unsafe concurrency or secret logging.
- [ ] Apply `#[must_use]` to pure functions and builders, `#[inline]` to small hot helpers and `#[cold]` to error paths; follow derive order `Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default` where meaningful. Use one concern per file, named constants, minimal intentional preludes, validated private fields and appropriate serde contracts.
[...]

PR #576, opened 2026-09-19
Relocate DELTA_THRESHOLD to Greeks and deprecate the library logger (#506)

joaquinbejar:
## Summary

Milestone 1 of the multi-crate roadmap (M1-09). Two cross-layer utilities lose their reverse edges: `DELTA_THRESHOLD` moves down to the Greeks layer that consumes it, and the library logger is deprecated with every internal caller moved off it. The owner of every helper under `src/utils` is recorded in ADR-0001 D2; no shared utility crate is introduced.

## Changes

- `DELTA_THRESHOLD` moves from `strategies::delta_neutral` to `greeks` (used by `calculate_delta_neutral_sizes`); `strategies::delta_neutral::DELTA_THRESHOLD` and `strategies::DELTA_THRESHOLD` are downward re-exports. `rg 'crate::strategies' src/greeks` is empty.
- `utils::logger::setup_logger` and `setup_logger_with_level` are `#[deprecated(since = "0.22.0")]` with a note pointing at `tracing_subscriber::fmt().with_max_level(..).init()`. The `utils` and prelude re-exports stay (allowed at the re-export site) so no 0.21 path breaks.
- New non-published workspace member `examples/support` (`osl-example-support`, reserved `osl-example-` prefix per ADR-0001) provides `setup_logger`/`setup_logger_with_level` for application code; all 164 example binaries import it (explicit import shadows the prelude glob, so the deprecated function is never called by the examples).
- The logger's own tests (`tests/unit/utils/logger_test.rs`, `src/utils/logger.rs`) pin the deprecated behaviour under `#![allow(deprecated)]` until M6-04 deletes both functions.
- `src/utils/mod.rs` docs updated (deprecated section, application-side example).

## Technical decisions

Removing the logger now would fail the `semver` gate against the published 0.21.3 (and would break every 0.21 caller in a patch-level diff), so M1-09 delivers the non-breaking half: deprecation plus migration of internal callers. The physical removal and the `tracing-subscriber` dependency drop are M6-04 (#545), after the 0.22.0 bump (ADR-0001, "Sequencing under the semver gate"). Because of that, the acceptance criterion "library code no longer installs a global tracing subscriber" is not yet met and #506 stays open after this PR; the other four criteria are met here.

## Public API impact

- New public path `optionstratlib::greeks::DELTA_THRESHOLD`; canonical path change reflected in `public-api/optionstratlib.txt`.
- Two functions deprecated; nothing removed.
- `cargo semver-checks --all-features`: no semver update required.
- `CHANGELOG.md` entries under `[Unreleased]` (`### Deprecated`, `### Changed`).

## Testing

- [x] Existing tests unchanged and green; support crate has its own unit tests

## Checklist

- [x] All `pub` items have `///` documentation
- [x] Module boundaries: removes one forbidden edge, adds none
- [x] No new external dependencies (the support package depends on `tracing` and `tracing-subscriber`, both already in the workspace)

Part of #506 (the issue stays open until M6-04 removes the deprecated functions).


joaquinbejar (commented):
Independent review of `175a44265904b4d8745c49091448ed47028abdfe` against `80efdc02656b8df198a654b71a628ca292023ae5`. No blocking findings for this non-breaking stage. All 164 example changes are imports only, the API snapshot removes no entries, and the Greeks strategy import is gone. The logger removal remains explicitly deferred to #545, with #506 kept open. Local verification: `cargo test --offline -p osl-example-support` passed both tests with no warnings using isolated build directories and two jobs; `git diff --check` was clean. The first test attempt encountered an environment build-directory permission error, resolved by overriding that directory locally. I did not run the full workspace suite or wait for CI. This COMMENT is not a formal approval or merge authorization.

joaquinbejar (commented):
Verified the mechanics. DELTA_THRESHOLD now lives in greeks::utils with the same value and doc, strategies::delta_neutral re-exports it downward, src/greeks has no strategies import left, and the snapshot only gains the greeks and prelude paths with the two strategies paths intact. The allow(deprecated) sites are exactly the two re-exports plus the logger's own tests, all twelve example manifests pull osl-example-support, the 164 binaries that called the library logger on main are the 164 that now import the support one at file top, and the 13 that never called it are untouched. Two wording points below, nothing blocking. COMMENT.

CHANGELOG.md
```diff
@@ -7,6 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Deprecated
+
+- **`utils::logger::setup_logger` and `setup_logger_with_level`** (#506,
+  multi-crate roadmap M1-09). A library must not install a global `tracing`
+  subscriber; install one from your binary with
+  `tracing_subscriber::fmt().with_max_level(..).init()`. The functions still
+  work in 0.22.0 and are removed in a later 0.22 pre-release (M6-04). The
```
joaquinbejar: 💡 **Suggestion**: This sentence and the `#[deprecated]` note (src/utils/logger.rs:90 and :128) promise the functions still work in 0.22.0 and go away in a later 0.22 pre-release. ADR-0001 sequences M6-04 as part of the batch that follows the 0.22.0 bump PR, before the release ships, so a 0.21.3 user upgrading to 0.22.0 may never see the deprecation warning at all. Worth checking which one is the plan and stating it the same way in both places, otherwise the note reads as a compatibility promise the roadmap does not make.

examples/support/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn test_parse_level_maps_names_and_defaults_to_info() {
+        assert_eq!(parse_level("debug"), Level::DEBUG);
+        assert_eq!(parse_level("ERROR"), Level::ERROR);
+        assert_eq!(parse_level("Warn"), Level::WARN);
+        assert_eq!(parse_level("trace"), Level::TRACE);
+        assert_eq!(parse_level("nonsense"), Level::INFO);
+    }
+
+    #[test]
+    fn test_setup_is_idempotent() {
```
joaquinbejar, line 66: 🔵 **Nitpick**: This test asserts nothing, so it only proves the three calls do not panic. `tracing::dispatcher::has_been_set()` after the first call would make the idempotency claim observable. Fine to leave since this package is never published.

CHANGELOG.md
joaquinbejar: You are right, the two texts disagreed. ADR-0001 sequences M6-04 inside the batch after the bump, so 0.22.0 never ships these functions. In e76d7c80b484867361d7d974f520a7cb7cbf753e the attribute says since 0.21.3 and both the note and the changelog now say deprecated on main after 0.21.3, removed in 0.22.0, so the deprecation is only the signal for anyone building from main in between.

examples/support/src/lib.rs
joaquinbejar: Done in e76d7c80b484867361d7d974f520a7cb7cbf753e, the test now asserts tracing::dispatcher::has_been_set() after the first call and again after the two no-op calls.

joaquinbejar (commented):
Delta review of `175a44265904b4d8745c49091448ed47028abdfe..e76d7c80b484867361d7d974f520a7cb7cbf753e` completed. The deprecation notes and changelog now consistently describe removal before the 0.22.0 release, and the support test checks that a subscriber has been installed. No additional blocking findings. Both `osl-example-support` tests passed locally without warnings using isolated build directories, and `git diff --check` was clean. This COMMENT is not a formal approval or merge authorization.

Issue #537, opened 2026-08-30
M5-02 — Define the generic simulated-path evaluation interface


joaquinbejar:
Milestone: 0.22 Milestone 5 — Extract simulation and backtesting. Draft: `.issues/milestone-5-extract-simulation-backtesting/02-define-the-generic-simulated-path-evaluation-interface.md` (local planning material, not in the repository).

## Problem

The current simulation entry point returns backtesting-owned results and
describes strategy evaluation directly. That signature recreates a lower-to-
upper dependency even if the source files are moved.

## Evidence

- `src/simulation/traits.rs:1260-1293` defines `Simulate::simulate` in terms of
  `Simulator`, `ExitPolicy`, and backtesting-owned `SimulationStatsResult`.
- `src/simulation/traits.rs:7` imports that result from backtesting.
- `src/simulation/randomwalk.rs:304` and
  `src/simulation/simulator.rs:401` expose strategy-shaped profit evaluation.

## Required work

- Define the smallest generic path-evaluator contract needed by the engine,
  using associated types or generics without boxed dynamic dispatch on hot
  loops unless benchmark evidence approves it.
- Move per-run and aggregate generic result types into simulation.
- Implement strategy and backtest adapters in their upper-layer crates.
- Preserve exit-policy behavior and error context without importing concrete
  strategy or report types.
- After simulation exists, move the isolated chain/series generators behind
  `optionstratlib-market`'s optional `synthetic` dependency on simulation.

## Acceptance criteria

- [ ] Simulation evaluates a test callback/struct without strategies or
      backtesting in its manifest.
- [ ] Strategy and backtesting adapters reproduce existing results through the
      generic interface.
- [ ] The hot path adds no unapproved heap allocation or dynamic dispatch.
- [ ] Minimal market still excludes simulation; enabling `synthetic` adds the
      dependency and generator APIs explicitly.
- [ ] No dependency cycle is introduced between market and simulation.

## Verification commands

```sh
cargo test -p optionstratlib-simulation generic_evaluator
cargo tree -p optionstratlib-simulation -e normal
cargo check -p optionstratlib-market --no-default-features
cargo check -p optionstratlib-market --no-default-features --features synthetic
cargo tree -p optionstratlib-market -e features --no-default-features --features synthetic
```

## Dependencies

- M1-07 defines the logical callback and result boundary.
- M1-15 isolates current synthetic generators.
- M3-04 and M3-05 provide a minimal market package and feature policy.
[...]

Issue #525, opened 2026-08-30
M3-05 — Gate market-data I/O and async dependencies


joaquinbejar:
Milestone: 0.22 Milestone 3 — Extract pricing and market data. Draft: `.issues/milestone-3-extract-pricing-market/05-gate-market-data-i-o-and-async-dependencies.md` (local planning material, not in the repository).

## Problem

CSV, JSON, ZIP, filesystem, and async operations are mandatory dependencies of
the monolith even for consumers that only construct in-memory option chains.
The market crate needs additive features with a measurable minimal graph.

## Evidence

- `Cargo.toml:35-40,52-59,71-74` declares async, JSON, CSV, and ZIP
  dependencies; only the async runtime dependencies are currently optional.
- `src/chains/chain.rs:53,1328,1361,1413,1486,1517,1548` combines sync and
  async CSV/JSON persistence with the core chain type.
- `src/utils/csv.rs:15,171` combines ZIP handling with Tokio
  `spawn_blocking`.

## Required work

- Implement the exact market I/O placement and feature names approved by
  M0-06.
- Separate in-memory data/serde representation from filesystem CSV/JSON/ZIP
  operations and async wrappers.
- Make optional dependencies use `dep:` feature wiring and propagate only the
  minimum required lower-layer features.
- Add minimal, each-feature, and all-feature dependency snapshots and compile
  tests.
- Reserve simulation-backed generation for the M5 `synthetic` activation; do
  not add a temporary market-to-facade or market-to-monolith dependency.

## Acceptance criteria

- [ ] `--no-default-features` supports in-memory market types without Tokio,
      Reqwest, Futures, CSV, or ZIP in the resolved graph.
- [ ] Each I/O feature compiles independently and exposes only its documented
      operations.
- [ ] Async APIs are unavailable without the async feature and available with
      it.
- [ ] Feature combinations preserve serialization formats and round-trip
      fixtures.
- [ ] The dormant synthetic capability has an explicit M5 owner and no hidden
      mandatory simulation dependency.

## Verification commands

```sh
cargo check -p optionstratlib-market --no-default-features
cargo check -p optionstratlib-market --no-default-features --features io
cargo check -p optionstratlib-market --no-default-features --features async
cargo test -p optionstratlib-market --all-features
cargo tree -p optionstratlib-market -e features --no-default-features
```

## Dependencies

- M0-06 approves market I/O and synthetic placement.
- M3-04 creates the market package.

## Follow-up ownership

[...]

Issue #512, opened 2026-08-30
M1-15: Isolate simulation-backed market generators


joaquinbejar:
## 0.22.0 implementation policy

- Develop the new 0.22.0 architecture and API. Compatibility with 0.21.3 or earlier releases is not an acceptance requirement. Necessary API removals, renames, signature and serialization changes do not require individual historical-compatibility approvals; explain their purpose and update consumers and documentation.
- Keep one defining owner for each type and capability. Remove wrappers, aliases and adapters maintained only for old paths. Retain explicit facade/prelude exports where they improve the new API without duplicating types or introducing reverse dependencies.
- Preserve useful capabilities and numerical correctness. A source-breaking change does not justify silently dropping behavior. Explain any capability change and provide an executable replacement workflow where applicable.
- Review API changes against the exact effective base of each PR across the six relevant feature surfaces: none, default, plotly, static_export, async and all. Do not use old-release compatibility gates, AB registers or per-break approvals. Tool/build/parser failures remain failures, even when API differences are informational.
- Use 0.22.0 for the new workspace packages. No tag, release or crate publication during implementation. Publication issues execute only in the final release phase, after implementation and release preparation are complete and the owner initiates release. Preserve existing history, tags and releases.

## Mandatory engineering acceptance

- [ ] Monetary values use `Decimal`/validated domain newtypes. Use checked Decimal and arithmetic counter operations; never saturating/wrapping financial arithmetic. Document rounding. Restrict `f64` to bounded numeric internals with finite-value validation and controlled boundary conversion.
- [ ] Public boundaries use domain types and validated constructors. Errors are concrete `thiserror` types owned by the failing layer; no `anyhow`, production `unwrap`/`expect`, unchecked indexing, or submitted placeholder implementations.
- [ ] Use structured `tracing`; no print/debug logging or library-installed global subscriber. Gate async/I/O explicitly, keep defaults synchronous, and never block async execution without `spawn_blocking`. No unsafe concurrency or secret logging.
- [ ] Apply `#[must_use]` to pure functions and builders, `#[inline]` to small hot helpers and `#[cold]` to error paths; follow derive order `Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default` where meaningful. Use one concern per file, named constants, minimal intentional preludes, validated private fields and appropriate serde contracts.
[...]

PR #578, opened 2026-09-19
Gate the simulation-backed market generators behind synthetic (#512)

joaquinbejar:
## Summary

Milestone 1 of the multi-crate roadmap (M1-15). The chain and series generators build option chains on top of random walks; they are the only production edge from the market layer into `simulation`, and the approved DAG allows it only as an optional capability (`Market -> Simulation` behind `synthetic`, ADR-0003). This PR draws that boundary inside the monolith with a feature that is on by default, so nothing changes for a 0.21 user.

## Changes

- New feature `synthetic = []`, in `default`. It gates `chains::generators` and `series::generators`, the re-exports `chains::generator_optionchain`, `series::generator_optionseries`, the prelude entries and the deprecated `chains::generator_positive` alias.
- `default-features = false` now gives a market surface that never touches the simulation engine; `make test` builds it (`cargo build --no-default-features`) so the gate cannot rot.
- Doc links from `simulation::walk_driver` to the generators are plain text now (rustdoc stays clean without the feature; also removes an upward doc reference).
- `src/lib.rs` feature documentation lists `synthetic`; README regenerated.

## Technical decisions

`ChainError::Simulation(Box<SimulationError>)` and `From<SimulationError> for ChainError` stay as they are: ADR-0002 rule 4 forbids feature-gated enum variants, and the payload change (`SyntheticError`) is in the breaking batch behind the 0.22.0 bump (ADR-0003 section 4). The cargo-tree assertions that distinguish the minimal and synthetic market surfaces need real crates; they are owned by M3-05 (#525) and M5-02 (#537) with the fixtures `osl-fixture-market-minimal` / `osl-fixture-market-synthetic` from ADR-0002 section 5. That acceptance criterion of #512 therefore stays open until extraction; the other four are met here.

## Public API impact

- New Cargo feature `synthetic` (default-on). No item added, removed or moved.
- `cargo semver-checks --all-features`: no semver update required.
- `CHANGELOG.md` entry under `[Unreleased]` / `### Added`.

## Testing

- [x] Generator tests unchanged and green (they sit inside the gated modules and run with default features)
- [x] `cargo build --no-default-features` clean

## Checklist

- [x] All `pub` items have `///` documentation
- [x] Module boundaries: the market-to-simulation edge is now optional; no new edge
- [x] No new dependencies

Part of #512 (the cargo-tree fixture criterion is owned by #525 and #537 after extraction).


src/chains/mod.rs
```diff
@@ -229,14 +229,21 @@ mod rnd;
...
 mod generators;
 
+#[cfg(feature = "synthetic")]
 #[deprecated(
     since = "0.17.4",
     note = "moved to `optionstratlib::simulation::generator_positive`; it never depended on option chains"
 )]
 pub use crate::simulation::generator_positive;
 pub use chain::OptionChain;
+#[cfg(feature = "synthetic")]
```
joaquinbejar, line 246: 🟠 **Warning**: Gating this export leaves the benchmark target uncompilable without default features. `cargo check --offline -p optionstratlib --no-default-features --bench benches` passes on the base but fails here with E0432 at `benches/chains/generators.rs:4`, which still imports `generator_optionchain` unconditionally. The generator module and its registration in `benches/mod.rs` also remain unconditional. Gate the synthetic benchmark and its registration together, and include that target in the feature-disabled check. The new `cargo build --no-default-features` check does not compile benches and therefore misses this regression.

joaquinbejar (commented):
Reviewed `c4e11970b1bf2524756e0b04f8734fa86995df9a` against `80efdc02656b8df198a654b71a628ca292023ae5`. One build regression should be fixed before merge, reported inline. The benchmark check with default features disabled passes on the base and fails on this head with an unresolved gated generator import. Both checks used isolated build directories and two jobs. Generator bodies are unchanged and the export gates are otherwise consistent. I did not run the full suite or wait for CI. This COMMENT is not a formal approval or merge authorization.

Makefile
```diff
@@ -21,6 +21,7 @@ release:
 .PHONY: test
 test:
 	LOGLEVEL=WARN cargo test
+	cargo build --no-default-features
```
joaquinbejar, line 24: 🟠 **Warning**: No workflow invokes `make test`. build.yml runs `make build`, lint.yml runs `make lint` with `--all-features`, and code_coverage.yml runs tarpaulin with `--all-features`, so nothing in CI ever compiles the crate without `synthetic` and this line only fires on a local `make pre-push`. The PR body says the gate cannot rot, but it can. Add a `cargo build --no-default-features` step to build.yml (devops owns that file) so the minimal surface is checked on every PR. Note that with no `default-members` a root `cargo build` selects only the `optionstratlib` package, so the example crates do not re-enable the feature through unification, which is what makes this step meaningful.

src/lib.rs
```diff
@@ -821,6 +821,9 @@
 //! - `plotly`: Enables interactive visualization using plotly.rs
 //! - `static_export`: PNG / SVG export via `plotly_static` (pulls in async runtime)
 //! - `async`: Enables asynchronous I/O operations for OptionChain and OHLCV data (tokio + reqwest + futures)
+//! - `synthetic` (default): simulation-backed `OptionChain` and `OptionSeries` generators
```
joaquinbejar, line 824: 💡 **Suggestion**: CLAUDE.md under Key Decisions still lists the feature flags as `plotly`, `static_export`, `async`. ADR-0001 gives that rewrite to M1-10, so not blocking, but adding `synthetic` there now keeps the two lists from drifting.

src/simulation/walk_driver.rs
```diff
@@ -6,8 +6,8 @@
 //! Generic walk driver shared by every step generator.
 //!
 //! The three public generators (`generator_positive` here,
-//! [`crate::chains::generator_optionchain`] and
-//! [`crate::series::generator_optionseries`]) used to duplicate the same
+//! `chains::generator_optionchain` (behind the `synthetic` feature) and
+//! `series::generator_optionseries`) used to duplicate the same
```
joaquinbejar: 🔵 **Nitpick**: The parenthetical only says `chains::generator_optionchain` is behind `synthetic`, but `series::generator_optionseries` is gated the same way. Same wording at lines 190 and 464.

joaquinbejar (commented):
The gating itself is right. synthetic is additive and default-on, the cfg sits on the two private modules, the four re-exports and the prelude pair, no enum variant is gated, the deprecated generator_positive alias is the only remaining chains to simulation reference and it is gated too, the snapshot is unchanged, and the README comes from the lib.rs docs so the template did not need touching. The one thing I would fix before calling the gate enforced is that CI never runs make test, see the Makefile comment. COMMENT.

src/chains/mod.rs
joaquinbejar: Right, the bench target was left behind. In 81f230a4949a92b9e49e0a70f91380857f8484e5 benches/chains/generators.rs gates the import and the real benchmark on synthetic and registers a no-op without it, so the criterion_group list stays unconditional and the target compiles both ways.

Makefile
joaquinbejar: Agreed, make test alone was not a gate. 81f230a4949a92b9e49e0a70f91380857f8484e5 adds a cargo build --no-default-features step to build.yml right after make build, so every PR compiles the minimal surface on both Ubuntu images.

src/lib.rs
joaquinbejar: CLAUDE.md is not tracked in this repository, it exists only in the local checkout, so I updated the local copy's feature list directly; the tracked AGENTS.md rewrite belongs to M1-10.

src/simulation/walk_driver.rs
joaquinbejar: Reworded in 81f230a4949a92b9e49e0a70f91380857f8484e5 at all three sites so both generators are named as gated.

src/chains/mod.rs
joaquinbejar: Verified on `81f230a4949a92b9e49e0a70f91380857f8484e5`. The same benchmark check now passes both with default features and with `--no-default-features`, so the E0432 regression is fixed. The new build workflow step also checks the feature-disabled library build. I have left the conversation unresolved for the author.

Issue #511, opened 2026-08-30
M1-14: Partition errors by target crate


joaquinbejar:
## 0.22.0 implementation policy

- Develop the new 0.22.0 architecture and API. Compatibility with 0.21.3 or earlier releases is not an acceptance requirement. Necessary API removals, renames, signature and serialization changes do not require individual historical-compatibility approvals; explain their purpose and update consumers and documentation.
- Keep one defining owner for each type and capability. Remove wrappers, aliases and adapters maintained only for old paths. Retain explicit facade/prelude exports where they improve the new API without duplicating types or introducing reverse dependencies.
- Preserve useful capabilities and numerical correctness. A source-breaking change does not justify silently dropping behavior. Explain any capability change and provide an executable replacement workflow where applicable.
- Review API changes against the exact effective base of each PR across the six relevant feature surfaces: none, default, plotly, static_export, async and all. Do not use old-release compatibility gates, AB registers or per-break approvals. Tool/build/parser failures remain failures, even when API differences are informational.
- Use 0.22.0 for the new workspace packages. No tag, release or crate publication during implementation. Publication issues execute only in the final release phase, after implementation and release preparation are complete and the owner initiates release. Preserve existing history, tags and releases.

## Mandatory engineering acceptance

- [ ] Monetary values use `Decimal`/validated domain newtypes. Use checked Decimal and arithmetic counter operations; never saturating/wrapping financial arithmetic. Document rounding. Restrict `f64` to bounded numeric internals with finite-value validation and controlled boundary conversion.
- [ ] Public boundaries use domain types and validated constructors. Errors are concrete `thiserror` types owned by the failing layer; no `anyhow`, production `unwrap`/`expect`, unchecked indexing, or submitted placeholder implementations.
- [ ] Use structured `tracing`; no print/debug logging or library-installed global subscriber. Gate async/I/O explicitly, keep defaults synchronous, and never block async execution without `spawn_blocking`. No unsafe concurrency or secret logging.
- [ ] Apply `#[must_use]` to pure functions and builders, `#[inline]` to small hot helpers and `#[cold]` to error paths; follow derive order `Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default` where meaningful. Use one concern per file, named constants, minimal intentional preludes, validated private fields and appropriate serde contracts.
[...]

Issue #507, opened 2026-08-30
M1-10: Enforce and document dependency boundaries


joaquinbejar:
## 0.22.0 implementation policy

- Develop the new 0.22.0 architecture and API. Compatibility with 0.21.3 or earlier releases is not an acceptance requirement. Necessary API removals, renames, signature and serialization changes do not require individual historical-compatibility approvals; explain their purpose and update consumers and documentation.
- Keep one defining owner for each type and capability. Remove wrappers, aliases and adapters maintained only for old paths. Retain explicit facade/prelude exports where they improve the new API without duplicating types or introducing reverse dependencies.
- Preserve useful capabilities and numerical correctness. A source-breaking change does not justify silently dropping behavior. Explain any capability change and provide an executable replacement workflow where applicable.
- Review API changes against the exact effective base of each PR across the six relevant feature surfaces: none, default, plotly, static_export, async and all. Do not use old-release compatibility gates, AB registers or per-break approvals. Tool/build/parser failures remain failures, even when API differences are informational.
- Use 0.22.0 for the new workspace packages. No tag, release or crate publication during implementation. Publication issues execute only in the final release phase, after implementation and release preparation are complete and the owner initiates release. Preserve existing history, tags and releases.

## Mandatory engineering acceptance

- [ ] Monetary values use `Decimal`/validated domain newtypes. Use checked Decimal and arithmetic counter operations; never saturating/wrapping financial arithmetic. Document rounding. Restrict `f64` to bounded numeric internals with finite-value validation and controlled boundary conversion.
- [ ] Public boundaries use domain types and validated constructors. Errors are concrete `thiserror` types owned by the failing layer; no `anyhow`, production `unwrap`/`expect`, unchecked indexing, or submitted placeholder implementations.
- [ ] Use structured `tracing`; no print/debug logging or library-installed global subscriber. Gate async/I/O explicitly, keep defaults synchronous, and never block async execution without `spawn_blocking`. No unsafe concurrency or secret logging.
- [ ] Apply `#[must_use]` to pure functions and builders, `#[inline]` to small hot helpers and `#[cold]` to error paths; follow derive order `Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default` where meaningful. Use one concern per file, named constants, minimal intentional preludes, validated private fields and appropriate serde contracts.
[...]

PR #586, opened 2026-09-19
Partition the error module by target crate without changing any type (#511)

joaquinbejar:
## Summary

Milestone 1 of the multi-crate roadmap (M1-14). Each file under `src/error` now states its target crate (ADR-0001 D6) in its module docs, the `error` module docs carry the ownership table, and the nine `From` conversions whose source error is owned by a higher layer than the target moved next to their source, so no lower layer's file names a higher layer's error type through an impl.

## Changes

- Moved (bodies unchanged): `From<StrategyError> for PositionError/ProbabilityError/SimulationError` to `strategies.rs`; `From<ChainError> for SimulationError/VolatilityError` to `chains.rs`; `From<PricingError> for OptionsError` to `pricing.rs`; `From<MetricsError> for CurveError` to `metrics.rs`; `From<GraphError> for CurveError/SurfaceError` to `graph.rs`.
- `error::simulation` imports `GraphError` from `error::graph` instead of through the prelude.
- What stays, and why: the enum variants that hold a higher layer's error (`OptionsError::Greeks`, `CurveError::{MetricsError, Greeks, Graph}`, `SurfaceError::Greeks` and its graph variants, `VolatilityError::Chain`, `SimulationError::{Strategy, Chain, GraphError}`) are public shape; removing them fails the published-baseline semver gate, so they are in the batch behind the 0.22.0 bump and listed in the module docs. The 207 error tests pass as before.
- The acceptance criterion "lower-layer errors do not import higher-layer types or errors" is therefore met for impls and imports and deferred for the listed variants; the issue stays open until the batch lands.

## Public API impact

- Additive only: no removal, rename or signature change; every historical path is preserved. `public-api/optionstratlib.txt` updated in the branch; `cargo semver-checks --all-features`: no semver update required.
- `CHANGELOG.md` entry under `[Unreleased]`.

## Testing

- [x] Moved tests keep their assertions; new unit tests for the new contracts
- [x] `make pre-push` run locally and clean (zero warnings)
- [x] Full local gate list green on the head commit (`cargo test --all-features --workspace`, `LOGLEVEL=WARN cargo test`, `cargo build --release`, `RUSTDOCFLAGS=-D warnings cargo doc --all-features --no-deps`, `make scan-banned`, `make public-api-check`, `cargo semver-checks --all-features`)

## Checklist

- [x] Code follows `rules/global_rules.md` and `CLAUDE.md`
- [x] All `pub` items have `///` documentation; `# Errors` on fallible functions
[...]

PR #587, opened 2026-09-19
Enforce the module boundaries with make check-graph (#507)

joaquinbejar:
## Summary

Milestone 1 of the multi-crate roadmap (M1-10). `scripts/check_module_boundaries.py` scans production code for `crate::<module>` references (comments dropped, `#[cfg(test)]` items skipped by brace counting, `use crate::{...}` groups expanded entry by entry including nested and multi-line ones, `crate::error::<file>` paths resolved to the file's owner) and compares every edge with the approved layer graph (ADR-0001 D2, D6, D9).

## Changes

- Lines marked `// facade-compat: <layer>` are exempt (the compatibility re-exports that become facade code at extraction). The four known reverse edges whose removal is a breaking change wait in the `DEFERRED` list with the issue that removes them (`SimulationError::GraphError`, `StrategyError::Simulation`, `ProfitLossRange::new -> ProbabilityError`, `Strategable: Graph`); a deferred edge that disappears is reported so the list is pruned. The `synthetic`-gated market-to-simulation files are the one optional edge.
- `--self-test` builds throwaway trees and proves the scanner catches a forbidden edge, a qualified error path and a multi-line group, and ignores comments, test modules, marked lines and the synthetic files (the issue's out-of-scope note about line-oriented checks that ignore multi-line imports is addressed).
- `make check-graph` runs the self-test and the scan; `make check` includes it; the `lint` workflow runs it after `scan-banned`, so CI fails when a forbidden edge is introduced. On this chain it reports no forbidden edge with four deferred edges tolerated.
- The last two library imports through `crate::prelude` (`pricing::telegraph`, `strategies::delta_neutral::model`) use canonical paths; the `[Unreleased]` changelog section is consolidated into one heading per kind after the stacked restacks duplicated them.
- `AGENTS.md` and `CLAUDE.md` are not tracked in this repository; their local copies already describe the `analytics` module and the `synthetic` feature. Post-extraction, `check-graph` switches to the `cargo metadata` crate-graph assertion (ADR-0004 section 4).

## Public API impact

- Additive only: no removal, rename or signature change; every historical path is preserved. `public-api/optionstratlib.txt` updated in the branch; `cargo semver-checks --all-features`: no semver update required.
- `CHANGELOG.md` entry under `[Unreleased]`.

## Testing

- [x] Moved tests keep their assertions; new unit tests for the new contracts
- [x] Full local gate list green on the head commit (`cargo test --all-features --workspace`, `LOGLEVEL=WARN cargo test`, `cargo build --release`, `RUSTDOCFLAGS=-D warnings cargo doc --all-features --no-deps`, `make scan-banned`, `make public-api-check`, `cargo semver-checks --all-features`)

## Checklist

- [x] Checked arithmetic only; no boxed dynamic dispatch on hot paths
- [x] Module boundaries: removes the listed reverse edges; the marked `// facade-compat` lines are the recorded residue for the 0.22.0 batch
- [x] No new dependencies

Stacked on #511's PR. Part of #507 (the AGENTS.md/CLAUDE.md rewrite and the post-extraction crate-graph form remain).


scripts/check_module_boundaries.py
```diff
@@ -0,0 +1,352 @@
...
+    "series/generators.rs",
+    "chains/mod.rs",
+    # `ChainError::Simulation` and `From<SimulationError> for ChainError`
+    # (ADR-0003 section 4: the payload becomes market-owned after the bump).
+    "error/chains.rs",
+}
+
+# (source module, target module) -> issue that removes the edge.
+# Populated from the state of `main` after the M1 PRs; keep it sorted.
+DEFERRED: dict[tuple[str, str], str] = {
```
joaquinbejar: 🟠 **Warning**: `DEFERRED` is keyed by module pair, so it tolerates any file in the pair, not only the four lines that carry the debt. I checked on the branch, a new `src/strategies/foo.rs` with `use crate::visualization::Graph;` and a new `src/model/bar.rs` importing `ProbabilityError` both pass with exit 0. That defeats the gate on exactly the edges we are most likely to regress on. Store the file set with each entry, something like `("strategies", "visualization"): ({"strategies/base.rs"}, "0.22.0 batch ...")`, and in `violations_of` only skip the edge when `set(files) <= allowed_files`, the same rule `SYNTHETIC_FILES` already applies.

joaquinbejar (commented):
Ran the self test and the scan on a fresh worktree of the head, both green with four deferred edges, and the deferred pairs resolve to exactly `error/simulation.rs`, `error/strategies.rs`, `model/profit_range.rs` and `strategies/base.rs`. Comment stripping, the cfg(test) brace skip and the brace group expansion all behave on the real tree, the Makefile target and the lint step are wired the way ADR-0004 section 4 asks, and the changelog now has one heading per kind. Two things need fixing before this is a gate I would trust, the module level `DEFERRED` allowlist lets new edges through in the deferred pairs (I reproduced it) and the allowed table grants `math` to strategies and backtest against D9. Needs changes. COMMENT posted as requested by the lead.

scripts/check_module_boundaries.py
```diff
@@ -0,0 +1,352 @@
...
+
+# Layer -> layers it may reference (the approved DAG, ADR-0001 D9).
+ALLOWED = {
+    "core": {"core"},
+    "math": {"core", "math"},
+    "pricing": {"core", "math", "pricing"},
+    "simulation": {"core", "math", "pricing", "simulation"},
+    "market": {"core", "math", "pricing", "market"},
+    "analytics": {"core", "math", "pricing", "market", "analytics"},
+    "strategies": {"core", "math", "pricing", "market", "analytics", "strategies"},
```
joaquinbejar: 🟠 **Warning**: `strategies` and `backtest` may reach `math` here, but neither the roadmap graph (Strategies and Backtest have no Math edge) nor ADR-0001 D9 lists that edge, and D9 says explicitly that `strategies -> math` is not needed. Nothing on the branch uses it today, I checked with the scanner, so dropping `"math"` from both sets costs nothing and keeps the script on the same DAG as the ADR. If we do want the edge, D9 should be amended first, not the enforcement table.

scripts/check_module_boundaries.py
```diff
@@ -0,0 +1,352 @@
...
+    # the core constructor gets a core-owned error in the 0.22.0 batch
+    # (ADR-0001 D6, M1-01).
+    ("model", "error/probability"): "0.22.0 batch (ADR-0001 D6, #498)",
+    # `Strategable: ... + Graph` (`src/strategies/base.rs`): the
+    # visualization supertrait bound leaves the strategy contract in the
+    # 0.22.0 batch (M1-08, #505); dropping a supertrait is a public break.
+    ("strategies", "visualization"): "0.22.0 batch (#505, Strategable: Graph bound)",
+}
+
+MARKER = "// facade-compat:"
```
joaquinbejar, line 168: 💡 **Suggestion**: A marked line is dropped whatever it contains and the run prints nothing about how many there are (27 on this branch, several of them plain `use` lines rather than re-exports). ADR-0004 section 4 wants the debt visible, so printing a per layer count of marked lines next to the deferred count would make marker creep obvious in the CI log without failing anything. Same thought for `chains/mod.rs` in `SYNTHETIC_FILES`, which exempts the whole file rather than the `synthetic` gated lines.

scripts/check_module_boundaries.py
```diff
@@ -0,0 +1,352 @@
...
+    }
+    failures = 0
+    for name, ((rel, content), expected) in cases.items():
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp) / "src"
+            target = root / rel
+            target.parent.mkdir(parents=True)
+            target.write_text(content)
+            edges, _ = scan(root)
+            got = len([v for v in violations_of(edges) if (rel.split("/")[0], ) ])
```
joaquinbejar: 🔵 **Nitpick**: `if (rel.split("/")[0], )` is a non empty tuple, so the filter is always true and does nothing. While here, two cases are missing from the self test, a deferred pair yielding zero violations and a stale deferred entry being reported, and both are cheap to add with the same throwaway tree.

scripts/check_module_boundaries.py
```diff
@@ -0,0 +1,352 @@
...
+    i = 0
+    while i < len(lines):
+        line = lines[i]
+        if line.strip() == "#[cfg(test)]":
+            depth = 0
+            started = False
+            i += 1
+            while i < len(lines):
+                depth += lines[i].count("{") - lines[i].count("}")
+                if depth > 0:
```
joaquinbejar: 🟠 **Warning**: A test item whose braces balance on the same line never sets `started`. With `#[cfg(test)]\nmod tests {}\nuse crate::strategies::Strategy;` in `src/model/fixture.rs`, the inner loop consumes the following production import before stopping at its semicolon. I reproduced exit 0, while the same import without the empty test module exits 1. This lets a forbidden production dependency bypass the new CI gate. Track whether an opening brace was encountered independently of the net depth, and add a fixture with an empty test module followed by a forbidden import.

scripts/check_module_boundaries.py
```diff
@@ -0,0 +1,352 @@
...
+        if source_layer is None:
+            print(f"unknown module {source!r} in {rel}; add it to LAYER_OF or ERROR_FILE_LAYER", file=sys.stderr)
+            sys.exit(2)
+        text = "\n".join(production_lines(strip_comments(path.read_text())))
+        targets: list[str] = []
+        for match in EDGE_RE.finditer(text):
+            targets.append(normalise(match.group(1), match.group(2)))
+        # `use crate::{a::b, c, d::{e, f}}`, possibly spanning several lines:
+        # every entry's first segment names a module (or a root re-export).
+        for match in re.finditer(r"crate::\{", text):
```
joaquinbejar: 🟠 **Warning**: Group expansion only starts at `crate::{`, so a qualified group such as `use crate::error::{graph::GraphError};` is reduced to the bare `error` target and then exempted by `ALWAYS_ALLOWED_TARGETS`. In a model fixture, that form exits 0 while `use crate::error::graph::GraphError;` exits 1; the multiline form also bypasses the gate. These are the same forbidden dependency, with only the import spelling changed. Preserve the prefix before each brace group when expanding it, and cover both qualified single-line and multiline error groups in the self-test.

joaquinbejar (commented):
Reviewed `b72dcdd799f82dcd70b7eea0ab62e47533308bea` against logical parent #586 at `5f5c67accbf7f6e3f1f90fa7ef5c6d44f7c09418`. Two additional blocking findings are attached: both allow forbidden production dependencies to pass the new gate. They are distinct from the existing DEFERRED scope and allowed-DAG threads.

`make check-graph` passes on the submitted tree, but isolated CLI fixtures reproduce exit 0 for both bypasses and exit 1 for their ordinary-import controls. Experiments used temporary trees outside the PR worktree and did not modify production code. I did not run Rust builds for this scanner review. This COMMENT review is not a formal approval or authorization to merge.

scripts/check_module_boundaries.py
joaquinbejar: Done (f58a3cef, restacked): `DEFERRED` is now keyed by module pair with a `frozenset` of the exact files carrying the debt (11 entries), and a hit in any other file of the same pair is reported as a violation with the note "deferred only for <file>". Your two probes (`src/strategies/foo.rs` importing `visualization::Graph`, `src/model/bar.rs` importing `ProbabilityError`) now exit 1; both shapes are in the self-test ("deferred pair in its file" / "deferred pair in another file").

scripts/check_module_boundaries.py
joaquinbejar: Done: `math` is dropped from both `strategies` and `backtest` allowed sets, with the D9 reference in a comment. `make check-graph` still passes on the branch, so nothing used that edge.

scripts/check_module_boundaries.py
joaquinbejar: Done: the success line now prints the count of `facade-compat` lines per layer (`facade-compat lines per layer: market=1, math=3` on this branch) next to the deferred-edge count, and the marker is no longer a blanket exemption: only `pub use` re-exports may carry it, everything else that was marked became either a plain import (allowed edge) or a `// deferred edge` entry in `DEFERRED`. 27 marked lines went down to 4.

scripts/check_module_boundaries.py
joaquinbejar: Done: the no-op tuple filter is removed; the self-test has "deferred pair in its file" (expected 0) and a "stale deferred entries are detected" case that runs the scan on an empty tree and checks all 11 entries are reported as stale.

joaquinbejar (commented):
Reviewed f58a3cef4ae9fca78b04129eef6f7aaab4029a02 incrementally against the previous review, including the rebase, file-scoped deferred edges, D9 allowlist changes and marker reporting. The two existing blocking findings remain reproducible on this head: [empty test module consumes the following production import](https://github.com/joaquinbejar/OptionStratLib/pull/587#discussion_r4053319628) and [qualified grouped error import bypasses ownership checking](https://github.com/joaquinbejar/OptionStratLib/pull/587#discussion_r4053319633). No duplicate inline threads were added.

The original CLI probes still return 0 for both bypasses, while the equivalent direct-import controls return 1. make check-graph passes on the repository with eleven deferred edges, so its current self-tests do not catch these cases. No Rust build or full suite was repeated for this Python/configuration delta. This COMMENT is not formal approval or merge authorization.

scripts/check_module_boundaries.py
joaquinbejar: Confirmed and fixed in 70f3667b: any `{` on a line now marks the test item as started before the depth is updated, so `mod tests {}` starts and ends on that line and the following `use crate::strategies::Strategy;` is scanned (exit 1). Self-test cases "empty test module then import" and "test module file then import" (`mod tests;`) cover both shapes.

scripts/check_module_boundaries.py
joaquinbejar: Confirmed and fixed in 70f3667b: the group scan now matches `crate::((\w+::)*)\{` and prefixes every expanded entry with the segments before the brace, so `use crate::error::{graph::GraphError};` yields `error/graph` (exit 1 in a model fixture), and the multiline form too. Self-test cases "qualified group" and "multiline qualified group". While there, the `facade-compat` marker now exempts only `pub use` lines; a marked plain import is scanned like any other (self-test "marked plain import is not exempt").

scripts/check_module_boundaries.py
joaquinbejar: Verified on 70f3667bee3679e74c3e92bf31e3410734b4354a: the original empty-test-module fixture now exits 1 and reports the following model -> strategies import. The direct-import control still exits 1, and the self-tests cover both an empty braced module and a mod tests; declaration. This addresses the reported cause.

scripts/check_module_boundaries.py
joaquinbejar: Verified on 70f3667bee3679e74c3e92bf31e3410734b4354a: the original multiline use crate::error::{graph::GraphError} fixture now exits 1 and reports the core -> visualization edge, just like the fully qualified control. Prefixing the expanded group entries addresses this bypass; the qualified-group self-tests also pass.

joaquinbejar (commented):
Reviewed the delta at 70f3667bee3679e74c3e92bf31e3410734b4354a. Both previously reported blocking findings are now verified corrected; confirmation is in their existing threads. No new blocking findings identified.

The four original CLI probes now all reject their forbidden edges with exit status 1, all 17 scanner self-tests pass, and make check-graph passes on the repository. Also inspected the restriction of facade-compat exemptions to pub use lines. The four earlier patches are unchanged after rebasing onto c2f43972450b4360c4c12719ecfa5e69a13235b8. No Rust build or full suite was repeated for this Python delta. This COMMENT is not formal approval or merge authorization.

PR #589, opened 2026-09-19
Make every Decimal square root total (#588)

### LINKED ISSUE #588: Decimal square roots can hit rust_decimal's circuit-breaker panic (model/decimal, volatility, simulation)
## Overview

`rust_decimal::MathematicalOps::sqrt` aborts the process with `geo mean circuit breaker` (an `assert!` in `rust_decimal-1.43.0/src/maths.rs:320`) when its Newton iteration oscillates at the last digit instead of converging. Every square root in the library goes through it, either via the crate helper `model::decimal::d_sqrt` or via `Positive::checked_sqrt` from the `positive` crate (28 production call sites), so a reachable panic sits on pricing, Greeks, simulation, RND and volatility paths. The property suite hit it in `garch_volatility`.

## Context

- Reproducer (rust_decimal 1.43.0): `dec!(4.0000000000000000000000000003).sqrt()` panics; `dec!(4.0000000000000000000000000001)` and `dec!(5.0000000000000000000000000004)` return normally. Any value one or a few units in the 28th decimal above a perfect square can trigger it.
- Failing property case: `garch_volatility(&[1, 1, 1, 1], omega = 1, alpha = 1e-28, beta = 1)` (`tests/property/volatility_panic_freedom_test.rs::test_garch_volatility_never_panics`): the variance sequence reaches `4.0000000000000000000000000003` on the third step. The test is unseeded, so it fails on roughly a third of the runs and blocked two gate runs of the M1 chain on 2026-09-19.
- `Positive::checked_sqrt` (`positive` 0.6.1, `src/positive.rs:1268`) is "checked" only for negative inputs; it forwards to the same panicking `sqrt`. The upstream fix belongs to the `positive` crate; this issue covers what this library can do now.
- `rules/global_rules.md`: zero panics in production code; `make scan-banned` already bans direct `.sqrt()` calls for this reason.

## Tasks

- [ ] Replace the body of `model::decimal::d_sqrt` with a total Newton iteration: same initial guess and update as upstream so every converging input yields the bit-identical result, plus a bounded loop that returns the better of the two oscillating candidates instead of asserting.
- [ ] Route the 28 production `Positive::checked_sqrt()` call sites through a crate-owned helper built on `d_sqrt` (keeping each site's error type), so no production path reaches the upstream `sqrt`.
- [ ] Add the exact reproducer as a unit test on `d_sqrt` and pin the GARCH case as a deterministic regression test beside the property test.
- [ ] Extend `make scan-banned` (or a unit test) so a new `checked_sqrt()` call in production code is flagged.
- [ ] Open the upstream fix on the `positive` crate (same owner) and record the version that carries it.

## Acceptance criteria

- `dec!(4.0000000000000000000000000003)` through `d_sqrt` returns `Ok(2.0000000000000000000000000000)` (or the nearest 28-digit value) and every previously converging input returns exactly what `Decimal::sqrt` returned.
- `PROPTEST_CASES=4000 cargo test --test property_tests test_garch_volatility_never_panics` passes.
- `rg -n 'checked_sqrt\(' src` outside `#[cfg(test)]` is empty.
- No numerical change on any input that did not panic.

## Out of scope

- Changing volatility, GARCH or RND formulas.
- The `positive` crate release itself (tracked from that repository).

joaquinbejar:
## Summary

`rust_decimal::MathematicalOps::sqrt` aborts the process with `geo mean circuit breaker` when its Newton iteration oscillates between two adjacent 28-digit values instead of converging (reproducer on 1.43.0: `dec!(4.0000000000000000000000000003).sqrt()`). Every square root in the library went through it, directly or via `Positive::checked_sqrt`, so a reachable panic sat on pricing, Greeks, simulation, RND and volatility paths. The property suite found it in `garch_volatility(&[1, 1, 1, 1], 1, 1e-28, 1)` and blocked two gate runs of the M1 chain.

## Changes

- `model::decimal::d_sqrt` owns the iteration: same seed and update step as upstream, so every input on which upstream converges yields the bit-identical result, plus a bounded loop that returns the oscillating candidate whose square is closest to the input instead of asserting. Negative input stays an error; intermediate overflows are typed errors instead of arithmetic panics.
- `model::decimal::p_sqrt` wraps it for `Positive` with the `PositiveError` type the 28 former `Positive::checked_sqrt` call sites already mapped, so those sites change only their call syntax.
- The 39 remaining `Decimal::sqrt` call sites call `d_sqrt` with their previous `unwrap_or` / `ok_or_else` / `match` semantics unchanged; the 14 `f64::sqrt` sites carry a `scan-banned: allow` marker (f64 returns NaN, it does not abort).
- `make scan-banned` now flags any `.sqrt()` / `.checked_sqrt()` in production code, so the upstream path cannot come back unnoticed.
- Tests: the exact reproducer and an upstream-equality check on `d_sqrt`; a deterministic GARCH regression test under `tests/unit/volatility`; `PROPTEST_CASES=4000 cargo test --test property_tests test_garch_volatility_never_panics` passes.

## Technical decisions

The fix is crate-local on purpose: the upstream fix belongs to `rust_decimal` and to the `positive` crate (same owner; `checked_sqrt` there still forwards to the panicking `sqrt`), and this library must not panic in the meantime. No formula, rounding or converging value changes; the only behavioural change is that inputs which used to abort the process now return the correct root (or, for a negative input at a formerly panicking site, the same fallback the site already used for `None`).

## Public API impact

- None (`d_sqrt` and `p_sqrt` are `pub(crate)`). `CHANGELOG.md` entry under `### Fixed`.

## Testing

- [x] Unit tests for the reproducer and upstream equality; deterministic GARCH regression; property suite at 4000 cases
- [x] Full local gate list green (`cargo test --all-features --workspace`, `LOGLEVEL=WARN cargo test`, `cargo build --release`, `RUSTDOCFLAGS=-D warnings cargo doc --all-features --no-deps`, `make scan-banned`, `make public-api-check`, `cargo semver-checks --all-features`)

## Checklist

- [x] Checked arithmetic only
- [x] No new dependencies

Closes #588


joaquinbejar (commented):
Reviewed dfdcdd975ced48e4a2bd263630bc9f1f47134c99 against its merge base eac33c4ff3bf90f21ecb2991eb957f62a36918d7. No blocking findings identified. This is a code review, not formal approval or merge authorization.

Checked the complete incremental diff, the replacement Newton iteration against the installed rust_decimal 1.43.0 source, Positive 0.6.1 checked_sqrt semantics, and the converted callers and error mappings. Locally, all three focused d_sqrt tests, the deterministic GARCH regression, and make scan-banned pass without warnings. The original upstream convergence algorithm is retained for previously convergent inputs.

Limitations: I did not rerun the full suite, the 4000-case property run, or CI. The branch predates the capability moves already on main; subsequent restacking should retain these conversions at their new locations.

src/model/decimal.rs
```diff
@@ -559,19 +560,86 @@ pub(crate) fn d_powd(
...
+    // half is not representable.
+    let mut result = x.checked_div(Decimal::TWO).ok_or_else(overflow)?;
+    if result.is_zero() {
+        result = x;
+    }
+    let mut last = result.checked_add(Decimal::ONE).ok_or_else(overflow)?;
+    let mut iterations = 0u32;
+    while last != result {
+        iterations += 1;
+        if iterations > SQRT_MAX_ITERATIONS {
```
joaquinbejar: 💡 **Suggestion**: The oscillation is a period 2 cycle and it shows up at iteration 3 for the reproducer, so waiting for 1000 iterations before resolving it costs about 71µs per call against 97ns for a converging neighbour like 4.0000000000000000000000000004 (measured in release on the PR body). In a GARCH or Heston loop that lands on such a variance at every step that is a 700x slowdown on the path you just made total.

Keeping a `before_last` and breaking into the same closest-square pick when `result == before_last` resolves it on the spot, with the 1000 bound left as the backstop for anything that is not a 2 cycle. While there, if `candidate * candidate` overflows in `error_of` (only possible for a candidate a hair above sqrt(MAX)) it may be friendlier to treat that as infinite error and pick the other candidate rather than fail the whole root.

src/model/decimal.rs
```diff
@@ -720,6 +788,53 @@ pub mod tests {
...
+            Decimal::ONE,
+            Decimal::TWO,
+            dec!(4),
+            dec!(0.25),
+            dec!(0.0000000000000000000000000001),
+            dec!(123456.789),
+            Decimal::MAX,
+        ];
+        for x in inputs {
+            let expected = x.sqrt().unwrap_or_default();
```
joaquinbejar, line 824: 💡 **Suggestion**: Eight inputs is thin for a bit identical claim, and `assert_eq!` on `Decimal` compares numerically so it would not catch a scale difference. Worth adding a seeded sweep (a `StdRng` from a fixed seed, a few thousand random mantissa and scale pairs, plus squares of short roots plus or minus a few units in the 28th decimal) and comparing `serialize()` bytes. It runs in a few milliseconds and it is the actual guarantee the CHANGELOG makes. For the record my scratch run over about 1.1M such inputs found zero divergences, so it will pass today.

src/model/decimal.rs
```diff
@@ -559,19 +560,86 @@ pub(crate) fn d_powd(
...
+/// no production path reaches the panicking upstream `sqrt` that
+/// `Positive::checked_sqrt` still wraps (#588).
+///
+/// # Errors
+///
+/// Returns [`PositiveError::ArithmeticError`] when the iteration overflows
+/// or the root cannot be represented as a `Positive`; the error type matches
+/// `Positive::checked_sqrt` so call sites keep their conversions.
+#[inline]
+pub(crate) fn p_sqrt(x: &Positive, op: &'static str) -> Result<Positive, PositiveError> {
```
joaquinbejar, line 642: 🔵 **Nitpick**: `p_sqrt` takes an `op` but then rewraps under the fixed op `"sqrt"`, so the call site name only survives inside the reason string, and several sites pass the same generic name anyway (`"volatility::utils::sqrt"` for six different functions, `"simulation::traits::sqrt"` for five). Either pass `op` through to `PositiveError::arithmetic_error(op, ...)` or give those sites the distinct names the `d_sqrt` sites got, otherwise the error will not say which function it came from.

src/volatility/utils.rs
```diff
@@ -400,7 +400,7 @@ pub fn garch_volatility(
         "volatility::garch::initial_variance",
     )?;
     let mut variance = to_positive_variance(seed, "initial_variance")?;
-    let mut volatilities = vec![variance.checked_sqrt()?];
+    let mut volatilities = vec![p_sqrt(&variance, "volatility::utils::sqrt")?];
```
joaquinbejar: 🔵 **Nitpick**: The `# Errors` section above (the running `checked_sqrt` line) still names `checked_sqrt`, which the body no longer calls. Same stale mention in the comment above `p_sqrt` in `greeks/equations.rs` (the checked counterpart line).

joaquinbejar (commented):
Ran the PR's d_sqrt body against upstream rust_decimal 1.43 sqrt in a scratch crate on about 1.1M inputs (perfect squares, squares plus or minus 1 to 12 units in the 28th decimal at many scales, powers of ten from 1e-28 to 1e27, Decimal::MAX and its neighbourhood, values around 1, seeded random mantissa and scale pairs). Every input where upstream converges gives a byte identical result via serialize, zero divergences, zero errors from the PR, max 54 iterations on converging inputs, and the only oscillator found in the whole set is the 4.0000000000000000000000000003 reproducer, which resolves to 2.0000000000000000000000000001, the better of the two neighbours. Negative inputs and negative zero error out where upstream returns None, the overflow reasoning on x / result holds since the iterate stays at or above the root after the first step, and every rerouted call site keeps its previous unwrap_or, map_err or match shape, with the 15 f64 markers all on genuine f64 receivers (the PR text says 14). The GARCH regression really does reach 4 plus 3 units on the third step. Two small suggestions inline, one on resolving the cycle early instead of after 1000 iterations, one on widening the equality test, plus two nits. COMMENT.

src/model/decimal.rs
joaquinbejar: Good measurement. 7da531f5 keeps a before_last and resolves the moment result equals the value two steps back, with the 1000 bound as the backstop; a candidate whose square overflows now counts as infinitely far so the other one wins. A timing test pins the immediate detection.

src/model/decimal.rs
joaquinbejar: Added in 7da531f5: a seeded StdRng sweep of 4000 mantissa and scale pairs plus squares of short roots nudged by up to six units in the 28th decimal, compared on serialize() bytes against upstream where it converges, and asserting d_sqrt is Ok where upstream aborts. Thanks for the 1.1M run, that is the guarantee the changelog leans on.

src/model/decimal.rs
joaquinbejar: Both done in 7da531f5, p_sqrt passes op through to PositiveError::arithmetic_error and every call site now names its enclosing function.

src/volatility/utils.rs
joaquinbejar: Fixed both mentions in 7da531f5.

joaquinbejar (commented):
Reviewed the delta from dfdcdd975ced48e4a2bd263630bc9f1f47134c99 to 7da531f5e02b1ef7205d751acbdaddb2653c1d32. No additional blocking findings. Checked the immediate cycle detection, candidate selection, operation identifiers and new regression coverage. All five focused d_sqrt tests pass locally without warnings, including the seeded byte-for-byte comparison against upstream. No full-suite or CI validation was performed for this delta; this COMMENT is not formal approval or merge authorization.

joaquinbejar (commented):
Reviewed the restack at a4cc5f71b5d2fed0bf99b5f96b80eb565e660e88 against 0fec5443d0d89bcbf7555e0233ddc11da221f280. No additional blocking findings. The previously reviewed root algorithm is preserved, and the four removed strategy call sites are now covered by the conversion in PathStatistics::from_outcomes.

Local checks pass without warnings: the single-leg golden regression (140 combinations), the eight simulation::outcome tests, and make scan-banned. Full-suite and CI validation were not repeated. This COMMENT is not formal approval or merge authorization.

Issue #508, opened 2026-08-30
M1-11 — Break the pricing-to-simulation cycle


joaquinbejar:
Milestone: 0.22 Milestone 1 — Normalize dependencies inside the monolith. Draft: `.issues/milestone-1-normalize-dependencies/11-break-pricing-to-simulation-cycle.md` (local planning material, not in the repository).

## Problem

PricingEngine stores a concrete Simulator while simulation depends on pricing, creating a direct crate cycle in the proposed extraction.

## Evidence

- `src/pricing/unified.rs:6` imports `simulation::simulator::Simulator`.
- `PricingEngine::MonteCarlo` stores `Simulator<Positive, Positive>`.
- `src/simulation/simulator.rs:9-10` depends on pricing payoff and Monte Carlo functions.
- `src/greeks/numerical.rs:15` consumes the unified PricingEngine API.

## Required work

- Define a generic pricing or path-provider contract owned at the lowest valid layer.
- Make simulation provide the concrete Monte Carlo implementation.
- Move the cross-capability dispatcher to the facade or approved integration layer.
- Refactor numerical Greeks to accept the generic pricing abstraction rather than a concrete cross-layer enum.
- Avoid boxed dynamic dispatch on hot paths unless benchmarks justify it.
- Preserve facade compatibility or document the 0.22 migration.

## Acceptance criteria

- [ ] Pricing contains no production import of simulation.
- [ ] Simulation may depend on pricing without a reverse edge.
- [ ] Closed-form and Monte Carlo pricing remain accessible through documented APIs.
- [ ] Numerical Greek tests work with the new generic abstraction.
- [ ] Pricing and Monte Carlo regression outputs remain unchanged.
- [ ] Any dynamic dispatch introduced on a hot path has benchmark evidence.

## Verification commands

```sh
rg -n --glob '*.rs' 'crate::simulation' src/pricing
```

## Dependencies

- M0-04 resolves dispatcher ownership.
- M1-07 defines generic simulation boundaries.

## Out of scope

- Changing Monte Carlo algorithms.
- Keeping a concrete Simulator field behind a type alias.

Issue #505, opened 2026-08-30
M1-08 — Move simulation and backtesting orchestration out of strategies


joaquinbejar:
Milestone: 0.22 Milestone 1 — Normalize dependencies inside the monolith. Draft: `.issues/milestone-1-normalize-dependencies/08-move-simulation-and-backtesting-orchestration-out-of-strategies.md` (local planning material, not in the repository).

## Problem

Concrete strategy modules import simulation and backtesting result types, reversing the intended dependency where backtesting composes strategies and simulation.

## Evidence

- `src/strategies/long_call.rs`, `long_put.rs`, `short_call.rs`, and `short_put.rs` import backtesting result types and simulation paths.
- `src/strategies/graph.rs:15` imports visualization.
- The target graph requires Backtest -> Strategies and Visualization -> Strategies.

## Required work

- Move strategy-specific simulation entry points and result construction into backtesting adapters.
- Keep pure strategy payoff, validation, optimization, and position composition in strategies.
- Move strategy chart construction into visualization.
- Preserve facade convenience functions through upper-layer wrappers where approved.
- Add compile fixtures proving strategies compile without simulation, backtesting, or visualization.

## Acceptance criteria

- [ ] Strategies has no production dependency on simulation, backtesting, or visualization.
- [ ] Backtesting still runs all previously supported strategy simulations.
- [ ] Strategy-only consumers compile without orchestration and chart dependencies.
- [ ] Regression outputs for long/short call and put simulations remain unchanged.

## Verification commands

```sh
rg -n --glob '*.rs' 'crate::(simulation|backtesting|visualization)' src/strategies
```

## Dependencies

- M1-07 generic simulation interface.

## Follow-up ownership

- M6-02 visualization extraction.

## Out of scope

- Removing supported backtests rather than relocating them.
- Changing strategy P&L behavior.

joaquinbejar:
Recording that this issue's acceptance criteria are not met on `main`, without reopening it.

`python3 scripts/check_module_boundaries.py --inventory` on `c3f2d0de` still lists two edges naming #505 as their owner:

| Edge | File |
|---|---|
| `strategies -> simulation` | `strategies/simulation_impls.rs` (`impl BasicAble` for `Simulator` and `RandomWalk`) |
| `strategies -> visualization` | `strategies/base.rs` (the `Strategable: Graph` supertrait bound) |

[...]

Issue #504, opened 2026-08-30
M1-07 — Decouple simulation from strategies and backtesting


joaquinbejar:
Milestone: 0.22 Milestone 1 — Normalize dependencies inside the monolith. Draft: `.issues/milestone-1-normalize-dependencies/07-decouple-simulation-from-strategies-and-backtesting.md` (local planning material, not in the repository).

## Problem

Simulation imports concrete strategy behavior and backtesting result types, preventing it from becoming an independent lower-level crate.

## Evidence

- `src/simulation/randomwalk.rs:17` and `src/simulation/simulator.rs:14` import `strategies::base::BasicAble`.
- `src/simulation/randomwalk.rs:19` and `src/simulation/simulator.rs:16` import visualization-owned `Graph` and `GraphConfig` contracts.
- `src/simulation/stats.rs:6` imports `backtesting::results::SimulationResult`.
- `src/simulation/traits.rs:7` imports `SimulationStatsResult` from backtesting.

## Required work

- Define the minimal callback or trait required to evaluate a generated path.
- Move generic simulation result/stat types into simulation.
- Move strategy adaptation and backtest reporting into backtesting.
- Remove visualization behavior from simulation or route it through the later adapter boundary.
- Preserve deterministic random seeds and statistical outputs.

## Acceptance criteria

- [ ] Simulation has no production dependency on strategies, backtesting, market, or visualization.
- [ ] A simulation-only fixture can generate and summarize paths.
- [ ] Backtesting adapts concrete strategies to the generic simulation interface.
- [ ] Deterministic simulation regression results remain unchanged.

## Verification commands

```sh
rg -n --glob '*.rs' 'crate::(strategies|backtesting|chains|series|visualization)' src/simulation
```

## Dependencies

- M0-04 ownership ADR.
- M1-08 moves orchestration upward.
- M1-11 defines pricing/simulation interaction.

## Out of scope

- Changing stochastic processes or statistical formulas.

Issue #502, opened 2026-08-30
M1-05 — Separate generic math from pricing and visualization


joaquinbejar:
Milestone: 0.22 Milestone 1 — Normalize dependencies inside the monolith. Draft: `.issues/milestone-1-normalize-dependencies/05-separate-generic-math-from-pricing-and-visualization.md` (local planning material, not in the repository).

## Problem

The proposed math layer currently contains option-specific curve and surface traits that import Greeks, plus graph implementations that import visualization. This creates Math-to-Pricing and Math-to-Visualization reverse edges.

## Evidence

- `src/curves/basic.rs:8` imports Greeks and its default methods price Options.
- `src/surfaces/basic.rs:7` imports Greeks and prices Options across volatility inputs.
- `src/curves/curve.rs:19`, `src/surfaces/surface.rs:38`, and geometry plotters import visualization.
- `BasicCurves` and `BasicSurfaces` are implemented for OptionChain, while Curve, Surface, Point2D, and Point3D are generic containers.

## Required work

- Keep generic geometry, interpolation, curves, surfaces, and points in math.
- Move `BasicCurves`, `BasicSurfaces`, and option-specific projection logic to analytics.
- Move graph implementations and plotters to visualization.
- Split error variants so generic math errors do not depend on pricing or visualization.
- Preserve facade paths where possible.

## Acceptance criteria

- [ ] Math has no production dependency on Greeks, pricing, market, analytics, or visualization.
- [ ] Analytics owns option-specific curve and surface projections.
- [ ] Generic Curve and Surface construction/interpolation tests pass in a math-only surface.
- [ ] OptionChain projection and visualization tests continue to pass in their owning layers.

## Verification commands

```sh
rg -n --glob '*.rs' 'crate::(greeks|pricing|chains|metrics|visualization)' src/curves src/surfaces src/geometrics
```

## Dependencies

- M0-04 ownership ADR.
- M1-12 moves metric implementations to analytics.

## Follow-up ownership

- M6-02 later extracts visualization.

## Out of scope

- Moving generic Curve or Surface containers into analytics.
- Changing interpolation algorithms.

PR #591, opened 2026-09-20
Resolve error types to their owning file in the boundary checker (#590)

### LINKED ISSUE #590: M1-10b — Boundary checker resolves error types through their bindings
Milestone: 0.22 Milestone 1 — Normalize dependencies inside the monolith. Follow-up of #507 (M1-10); local planning material: `doc/M1-EXIT-PROPOSAL.md` §1b (not in the repository).

## Problem

`scripts/check_module_boundaries.py` classifies a bare `crate::error::Name` import as a reference to the facade-level `error` module (`ALWAYS_ALLOWED_TARGETS`), so a lower layer that names a higher layer's error type in a signature is never reported. The "11 deferred edges" it prints today are only the edges written with a file-qualified path or a non-error module.

## Evidence

A one-off resolver over production lines (comments and `#[cfg(test)]` excluded) finds twelve edge groups the checker does not see, among them:

- core → pricing through `PricingError` in `src/model/leg/{traits,leg_enum,spot,future,perpetual}.rs` and `src/model/position.rs` (`LegAble::pnl_at_price`, `Position::pnl_at_expiration`)
- core → market through `ChainError` in `src/model/utils.rs` (`calculate_optimal_price_range`)
- math → analytics through `MetricsError` in `src/curves/curve.rs`, `src/surfaces/surface.rs`, `src/geometrics/analysis/traits.rs`
- math → market through `ChainError` in `src/geometrics/construction/types.rs` (`ResultPoint`)
- pricing → simulation through `SimulationError` in `src/volatility/utils.rs` (`generate_ou_process`)
- simulation → market through `ChainError` in `src/simulation/walk_driver.rs`
- inside `src/error/`: bare imports such as `use crate::error::{ChainError, StrategyError}` in `simulation.rs` (the #511 variant payloads) are invisible for the same reason

The same one-off resolver also produced false positives (`From<Error>` where `Error` is `std::io::Error`), which is why name resolution must follow the file's `use` bindings, not the last path segment.

## Required work

- Map every error type to the `src/error/*.rs` file that defines it and classify each reference by that file's layer (`ERROR_FILE_LAYER`), for bare, grouped, nested, aliased (`as`), file-qualified and expression-position paths.
- Resolve names through the file's `use` bindings: a name bound from `std::io::Error` or `csv::Error` is foreign; a name bound from `super::`/`self::` is resolved through the module path and the crate's `pub use` re-export graph; a glob (`crate::error::*`, `crate::prelude::*`) scans the file for the error names it brings in.
- Detect a type alias that wraps an error type at its definition site.
- Record every newly surfaced reverse edge in `DEFERRED` (file-scoped, with the issue that owns its resolution) so `make check-graph` stays green and the debt is printed.
- Add `--inventory`: print every `// facade-compat` line and every deferred edge as a table (input for the facade modules and the M1 exit check).
- Extend the self-test with the cases listed under acceptance criteria. Do not claim completeness beyond "every reverse edge expressed through a resolvable path in production code".

## Acceptance criteria

- [ ] `--self-test` covers: alias (`as`), grouped and nested groups, file-qualified path (`crate::error::chains::ChainError`), expression-position path (`crate::error::ChainError::invalid_parameters(..)`), a name re-exported from a non-`error` module, a `super::` import inside a submodule, a glob import from `crate::error::*`, a same-named foreign `Error` bound by `use std::io::Error` (no edge), a type alias wrapping an error, and a stale `DEFERRED` entry.
- [ ] `make check-graph` on `main` reports zero forbidden edges and prints every deferred edge, including the ones surfaced by the resolver, each with its file and owning issue.
- [ ] `--inventory` lists every `facade-compat` line and every deferred edge.
- [ ] The script's docstring states what the resolver does not see.
- [ ] No production Rust code changes in this issue.

## Verification commands

```sh
python3 scripts/check_module_boundaries.py --self-test
make check-graph
python3 scripts/check_module_boundaries.py --inventory
```

## Dependencies

- #507 (merged part: the checker exists).

## Out of scope

- Resolving or removing any of the surfaced edges (D2/D3 of the proposal, separate issues).

joaquinbejar:
## Summary

Closes the blind spot #507's checker left: a reference spelled `crate::error::Name` was classified as a reference to the facade-level `error` module (`ALWAYS_ALLOWED_TARGETS`), so a lower layer naming a higher layer's error type in a signature was never an edge. Twenty further reverse edges become visible. **No production Rust changes and no public API change**; every surfaced edge is recorded as deferred with the issue that owns its resolution, so `make check-graph` still passes and the debt is printed on every run.

## Changes

- `error_types()` maps every `pub enum|struct|type` under `src/error/` to the file that defines it.
- `use_entries()` / `absolute()` / `reexport_map()` / `resolve_error_refs()` resolve a name through the file's `use` bindings and the crate's `pub use` re-export graph: bare, grouped, nested, `as` aliases, file-qualified paths (`crate::error::chains::ChainError`), `super::`/`self::` paths, globs (`use crate::error::*` plus a use of the name in the body) and paths in expression position. A name bound from outside the crate (`use std::io::Error`) is foreign and creates no edge.
- A `pub use` carrying `// facade-compat:` is still followed by the re-export map (`strip_comments(..., exempt_marked=False)`): the marker keeps a path alive for compatibility, it does not hide the type from the consumers of that path.
- `DEFERRED` grows from 11 to 31 entries, file-scoped, each naming its owner: the `PricingError`/`GreeksError` channels of `LegAble` and `Position`, `ChainError` in `model::utils` and `geometrics::construction`, `MetricsError` in curves/surfaces/geometrics, `SimulationError` in `volatility::utils`, `OhlcvError` in `utils::csv`, the unified `Error` in `utils::others`, and the `src/error/` variant payloads that hold a higher layer's error (`OptionsError::Greeks`, `CurveError::{Greeks, Graph}`, `SurfaceError::{Greeks, Graph}`, `VolatilityError::Chain`, `SimulationError::{Chain, Strategy}`).
- New `--inventory` mode prints the deferred edges and the `facade-compat` lines as Markdown tables (input for the facade modules at extraction and for the M1 exit check).
- Self-test: 16 → 28 cases, adding every resolution shape plus two negative cases (foreign `Error`, glob import whose name is unused) and a re-export of a re-export.
- The module docstring states what the resolver does **not** see: names reached through a type alias defined in another file, and macro-generated code. The claim is "every reverse edge expressed through a resolvable path in production code", not "every reverse edge".

## Technical decisions

The surfaced edges are recorded, not fixed: their resolution is a mix of re-homings with no public change (`MetricsError` to math, `calculate_optimal_price_range` to chains, `generate_ou_process` to simulation, `utils::csv` to market, the walk driver to the `synthetic` market generators) and breaking changes behind the 0.22.0 decision (the `PricingError` channel of `LegAble`, `ResultPoint`, the variant payloads). Each entry names the issue that owns it; none of them is approved or implemented here.

## Public API impact

None. No `pub` item added, removed or retyped; `public-api/optionstratlib.txt` unchanged; `cargo semver-checks --all-features` reports no required update.

## Testing

- [x] `python3 scripts/check_module_boundaries.py --self-test`: 28 cases, all pass
- [x] `make check-graph`: `OK: no forbidden module edge (31 deferred edges tolerated; facade-compat lines per layer: market=1, math=3)`

## Checklist

- [x] No production Rust code changed
- [x] No new external dependencies
- [x] Module boundaries: adds no edge, records twenty existing ones

Closes #590.


scripts/check_module_boundaries.py
```diff
@@ -215,8 +272,167 @@ def production_lines(text: str) -> list[str]:
...
+    table: dict[tuple[tuple[str, ...], str], str] = {}
+    for name, stem in types.items():
+        table[(("error",), name)] = stem
+        table[(("error", stem), name)] = stem
+    files = []
+    for path in sorted(src.rglob("*.rs")):
+        rel = path.relative_to(src).as_posix()
+        text = "\n".join(production_lines(strip_comments(path.read_text(), exempt_marked=False)))
+        module = module_path_of(rel)
+        pubs = [m.group(0) for m in re.finditer(r"\bpub(?:\([^)]*\))?\s+use\s+[^;]+;", text, re.S)]
```
joaquinbejar, line 412: 🟠 **Warning**: Include private parent bindings when resolving `super::` imports. A child module can access a private import in its parent, but this table only records `pub use`. With `use crate::error::PricingError; pub mod child;` in `model/position.rs` and `use super::PricingError; pub fn f() -> PricingError { PricingError::A }` in `model/position/child.rs`, the checker exits 0: the parent is deferred and the new child dependency is invisible. Changing only the parent import to `pub use` makes it correctly exit 1. Both fixtures compile.

Please track visible private bindings as well as public re-exports and assert the consumer file in the regression test. The current “super import inside a submodule” test passes from the parent edge alone; counting module-pair violations does not prove that the child resolved.

scripts/check_module_boundaries.py
```diff
@@ -215,8 +272,167 @@ def production_lines(text: str) -> list[str]:
...
+            continue
+        if abs_path[-1] == "*":
+            prefix = tuple(abs_path[:-1])
+            for (mod, name), stem in reexports.items():
+                if mod == prefix:
+                    glob_names[name] = stem
+            continue
+        stem = reexports.get((tuple(abs_path[:-1]), abs_path[-1]))
+        if stem is not None:
+            stems.add(stem)
```
joaquinbejar: 🟠 **Warning**: Resolve module aliases before examining expression and type paths. `use crate::error as errors; pub fn f() -> errors::PricingError { errors::PricingError::A }` in a new core file compiles, but the checker exits 0. The import names the allowed `error` module, and `QUALIFIED_RE` only visits paths starting with `crate::`, so the higher-layer type is never resolved. The equivalent direct import exits 1.

Keep module bindings in the resolver and expand paths rooted at those bindings before the ownership lookup. This also needs a regression where the aliased module itself is allowed, otherwise the ordinary module-edge check can mask the missing type resolution.

scripts/check_module_boundaries.py
```diff
@@ -215,8 +272,167 @@ def production_lines(text: str) -> list[str]:
...
+    for match in QUALIFIED_RE.finditer(text):
+        mods = tuple(seg for seg in match.group(1).split("::") if seg)
+        stem = reexports.get((mods, match.group(2)))
+        if stem is not None:
+            stems.add(stem)
+    if glob_names:
+        for ident in set(IDENT_RE.findall(text)):
```
joaquinbejar, line 562: 🟠 **Warning**: Respect local names that shadow a glob import. For `use crate::error::*; pub struct PricingError; pub fn f() -> PricingError { PricingError }`, Rust resolves `PricingError` to the local struct, but this identifier scan resolves it to `error/pricing` and rejects the core file. I reproduced exit 1 on this fixture although there is no dependency on that error type.

Exclude names supplied by local definitions and other explicit bindings before resolving glob candidates, rather than only excluding `bound_foreign`. Add a negative fixture for this shadowing case so introducing a local type does not turn a valid dependency graph red.

joaquinbejar (commented):
Reviewed 691c00e88540e531e4bf98d3cae299d0d8ed297f against main at 82c4f5d3d8c45621823d41bb41173202b24eafee for #590. Three blocking findings in name resolution are detailed inline: private parent bindings, module aliases, and glob shadowing.

Validation: the included self-tests, make check-graph and --inventory completed successfully. Five independent temporary Rust fixtures compiled with rustc --edition=2024 -D warnings; the checker missed both forbidden-edge fixtures and rejected the local-shadowing fixture, while the two direct/public controls correctly failed. No production Rust was changed, and I did not rerun the Rust workspace suite for this Python-only change. This COMMENT review is not a formal approval or merge authorization.

joaquinbejar (commented):
## Review: #591 (adversarial pass on the resolver)

I re-ran the checker and the self-test on a throwaway worktree of `origin/issue-590-boundary-error-resolution` (28/28 self-test cases pass; `OK: no forbidden module edge (31 deferred edges tolerated; facade-compat lines per layer: market=1, math=3)`; 0.30 s wall on the real tree). No cargo was run. I then drove `scan()` directly against ~20 synthetic `src/` fixtures to look for holes.

**Verdict: mergeable.** The resolver is a strict improvement: every edge it now reports is real, and I could not make it report an edge that a `--self-test` case does not already cover as legitimate. One factual correction is needed before merge (a claim about the docstring that the diff does not deliver); everything else is follow-up work that belongs in #590's successor, not here.

---

### 🟠 The docstring says none of this

The PR description and `CHANGELOG.md` both state:

> The script's docstring states what the resolver does not see (names reached through a type alias defined in another file, macro-generated code).

`git diff main...origin/issue-590-boundary-error-resolution -- scripts/check_module_boundaries.py` touches nothing above line 117. The module docstring at `scripts/check_module_boundaries.py:1-33` is byte-identical to `main`: it still describes only the `crate::<module>` scan, never mentions `error_types` / `resolve_error_refs`, never mentions `--inventory`, and states no limit. Every limitation below is therefore **undeclared**, which is what makes them worth listing.

Two further statements in that unchanged docstring are now stale:

* line 24: *"every such line is annotated `// deferred edge` in the source"*. `grep -rn 'deferred edge' src/` returns 22 hits covering 9 of the 31 entries. The 20 new entries carry no annotation (by design: no production Rust changed), and neither do the two pre-existing `error/*` entries. Either annotate them (a comment-only change, zero Rust semantics) or drop the invariant from the docstring.
* lines 13-17: the `facade-compat` bullet says marked lines are *exempt*. As of this PR they are exempt from the **edge scan** but deliberately **followed** by the re-export map (`strip_comments(..., exempt_marked=False)`). That distinction needs one sentence.

Fix: add a "What the resolver does not resolve" paragraph listing the four shapes below, document `--inventory`, and correct the two stale claims. One edit, no logic change.

### 🟠 Three shapes that produce **zero** violations

Each fixture was run through `scan()` + `violations_of()` on an otherwise empty `src/`.

**(a) Module alias: `use crate::error as err;`**

```rust
// src/error/chains.rs
pub enum ChainError { A }
// src/curves/x.rs
use crate::error as err;
fn f() -> err::ChainError { todo!() }
```
→ `0 violations`. `absolute()` yields `["error"]`, `reexports[(( ), "error")]` misses (lowercase, not a type), and `QUALIFIED_RE` only fires on `crate::…`, so `err::ChainError` is never seen. `EDGE_RE` sees `crate::error`, which is in `ALWAYS_ALLOWED_TARGETS` (`:123`). `use crate::error::{self as err};` behaves identically. Note the *sub*module alias `use crate::error::chains as ch;` **is** caught, but only incidentally, via `EDGE_RE` on `crate::error::chains`, not by the resolver.

Fix: in `resolve_error_refs`, when `abs_path` names a module rather than a type, record `alias -> abs_path` and re-scan for `alias::Name`.

**(b) A uniform-path (bare-relative) hop breaks the re-export chain**

```rust
// src/error/chains.rs        pub enum ChainError { A }
// src/curves/inner.rs        pub use crate::error::ChainError; // facade-compat: market
// src/curves/mod.rs          pub mod inner;
//                            pub use inner::ChainError;
// src/geometrics/x.rs        use crate::curves::ChainError;
//                            fn f() -> ChainError { todo!() }
```
→ `0 violations`. `absolute()` (`:349`) returns `None` for any head that is not `crate`/`self`/`super`, so Rust-2018 uniform paths are classified as foreign. Without the marker the fixture still misses the **consumer** and reports only `curves/inner.rs`.

This is not hypothetical: bare-relative `pub use` is the house style in this repo: `src/error/mod.rs:260-279`, `src/lib.rs:1401-1403`, `src/series/mod.rs:15-17`, `src/visualization/mod.rs:329-331`. The `error/mod.rs` re-exports only resolve today because `reexport_map:358-360` **seeds** `(("error",), name)` and `(("error", stem), name)` unconditionally for every type, whether or not `error/mod.rs` actually re-exports it. That seeding papers over `absolute()`'s gap for `src/error/` and nowhere else. I patched `absolute()` to resolve uniform paths and re-ran: 258 → 277 edges, all 19 new ones `error/mod -> error/*` (facade, allowed), 0 new violations. So the gap costs nothing **today**; it is one `pub use` in `series/mod.rs` away from costing something.

**(c) Two error files defining the same type name**

`error_types` (`:302`) uses `names.setdefault(name, stem)`: on a collision the **alphabetically first file wins, silently**. The current tree already has two collisions:

| name | files | resolver picks |
| --- | --- | --- |
| `StrategyErrorKind` | `error/chains.rs:453` (market), `error/position.rs` (core) | `chains` |
| `PriceErrorKind` | `error/probability.rs` (analytics), `error/strategies.rs:174` (strategies) | `probability` |

Neither is re-exported from `error/mod.rs`, so nothing is misreported right now. But the failure mode is silent *and* permissive:

```rust
// src/error/decimal.rs      pub enum WrapKind { A }     (core)
// src/error/strategies.rs   pub enum WrapKind { B }     (strategies)
// src/error/mod.rs          pub use strategies::WrapKind;
// src/model/x.rs            use crate::error::WrapKind;
//                           fn f() -> WrapKind { todo!() }
```
→ `0 violations`. `decimal` sorts before `strategies`, `decimal` is core, core→core is allowed, the strategies→core inversion disappears. Sorted-order-dependent silence is the worst property a boundary checker can have.

Fix: make `error_types` detect a collision and either `sys.exit(2)` with the two files, or drop the ambiguous name from the bare `(("error",), name)` seed while keeping the file-qualified `(("error", stem), name)` entries (which `EDGE_RE` already covers). Cheap, and it converts a silent miss into a loud one.

### 🟠 Type aliases: the edge is attributed to the definition site, never to the consumer

```rust
// src/error/chains.rs               pub enum ChainError { A }
// src/geometrics/types.rs           use crate::error::ChainError;
//                                   pub type ResultPoint<P> = Result<P, ChainError>;
// src/curves/x.rs                   use crate::geometrics::ResultPoint;
//                                   pub fn f() -> ResultPoint<u8> { todo!() }
```
→ 1 violation, `geometrics -> error/chains in geometrics/types.rs`. `curves -> error/chains` is **not** reported.

This exact shape is live and public: `src/geometrics/construction/types.rs:16` defines `pub type ResultPoint<Point> = Result<Point, ChainError>`, re-exported from `geometrics/construction/mod.rs:46`, `geometrics/mod.rs:18` and `prelude.rs:93`. A math-layer alias over a market-layer error, on the public surface. Today only `geometrics` and a `#[cfg(test)]` block in `surfaces/surface.rs:2467` use it, so nothing is hidden; but the first production `use crate::geometrics::ResultPoint;` in `curves/` will cross math→market invisibly. `("geometrics", "error/chains")` is deferred to "retype ResultPoint (#511)", which is the right disposition; the point is that the deferral is file-scoped to the *definition* and therefore does not constrain consumers at all.

The PR body admits this limit. The docstring does not (see above). No code change needed here; a docstring sentence plus a note on the `#511` entry is enough.

**Macros:** the PR body's claim that macro-generated code is unseen is materially wrong in the direction that matters: the `crate::error::X` tokens sit literally in the macro body, so `scan` reports the edge against the file that **defines** the macro. What is missed is the attribution to the invocation site:

```rust
// src/utils/m.rs   #[macro_export] macro_rules! bail_strat {
//                     () => { return Err(crate::error::StrategyError::A) } }
// src/model/x.rs   fn f() { bail_strat!() }
```
→ 1 violation, `utils -> error/strategies in utils/m.rs`; `model -> error/strategies` missed. Conservative enough, but the docstring should say *which* side is missed.

### 💡 Glob imports ignore shadowing and scan string literals

`resolve_error_refs:425-428` scans the whole file body with `IDENT_RE` for every name a glob could bind. Three confirmed false positives, all gated behind a production `use crate::error::*;`:

```rust
use crate::error::*;  struct ChainError;                                   // local definition shadows -> still 1 violation
use crate::error::*;  fn f() { tracing::warn!("ChainError while parsing"); } // string literal -> still 1 violation
use crate::error::*;  use crate::curves::o::ChainError;                    // crate-local rebind -> still 1 violation
```

The third is the interesting one: `bound_foreign` (`:407`) only suppresses names rebound from **outside** the crate, so `use std::io::Error` is handled (self-test covers it) but `use crate::model::other::Error` is not.

Latent, not live: `grep -rn 'use [a-z_:]*::\*;' src/` finds no production `use crate::error::*`; the only globs are `use super::*` inside `#[cfg(test)]` and six `pub use crate::<module>::*` in `prelude.rs`. Since the risk is confined to a shape the codebase does not use, a docstring note ("a glob binding is resolved by textual identifier match; do not glob-import `crate::error`") costs nothing and is probably the right trade. If you want it tighter, suppress any name that the file defines itself or rebinds with a later `use`.

`error_types:294` also uses a non-recursive `(src/"error").glob("*.rs")`, while `scan:440` classifies `error/<x>/mod.rs` as source `error/<x>`. Flat today; a future `src/error/chains/mod.rs` would silently drop every type it defines out of the table while still being scanned as a source. One-line fix (`rglob` + stem from the directory).

### ✅ The `DEFERRED` table checks out

I dumped `scan()`'s actual `edges` on the real tree and compared each key's file set against its declared `frozenset`. **All 31 entries match exactly**: no entry is over-broad, no declared file is unused, so no future file can slip in under an existing entry. The stale-entry guard in the self-test confirms 31 of 31 are live. Spot-checked the prose against the source and every one is accurate: `model/utils.rs:6` `ChainError`, `utils/others.rs:13` `Error`, `volatility/utils.rs:13` `SimulationError`, `simulation/walk_driver.rs:17` `ChainError`, `geometrics/analysis/traits.rs:1` `MetricsError`, `utils/csv.rs:7` `OhlcvError`, `curves/curve.rs:10` and `surfaces/surface.rs:26` `MetricsError`. `model/option.rs:354` mentions `crate::error::PricingError` in a doc comment only, and correctly produces no `model -> error/pricing` edge from that file; the deferral is scoped to `model/leg/*` + `model/position.rs`, which is exactly what the scan reports.

### 🔵 Six owner issues are closed

`#498`, `#499`, `#502`, `#504`, `#505`, `#508` are all CLOSED. For the `0.22.0 batch (…)` dispositions that is the established convention: the work is parked behind the semver gate and the deferred list is the tracker. But three of the new entries carry a *concrete, unbatched* action against a closed issue:

* `("model", "error/chains")` → "re-home to chains::utils (**#498**)": closed
* `("volatility", "error/simulation")` → "re-home to simulation (**#502**)": closed
* `("model", "error/pricing")` → "M1 exit proposal row 5 (#498)": closed, and the referenced proposal is not in the repo

Nothing will resurface these except someone reading `DEFERRED`. Suggest repointing them at an open issue (`#511` or a new M1-exit ticket), or adding an explicit `M1 exit blocker` marker so `--inventory` can separate "parked behind 0.22.0" from "needs a home".

Related: `doc/DEPENDENCY-MATRIX.md` (cited in the docstring), `ADR-0001`, `ADR-0003` and `API-BASELINE` are cited 31 times between them and **none is tracked in git** (`git ls-files doc/` → `generate_md_docs.sh`, `images/logo.png`). They live in a locally-excluded `doc/`. Anyone reading this script from a fresh clone cannot follow a single citation. Pre-existing, and possibly deliberate, but this PR triples the citation count so it is worth a decision.

### 💡 `--inventory`: the "Layer" column means the opposite of the run's "per layer"

`inventory:688-693` prints the text **after** the marker, which is the *target* layer, under the header `Layer`:

```
| chains/mod.rs:242   | analytics     |   <- chains is market, the analytics is the target
| curves/mod.rs:7     | analytics     |
| geometrics/mod.rs:34| visualization |
| surfaces/mod.rs:18  | analytics     |
```

while the normal run prints `facade-compat lines per layer: market=1, math=3` from `marked_lines()`, which counts by *source* layer. Two outputs, same word, opposite meanings. Rename to `| File | Source layer | Compat target |` and derive the source layer the way `marked_lines` does.

Also: `inventory` reads raw `path.read_text()` lines, so it does not strip `#[cfg(test)]` or comments. Harmless today (4 markers, all production) and consistent with `marked_lines`, but if the marker string ever appears in a doc comment the two counts will still agree while both being wrong.

### ✅ Fixed point, determinism, performance

`reexport_map`'s loop is monotone (`table` only grows, `changed` is set only on an insert) and bounded by `(#modules + 1) x #error-type-names`, so it terminates. On the real tree: 0.30 s total including three full passes over `src/` (`error_types`, `reexport_map`, `scan`), against 0.02 s on `main`. Fine. Iteration order is `sorted(rglob(...))` and `sorted(glob(...))` throughout and every collision resolution is first-writer-wins over a sorted sequence, so the output is deterministic; which is exactly *why* finding (c) is silent rather than flaky. `--inventory` sorts both tables. Python floor is 3.9 (`str.removesuffix`), unchanged by this PR; `str | None` annotations are safe under `from __future__ import annotations`.

### ✅ The marker asymmetry is right

Following `// facade-compat` `pub use` lines in `reexport_map` while still exempting them in the edge scan is the correct pairing, and it closes a hole rather than opening one. The marker's meaning is "this path survives extraction as a facade module file"; it licenses the *line*, not the *type*. If it also hid the type from the re-export map, then

```rust
// src/model/mod.rs   pub use crate::error::StrategyError; // facade-compat: strategies
// src/model/x.rs     use crate::model::StrategyError;
```

would launder a strategies-owned error into core behind a one-word comment. I verified the current behaviour reports `model -> error/strategies` for that fixture. The self-test's "error re-exported from another module" case pins it. Keep it.

The one residue is finding (b): a marked `pub use` *plus* a bare-relative hop is silent end to end. Fixing `absolute()` closes that too.

---

### Summary

| | |
| --- | --- |
| 🟠 | Docstring/CHANGELOG claim a limits section the diff does not add; two other docstring claims now stale; **fix before merge** |
| 🟠 | Silent false negatives: module alias `use crate::error as err`; uniform-path re-export hop; same-named types in two error files |
| 🟠 | Type-alias channel attributes the edge to the definition site only (`ResultPoint` is this shape, live and public) |
| 💡 | Glob resolution ignores shadowing and matches identifiers in string literals (latent: no production glob of `crate::error`) |
| 💡 | `error_types` is non-recursive; `--inventory`'s `Layer` column is the target layer while the run's is the source layer |
| 🔵 | Six owner issues closed, three of them on unbatched actions; the four cited design docs are untracked |
| ✅ | All 31 `DEFERRED` entries match the scan exactly; no over-broad entry |
| ✅ | Fixed point terminates, output deterministic, 0.30 s on the real tree |
| ✅ | `facade-compat` asymmetry is correct and closes a laundering hole |

None of the gaps is a regression: every one of them was equally invisible before #591, and the PR converts 20 previously-unseen edges into recorded, file-scoped debt. Ship it with the docstring correction; open a follow-up for (a)/(b)/(c) and the alias channel so the "every reverse edge expressed through a resolvable path" claim has a precise complement written down.

scripts/check_module_boundaries.py
joaquinbejar: Confirmed and fixed in 4fc3cf24. `resolution_tables()` now builds two tables: `public` (only `pub use`, reachable from anywhere) and `private` (every `use`, per module). `resolve_error_refs` falls back to the private table only when the resolved module is an ancestor of the consumer's module, which is exactly the `super::` case. Your fixture is self-test case "private parent binding seen through super" (`model/position/mod.rs` private import + `model/position/child.rs` using `super::StrategyError`), and it asserts the violation text contains `model/position/child.rs`, so it cannot pass on the parent's edge: the harness now takes an optional substring that pins which file carries the edge, and several cases use it.

scripts/check_module_boundaries.py
joaquinbejar: Confirmed and fixed in 4fc3cf24. Module bindings are tracked: `use crate::error as errors` and `use crate::error::{self as errors}` record `errors -> (error,)`, and every `<alias>::[<mods>::]Name` occurrence in the file is then resolved through the same ownership lookup as a `crate::` path. Two self-test cases cover it ("module alias", "module alias through self"), both asserting the consumer file, and both alias the *allowed* `error` module so the ordinary module-edge check cannot mask the type resolution, as you asked. The same commit also fixes the related hop you did not hit: `absolute()` now understands the uniform bare-relative form (`pub use inner::X;`), which is this crate's `mod.rs` style and previously ended a re-export chain.

scripts/check_module_boundaries.py
joaquinbejar: Confirmed and fixed in 4fc3cf24. The glob pass now skips any name that is defined in the same file (`LOCAL_DEF_RE` over `enum|struct|type|trait|union`) or bound by another explicit `use` in that file, not just names bound from outside the crate. Two negative self-test cases: "glob name shadowed by a local definition" (your fixture, expects 0) and "glob name shadowed by an explicit import" (expects 0). The docstring now also states that a `use` inside a function body that rebinds a name is not seen.

joaquinbejar:
Both reviews addressed in 4fc3cf24 (full gate list rerun, green). Summary of what changed beyond the three inline threads:

**False negatives closed** (each with a self-test case, several asserting which file carries the edge):
- module aliases (`use crate::error as err; err::ChainError`, and the `{self as err}` form);
- the uniform bare-relative re-export hop (`pub use inner::X;`), which is this crate's `mod.rs` style and used to end a re-export chain: `absolute()` now resolves a head that names a submodule of the current module;
- private parent bindings reached from a child through `super::`;
- `pub type Alias<..> = Result<_, SomeError>`: the edge now reaches every consumer of the alias, not only the file that defines it. That is live today, so `("geometrics", "error/chains")` grows from one file to three (`construction/types.rs`, `construction/mod.rs`, `mod.rs`); the `surfaces` uses of `ResultPoint` are `#[cfg(test)]`;
- ambiguous names: `error_types()` no longer takes the alphabetically-first file silently. A name defined by two error files resolves to **every** candidate (the conservative direction), and `--inventory` lists them. The two that already exist are `StrategyErrorKind` (`error/chains.rs`, `error/position.rs`) and `PriceErrorKind` (`error/probability.rs`, `error/strategies.rs`).

**False positive closed**: under a glob, a name shadowed by a local definition or by another explicit import is skipped.

**Factual correction**: the previous commit message and the CHANGELOG claimed the module docstring stated the resolver's limits. It did not. The docstring now describes every resolved shape, states what is not seen (a type reached with no binding and no `crate::` path, macro-generated code, a `use` inside a function body), corrects the stale claim that every deferred line is annotated `// deferred edge` (20 of 31 are not, by design: recording them changed no production Rust) and corrects the facade-compat bullet, since a marked `pub use` raises no edge itself but its type is still resolved for the consumers of that path.

**`--inventory`**: columns are now `File | Source layer | Compat target`, so the source-layer counts in the normal run and the table use the same word for the same thing; ambiguous error names get their own table.

**Owner strings**: the three entries carrying an unbatched action pointed at closed issues (#498, #502). They now point at #507, which is open and owns the boundary documentation, with the decision id from the local proposal.

Not changed, for the record:
- `DEFERRED` file sets were verified exact against the tree in the review; the two new entries added here (`geometrics/construction/mod.rs`, `geometrics/mod.rs`) are the alias re-exports and nothing else.
- The marker asymmetry stays as reviewed: the re-export map follows `// facade-compat` `pub use` lines while the edge scan exempts them.
- `doc/` is excluded from the index by the owner's local configuration; the citations in the script point at documents that live outside the repository. That is a standing decision, flagged to the owner rather than changed here.

scripts/check_module_boundaries.py
joaquinbejar: Verified on 4fc3cf24 with the original private-parent fixture: the checker now exits 1 and identifies model/position/child.rs, while the public-parent control still exits 1. The new test also asserts the consumer filename. This finding is corrected.

scripts/check_module_boundaries.py
joaquinbejar: Verified on 4fc3cf24: the original module-alias fixture now exits 1 with the same forbidden edge as the direct-import control. The alias and grouped-self regression tests pass. This finding is corrected.

scripts/check_module_boundaries.py
joaquinbejar: Verified on 4fc3cf24: the original local-shadowing fixture now exits 0. The fixture also imports a permitted core error through the glob, so it compiles with -D warnings without suppressing the unused-import warning. This finding is corrected.

scripts/check_module_boundaries.py
```diff
@@ -215,8 +280,248 @@ def production_lines(text: str) -> list[str]:
...
+        rel = path.relative_to(src).as_posix()
+        text = "\n".join(production_lines(strip_comments(path.read_text(), exempt_marked=False)))
+        pubs = [m.group(0) for m in re.finditer(r"\bpub(?:\([^)]*\))?\s+use\s+[^;]+;", text, re.S)]
+        alls = [m.group(0) for m in re.finditer(r"\buse\s+[^;]+;", text, re.S)]
+        files.append((module_path_of(rel), pubs, alls))
+        # `pub type Alias<..> = .. SomeError ..` makes the alias a path to the
+        # error type for every consumer of the alias.
+        for alias_name, rhs in ALIAS_DEF_RE.findall(text):
+            for ident in IDENT_RE.findall(rhs):
+                stem = types.get(ident)
```
joaquinbejar: 🟠 **Warning**: Resolve the alias RHS through its defining module's bindings before assigning an error owner. `types.get(ident)` matches only the spelling, so `use std::io::Error; pub type IoResult<T> = Result<T, Error>;` in `curves/alias.rs` is registered as an alias of this crate's unified `Error`. A permitted pricing consumer of `crate::curves::alias::IoResult` then fails with `pricing -> error/unified`, although the actual error is external. I compiled this fixture with `-D warnings`; the checker exits 0 before this delta and 1 afterward.

Please derive alias ownership from resolved RHS paths and local imports, including foreign-name shadowing, and add this negative case alongside the positive alias-consumer test. Otherwise the new alias support reintroduces the same same-named-foreign-error false positive that #590 explicitly requires avoiding.

joaquinbejar (commented):
Delta review of 4fc3cf2461b3252478b6f097e71174d98aa3b737 from 691c00e88540e531e4bf98d3cae299d0d8ed297f. All three original findings are verified corrected, with confirmation in their existing conversations. One blocking regression in the newly added type-alias resolution is detailed inline.

Validation: all 37 self-test cases plus the stale-entry check pass; make check-graph and --inventory pass. All five original checker fixtures now return the expected result. A new fixture using a std::io::Error alias compiles with rustc --edition=2024 -D warnings; it passes the previous checker and fails this head with a spurious pricing -> error/unified edge. No Rust workspace suite rerun for this Python-only delta. This COMMENT is not a formal approval or merge authorization.

scripts/check_module_boundaries.py
joaquinbejar: Confirmed and fixed in 264b6b5c. The alias pass no longer matches the right-hand side by spelling: it resolves the named type through the bindings of the file that defines the alias (`file_error_bindings`, the same public/private tables the rest of the resolver uses) and, when the type is written in full, through the `crate::error::...` path. A name bound from another crate is absent from those bindings, so your fixture (`use std::io::Error; pub type IoResult<T> = Result<T, Error>;` in `curves/alias.rs`, consumed from `pricing/`) now yields zero violations, and it is self-test case "alias over a foreign error of the same name". The positive counterpart is "alias over a qualified crate error", which asserts the consumer file. The docstring's alias bullet states the rule.

joaquinbejar: Verified on 264b6b5c: the original std::io::Error alias fixture now exits 0, while the direct qualified crate-error alias test still detects the consumer. The original three fixtures also retain their corrected behavior. This finding is corrected; I found a separate regression in propagation through a re-export and have reported it on the new lines.

scripts/check_module_boundaries.py
```diff
@@ -215,8 +280,293 @@ def production_lines(text: str) -> list[str]:
...
+                    stem = public.get((mods, match.group(2)))
+                    if stem is not None:
+                        break
+                if stem is None:
+                    for ident in IDENT_RE.findall(rhs):
+                        if ident in local:
+                            stem = local[ident]
+                            break
+                if stem is not None and (module, alias_name) not in public:
+                    public[(module, alias_name)] = stem
```
joaquinbejar, line 467: 🟠 **Warning**: Feed newly resolved aliases back into the re-export fixed point. The public/private `use` passes have already finished when this loop inserts an alias, and `added` only repeats the alias pass, so a `pub use` of that alias never becomes resolvable. For example, define `StratResult<T> = Result<T, crate::error::StrategyError>` in `strategies/mod.rs`, re-export it from `curves/mod.rs` with `pub use crate::strategies::StratResult; // facade-compat: strategies`, and consume `crate::curves::StratResult` from pricing. This valid Rust fixture exits 1 on 4fc3cf24 and 0 here: the forbidden pricing dependency disappears behind the compatibility path. An ordinary unmarked re-export also loses the consumer edge.

Please iterate alias discovery and import/re-export propagation together until neither table changes, and add a regression asserting the downstream consumer file. Checking direct imports of the alias alone does not exercise this propagation step.

joaquinbejar (commented):
Delta review of 264b6b5cc182b7c8fb3a90df43f42c3531461751 from 4fc3cf2461b3252478b6f097e71174d98aa3b737. The foreign-error alias finding is verified corrected, and the three original fixes remain effective. One blocking regression is detailed inline: aliases added after the re-export pass do not propagate to downstream consumers.

Validation: 39 self-test cases plus the stale-entry check, make check-graph and --inventory pass. The six existing probe cases return the expected result. Two new alias/re-export fixtures compile with rustc --edition=2024 -D warnings. Compared with the previous head, this head loses the pricing consumer edge in both; the compatibility re-export fixture changes from exit 1 to exit 0. No full Rust workspace suite rerun for this Python-only delta. This COMMENT is not a formal approval or merge authorization.

scripts/check_module_boundaries.py
joaquinbejar: Confirmed and fixed in ff65ea03. The  passes and the alias pass now run inside a single outer fixed point (`while progress:` around both), so an alias that becomes resolvable makes every later re-export of it resolvable too, and vice versa. Your fixture is self-test case "re-export of an alias resolves" (alias defined in `strategies/mod.rs`, re-exported from `curves/mod.rs` with the compat marker, consumed as `crate::curves::StratResult` from `pricing/x.rs`), asserting the violation is attributed to `pricing/x.rs`. It also covers the unmarked form: the marker is irrelevant to the resolver, since `reexport_map` reads marked lines too. The real tree is unchanged: 31 deferred edges, no new violation, run time still 0.3 s.

joaquinbejar: Correction to the previous reply, where a shell expansion swallowed two words: it is the `use` passes (public and private) and the alias pass that now run inside a single outer fixed point. Nothing else in that reply changed.

joaquinbejar: Verified on ff65ea03: both independent re-export fixtures now report the pricing consumer again. The compatibility-path fixture returns 1 instead of 0, and the unmarked fixture retains both source and consumer edges. All six earlier probes retain their expected outcomes. The shared fixed point addresses this finding; I am leaving conversation resolution to the author.

joaquinbejar (commented):
Delta review of ff65ea036a50695a130795298ce5716cfd6088a5 from 264b6b5cc182b7c8fb3a90df43f42c3531461751. The alias/re-export propagation finding is verified corrected. All five findings raised in this review sequence are now verified fixed; I found no additional blocking issue in this delta.

Validation: 40 self-test cases plus the stale-entry check pass; make check-graph and --inventory pass. All eight independent checker fixtures return the expected results, including both downstream re-export consumers, private parent bindings, module aliases, local shadowing and the foreign-error alias. Review was limited to the Python/documentation delta; no full Rust workspace suite was rerun. This COMMENT is not formal approval or merge authorization.

PR #593, opened 2026-09-20
Add the accepted-breaks register and the three-baseline semver gate (#592)

### LINKED ISSUE #592: M1-17 — Accepted-breaks register and three-baseline semver gate
Milestone: 0.22 Milestone 1 — Normalize dependencies inside the monolith. Local planning material: `doc/M1-EXIT-PROPOSAL.md` section 3 (not in the repository).

## Problem

Normalising the dependency graph needs a batch of incompatible changes (removed wrappers, removed error variants, retyped error channels). The manifests stay at 0.21.3 and nothing is published during the migration, so `cargo semver-checks` compares against the published 0.21.3 and refuses every removal; at the same time the gate must keep failing automatically on any break nobody authorised.

`cargo-semver-checks` has no per-item allowlist and no machine-readable output, so the policy has to live beside it.

## Required work

Land the mechanism **without approving or landing any break**: the register ships empty and the workflow runs alongside the existing `semver` job until it is judged trustworthy.

- `public-api/accepted-breaks.toml`: one entry per authorised break, with `id`, `surfaces`, `baseline` (which comparisons must report it), `findings` (`check`, `lint`, `item`), `decision`, `issue`, `migration`, `status` (`proposed` | `approved` | `landed`), `approval` (comment reference, author login) and `landed.pr`. No commit SHA is stored: the landing commit is resolved from the `Accepted-Breaks:` trailer of the squash message.
- `scripts/check_accepted_breaks.py`: fail-closed parser of the pinned `cargo-semver-checks` report plus the register comparison, with a self-test over checked-in fixtures.
- Three comparisons per surface:
  - **C1 cumulative** against the published 0.21.3;
  - **C2 controlled reference**: `--baseline-rev B`, B = newest commit carrying an `Accepted-Breaks:` trailer among the ancestors of the state *before* the change under test (ancestors of `base.sha` on a PR, of `HEAD^` on a push to `main`, excluding the commit itself), initial B = tag `v0.21.3`;
  - **C3 incremental**: baseline = the asserted `HEAD^1` of the synthetic merge commit, current = that synthetic commit.
- Six surfaces on every PR: `none`, `default` (`synthetic`), `plotly`, `static_export` (+`plotly`), `async`, `all`.
- Pin the synthetic commit: assert `HEAD^1 == base.sha` and `HEAD^2 == head.sha` from the event payload and fail on a mismatch, so a re-run cannot follow a moved ref.
- `make check-breaks`, a `.github/workflows/accepted_breaks.yml` job and the pinned tool version.
- A rule in `AGENTS.md`: an agent may add an entry with `status = "proposed"`; only the repository owner sets `approved`, through a comment on the register tracking issue.

## Acceptance criteria

- [ ] The register exists, is empty of `approved` entries, and its schema is documented in the script's docstring.
- [ ] The parser fails (never reports "zero breaks") on: a non-zero, non-100 exit; a missing `Checked`/`Summary` line; a block count that disagrees with the reported failure count; any truncation marker inside a `Failed in:` list; an unknown lint; a tool version that does not match the pinned one. Fixtures for each case are checked in and exercised by the self-test.
- [ ] A finding is keyed `(surface, check, lint, item)`; two different lints on one item are two findings.
- [ ] The four-step sequence (normal PR adds an item; authorised break removes a 0.21.3 item; normal unlabelled PR; authorised break removes the item added in step 1) is reproduced by a checked-in fixture script and yields: C1 blind to the step-4 removal, C2 and C3 reporting it, and step 3 green on C2/C3 with no label.
- [ ] The same sequence is exercised both as a PR (synthetic merge commit) and as the push that integrates it, and C2 never selects the commit under test as its own baseline.
- [ ] All six surfaces run on every pull request.
- [ ] The job's verdicts are *unforeseen*, *unapproved*, *stale* and *misdeclared*, each with a message naming the entry or the finding.
- [ ] No existing gate is weakened: the current `semver` job keeps running unchanged.

## Verification commands

```sh
python3 scripts/check_accepted_breaks.py --self-test
make check-breaks
```

## Dependencies

- None. It must land before any breaking change of the 0.22 batch.

## Out of scope

- Approving or landing any break (D3, D4, D5 of the proposal).
- Turning the new job into a required check: that is a separate decision once the implementation has been reviewed.
- The `unpublished` exemption (D6c): the gate ships without it.

joaquinbejar:
## Summary

The mechanism that will let the 0.22 batch land while the manifests stay at 0.21.3 and nothing is published. **It authorises nothing by itself**: the register ships with no approved entry, the new workflow is informational (not a required check), the existing `semver` job is unchanged, and no production Rust is touched.

Activation as a blocking check is a separate decision, after this implementation and its results have been reviewed.

## Why

`cargo semver-checks` compares against the published 0.21.3, so it refuses every removal, and it has no per-item allowlist and no machine-readable output. The policy therefore lives beside it, and it has to hold three properties that one comparison cannot:

- a break authorised once must not have to be re-authorised by every later pull request;
- a removal of an item that was **added after 0.21.3** must still be caught, although the published baseline has never seen it;
- an item can leave the default surface and stay behind a feature, which `--all-features` cannot see.

## The three comparisons

| Check | Baseline | Current | Sees |
|---|---|---|---|
| C1 cumulative | published `0.21.3` | the checked-out tree | what a consumer of the release sees |
| C2 controlled reference | newest commit carrying an `Accepted-Breaks:` trailer among the ancestors of the state **before** the change under test (ancestors of `base.sha` on a PR, of `HEAD^` on a push), initial reference `v0.21.3` | same | anything not already authorised at that reference |
| C3 incremental | the first parent of the merge commit CI checks out (the PR's base) | that merge commit | the PR's own delta, including removals invisible to C1 |

C2 deliberately excludes the commit under test: on the push that integrates a break, the "newest trailer-carrying ancestor" *is* that commit, and the job would compare it with itself.

C3's baseline is not `git merge-base`: measured in a throwaway repository, after a parent squash-merges and the child is retargeted, the merge base is the pre-branch commit and the squash commit is not an ancestor of the child, so the child would be blamed for the parent's removals. The merge commit's first parent is the base by construction, and the job pins it by asserting `HEAD^1 == base.sha` and `HEAD^2 == head.sha` from the event payload, so a re-run cannot follow a moved ref.

## Fail-closed parsing

An exit status other than 0 or 100, a missing `Checked`/`Summary` line, a block count that disagrees with the announced failure count, a truncation marker inside a `Failed in:` list, an unknown lint or a tool version other than the pinned `0.50.0` is an error. A report that cannot be trusted is never read as "zero breaks". Findings are keyed `(surface, check, lint, item)`, so two changes on one type, and one item reported under two lints, stay two findings.

## Fixtures

`tests/fixtures/semver-reports/` (see its README):

- six **real** reports from a throwaway probe crate: seven lints with 69 items, a clean run, the feature-gated removal that only `--only-explicit-features` sees, a compile error, a missing baseline, conflicting flags;
- six **derived** reports, each with one defect the parser must refuse, plus the duplicate-item case;
- `sequence.sh`, which builds a throwaway repository and walks a probe crate through four states, running all three comparisons on each **twice**: as a pull request (on the synthetic merge commit) and as the push that integrates it. Recorded output in `sequence-results.txt`; re-running reproduces it with different SHAs.

Sequence result, which is the evidence for the three properties above:

| Step | Change | C1 | C2 | C3 |
|---|---|---|---|---|
| 1 | normal PR adds `added_fn` | - | - | - |
| 2 | authorised break removes `old_fn` (in 0.21.3) | reports it | reports it | reports it |
| 3 | normal, unlabelled PR changes a body | still reports the landed break | green | green |
| 4 | authorised break removes `added_fn` | **blind** | reports it | reports it |

## Authorisation

An agent may add an entry with `status = "proposed"`; a proposed entry never makes the gate pass. Only the repository owner sets `approved`, through a comment on #592 naming the entry ids, which the script verifies (author login, reference, id). The landing PR carries the `accepted-break` label and an `Accepted-Breaks: AB-xx` trailer; it fills `landed.pr` and may not add or approve entries. No SHA is stored in the register: the landing commit is resolved from the trailer. This is a procedural control, not a technical one, since one account serves both roles today; `CODEOWNERS` with a second identity is the alternative if you want enforcement.

## Public API impact

None. No Rust changed; `public-api/optionstratlib.txt` unchanged; `cargo semver-checks --all-features` reports no required update.

## Testing

- [x] `make check-breaks` (parser self-test, 17 cases over the fixtures)
- [x] `sequence.sh` in a fresh work directory reproduces `sequence-results.txt` identically modulo SHAs
- [x] Full local gate list green: `cargo test --all-features --workspace`, `LOGLEVEL=WARN cargo test`, `cargo build --release`, `RUSTDOCFLAGS=-D warnings cargo doc --all-features --no-deps`, `make scan-banned`, `make public-api-check`, `make check-graph`, `cargo semver-checks --all-features`

## Checklist

- [x] No production Rust changed
- [x] No new external dependency (`tomllib` is stdlib on Python 3.11+; the Makefile picks the newest available interpreter)
- [x] The existing `semver` job keeps running unchanged
- [x] The register contains no approved entry

Closes #592.


scripts/check_accepted_breaks.py
```diff
@@ -0,0 +1,317 @@
...
+    pending: set[tuple[str, str]] = set()
+    for entry in register.get("break", []):
+        if surface not in entry.get("surfaces", []):
+            continue
+        for finding in entry.get("findings", []):
+            if finding.get("check") != check:
+                continue
+            key = (finding["lint"], finding["item"])
+            approved = entry.get("status") in ("approved", "landed")
+            if approved and (entry.get("status") == "landed" or not landed_only):
```
joaquinbejar: 🟠 **Warning**: Scope expected findings to the selected baseline and landing history. Every `landed` entry remains in `ready` for C2 and C3 forever. After AB-01 lands, an ordinary subsequent PR has no C2/C3 findings for that removal, but `verdicts()` reports the retained AB-01 entries as stale and fails both checks. I reproduced this using a landed entry declaring C1/C2/C3: C1 matches, while clean C2 and C3 each fail as stale. This breaks step 3 of the required sequence.

Resolve the entry's landing commit and determine whether the break lies after each comparison baseline; retain it cumulatively for C1 but do not require old removals from later C2/C3 deltas. Extend the sequence test to execute the register comparison itself: sequence.sh currently runs cargo-semver-checks and extracts text with awk, so its green step 3 does not test this policy.

scripts/check_accepted_breaks.py
```diff
@@ -0,0 +1,317 @@
...
+    ready: set[tuple[str, str]] = set()
+    pending: set[tuple[str, str]] = set()
+    for entry in register.get("break", []):
+        if surface not in entry.get("surfaces", []):
+            continue
+        for finding in entry.get("findings", []):
+            if finding.get("check") != check:
+                continue
+            key = (finding["lint"], finding["item"])
+            approved = entry.get("status") in ("approved", "landed")
```
joaquinbejar: 🟠 **Warning**: Validate authorisation before accepting an entry as ready. At present `status = "approved"` or `"landed"` is the entire authorisation check: neither the approval comment, its author and entry ids, nor the landing PR label/trailer is inspected anywhere in the execution path. Using a recorded semver failure and an entry marked approved with no approval metadata and `landed.pr = 0`, the CLI returns 0 and prints “authorised findings”. A PR can therefore add its own approved entry and make its break pass.

Please implement the approval and landing checks described in this PR before populating `ready`, including rejection when the landing PR adds or approves its own entries. Sharing one GitHub identity makes human approval procedural, as documented, but it does not justify skipping the promised reference, author, id and landing validation altogether.

scripts/check_accepted_breaks.py
```diff
@@ -0,0 +1,317 @@
...
+    register = load_register()
+    surfaces = {surface["id"]: surface for surface in register["register"]["surfaces"]}
+    if args.surface not in surfaces or args.check is None:
+        parser.error("--check and --surface are required unless --self-test is given")
+    try:
+        if args.base_sha and args.head_sha:
+            assert_synthetic(args.base_sha, args.head_sha)
+        published = register["register"]["published"] if args.check == "C1" else None
+        text, code = run_semver(args.baseline or "", surfaces[args.surface], published)
+        found = parse_report(text, code, register["register"]["tool"], known_lints(text))
```
joaquinbejar: 🟠 **Warning**: Obtain known lint names independently of the report being validated. `known_lints(text)` returns exactly the lint names in that report, so the membership check inside `parse_report()` can never reject an unknown lint in production. Passing the checked-in unknown-lint.txt through this exact call accepts all 69 findings, including brand_new_lint; the self-test only rejects it because it supplies names from a different fixture.

Use a checked-in lint inventory for the pinned tool, or obtain and validate its supported lint list independently. Exercise the same production call in the malformed-report tests so an unknown lint produces a parser error even when a register entry happens to match it.

joaquinbejar (commented):
Reviewed 5b630ecf2f3c2c0243736536c9b9c6713bbcbcce against main at 82c4f5d3d8c45621823d41bb41173202b24eafee for #592. Three blocking findings are detailed inline: landed entries remain stale in subsequent incremental comparisons, approval/landing metadata is not validated, and unknown lint validation is circular.

Validation: make check-breaks passes. Focused policy experiments using the checked-in reports reproduce all three findings, including the production parser invocation and a CLI invocation with the semver subprocess mocked to return the real recorded report. The sequence fixture was inspected but not executed; it creates commits and exercises raw cargo-semver-checks rather than the register verdicts. No Rust production changes are present, and no workspace build/test suite was rerun. This COMMENT is not formal approval or merge authorization.

joaquinbejar (commented):
Reviewed adversarially: read the script and workflow line by line, ran the real `--self-test` (needs Python ≥3.11; my system `python3` is 3.9, used `python3.13` directly), hand-crafted malformed reports and fed them straight to `parse_report`, built a throwaway probe crate and ran real `cargo-semver-checks 0.50.0` against it to check the report grammar and the `--only-explicit-features`/feature-implication assumptions, and ran `tests/fixtures/semver-reports/sequence.sh` end to end.

## 1. Parser soundness

Fed these directly to `parse_report()` (scripts/check_accepted_breaks.py:100-139):

| Case | Result |
| --- | --- |
| Item line wrapped onto a continuation line (`  function ...\n    in file ...:33`) | 🔴 **mis-parses silently.** The loop at lines 126-134 treats every line starting with two spaces as its own item. The continuation line, once run through `LOCATION_RE`, reduces to `""`, so you get `{('function_missing', 'function probe::...'), ('function_missing', '')}` — a phantom finding with an empty item, no `ReportError`. I confirmed with a real crate that `cargo-semver-checks 0.50.0` never wraps item lines even with `COLUMNS=40` on a captured (non-tty) stream, so this isn't reachable with the pinned tool today, but the parser has no structural check that would catch it if a future point release ever did wrap. |
| Item text containing the substring `" in file "` (e.g. `probe::helper_in_file_thing`) | Handled correctly — `LOCATION_RE` is anchored at the end (`$`), so only the real trailing location is stripped. |
| `Failed in:` list with a blank line between two items | 🟠 **silently drops the second item, no error.** Line 128's `if not line.startswith("  ") or not line.strip(): if items: break` treats the first blank line as the end of the list. I verified against a real multi-item report (69 items across different files/modules in `breaks.txt`, and a fresh run of the real tool against a crate with items in different files) that `cargo-semver-checks 0.50.0` never inserts a blank line inside one `Failed in:` list — items are always consecutive — so this is not exploitable with the pinned version either, but it's the same class of problem as the previous row: on a genuine grammar drift, this degrades to "found fewer breaks than really happened" instead of raising `ReportError`, which is exactly what the fail-closed design is supposed to prevent. |
| Windows line endings (CRLF) | Parses fine (`.strip()` eats the `\r`). |
| `N warn` present alongside `N fail` | Parses fine — `CHECKED_RE` doesn't require anchoring past the optional fail group. |
| `--- failure` substring inside a description paragraph (not at start of line) | Handled correctly, `FAILURE_RE` is `^`-anchored with `re.M`. Note descriptions come from the tool's fixed lint text, not attacker input, so this isn't really an adversarial surface. |
| Item with no source-location suffix at all | Parses fine, item kept verbatim (matches the real `enum E in <root>/current/src/lib.rs:5` shape already in `clean.txt`/`breaks.txt`). |

Net: for the exact pinned tool version (0.50.0), I could not get the parser to produce a *silently wrong but passing* result on real report shapes — every genuine 0.50.0 output I could produce (including a 69-item, multi-file, multi-lint report) round-trips correctly, and every malformed report the fixtures cover is rejected. The two rows marked 🔴/🟠 are latent: they only fire if a future point release of `cargo-semver-checks` changes wrapping or blank-line grouping, which the version pin (`IMPL_VERSION_RE` check, line 107-109) is supposed to guard against — but note that guard only fires if the report happens to contain an `impl:` link (i.e., at least one failure block); a hypothetical future version that changed formatting *and* still uses `0.50.0`-style single-line items wouldn't be caught by the version check at all if the version string itself didn't change. 💡 Suggestion: make the per-item loop assert a location suffix is present (or track line-start indices) rather than accepting a location-free residual line as a distinct item, so a wrapped continuation can't inject an empty-item finding.

## 2. The three comparisons

Read `check_accepted_breaks.py` together with `accepted_breaks.yml`.

- **C2 excludes the commit under test in both cases**, and I traced the "before" value correctly: PR case uses `base.sha` (line 66), push case uses `HEAD^` (line 74) — never the commit under test. Ran the full `sequence.sh` scenario (see §5) and it holds.
- 🟠 **Dead code / duplicated logic.** `reference_commit()` (scripts/check_accepted_breaks.py:146-159) is never called from anywhere in the script — grepped, zero call sites. The actual C2 reference resolution that runs in CI is a *second, independent* implementation inline in the workflow's "Resolve baselines" step (`.github/workflows/accepted_breaks.yml:76-80`):
  ```
  ref=$(git log --format=%H --grep='^Accepted-Breaks:' "$before" | head -1)
  [ -n "$ref" ] || ref=$(git rev-parse v0.21.3)
  ```
  This hardcodes the tag `v0.21.3` literally, instead of reading `register["register"]["initial_reference"]` from `public-api/accepted-breaks.toml` the way the (unused) Python function does. If `initial_reference` is ever bumped in the TOML (e.g. once 0.22.0 ships and the migration needs a new floor), the bash step will keep pointing at `v0.21.3` silently — only the dead Python function would have picked up the change. Either delete `reference_commit()` or make the workflow call it (`python3 -c` one-liner, or a small `--resolve-reference` script mode) so there's one source of truth.
- **First push after this PR merges, no trailer anywhere**: falls back to `v0.21.3` cleanly (line 80 in the workflow / `initial_reference` in the register) — verified in `sequence.sh` step 1's "as push to main" row (`C2 reference=<v0.21.3 sha>`).
- **`v0.21.3` not fetched (shallow clone)**: not reachable here — the workflow always uses `fetch-depth: 0` (`accepted_breaks.yml:39-42`), which fetches full history, and `actions/checkout` auto-follows tags reachable from fetched commits even without an explicit `fetch-tags: true`. Still, there's no explicit guard: if the tag were ever missing, `git rev-parse v0.21.3` fails, the bash step's `set -eu` (line 64) kills the job — fails closed, just with a confusing raw `git` error rather than a clear message. 💡 nitpick only.
- **PR targeting `release/**`**: the workflow triggers on it (`accepted_breaks.yml:19-21`) and the base-sha resolution works the same way (base sha = release branch tip). C1's baseline stays "published 0.21.3" regardless of target branch, which is consistent with the current single-version-in-flight migration model, but would need revisiting if this repo ever maintains multiple release branches with breaks accepted at different baselines. Not a bug for the stated scope.
- **Merge commit with more than two parents / no second parent**: `assert_synthetic()` (line 162-175) does `git rev-parse HEAD^2`; if a PR's synthetic merge ref doesn't exist (e.g. genuine merge conflict), `actions/checkout` itself fails before the script runs, so this is moot. If for some other reason `HEAD^2` doesn't resolve, `git()`'s `subprocess.run(..., check=True)` (line 143) raises an uncaught `CalledProcessError` — not wrapped as `ReportError`, so you get a raw traceback instead of the clean "re-run the job on a fresh merge" message. Still fails closed (nonzero exit), just ugly. 🔵 nitpick.
- **Fork PRs**: `github.event.pull_request.base.sha`/`head.sha` are available and correct for fork PRs too (no `secrets` used, no elevated permissions requested), so this should work unchanged. `GITHUB_TOKEN` default read-only permissions are fine since the job only reads.

## 3. Surfaces vs. `Cargo.toml`

Compared `public-api/accepted-breaks.toml`'s `[[register.surfaces]]` against the real `[features]` block (Cargo.toml:28-45: `default = ["synthetic"]`, `plotly = ["dep:plotly"]`, `static_export = ["plotly", "async"]`, `async = [...]`).

- `default` surface correctly spells out `features = ["synthetic"]` explicitly, matching today's default — good, this is exactly the kind of drift the comment (accepted-breaks.toml:23) says it's guarding against.
- `static_export` surface lists `features = ["static_export", "plotly"]`. I verified empirically (throwaway crate with `static_export = ["plotly", "async"]`) that requesting `--only-explicit-features --features static_export` alone already pulls in both `plotly`- and `async`-gated items via Cargo's normal feature-dependency resolution — `--only-explicit-features` only suppresses the "enable everything except unstable/nightly/..." heuristic, it does not disable a feature's own declared sub-features. So listing `plotly` explicitly here is redundant, not necessary, but harmless (it doesn't add or remove anything from the resolved feature set). 🔵 nitpick: could drop it, but no correctness issue either way.
- `--only-explicit-features` + `--features X` is confirmed (via `cargo semver-checks check-release --help` on the pinned 0.50.0 binary) to be the correct way to express "exactly X and its declared dependencies, nothing else."
- The `none` surface (`features = [], default_features = false`) round-trips through `surface_flags()` (line 178-184) to `--only-explicit-features` with no `--features` flags at all — correctly expresses a bare, feature-free build.
- 🟡 `default_features = false` is set on **every** surface entry in the TOML but is **never read** by `surface_flags()` — grepped, no reference to that key anywhere in the Python. It's dead config. Harmless today (since `--only-explicit-features` already achieves the same effect regardless), but it's misleading: a reader of the TOML would reasonably assume that key does something. 💡 Suggestion: drop the key or wire it in for documentation-as-code parity.

## 4. Register semantics

- `expected()`/`verdicts()` (lines 194-220): only three verdicts are actually implemented — **unforeseen** (215-216), **unapproved** (217-218), **stale** (219-220).
- 🟠 **"misdeclared" is documented but not implemented.** The module docstring (lines 35-37) promises a fourth verdict: *"misdeclared: an entry declares the finding for a surface or a check that did not produce it."* I grepped the whole script: the string `"misdeclared"` occurs exactly once, in that docstring comment — nowhere in `expected()`, `verdicts()`, or `self_test()`. Since each `--check`/`--surface` invocation is a separate process with no shared state, there's no code path that could ever compare "declared for check C" against "actually produced by check D" — a genuine cross-check misattribution would currently surface only indirectly, as an `unforeseen` on the check that really produced it plus a `stale` on the check it was wrongly declared under, never as a single clearly-labelled "misdeclared" message. `self_test()` has no case for this either. Either implement it or strike it from the docstring so the four-verdict claim matches the three-verdict reality.
- `status = "proposed"` can never make any check pass: `expected()` line 205 only sets `approved = True` for `status in ("approved", "landed")`; a `proposed` entry always lands in `pending`, and any real finding matching a `pending` key is reported as `unapproved` (line 217-218), which fails the run. Confirmed coherent.
- `--landed-only` is coherent but has a sharp edge worth calling out explicitly: on a PR (landed_only unset), a finding with `status = "approved"` is `ready`; on the *next push to main* (`landed_only` set, line 71 vs. 75/81 in the workflow), that same `status = "approved"` entry becomes `pending` again (line 206: `entry.get("status") == "landed" or not landed_only` is false), so its finding is reported as `unapproved`. That means the PR that lands a break **must** also flip the entry's `status` to `"landed"` (and presumably fill in `landed.pr`) in the very same commit, or the very first push-triggered run on `main` after merge fails. This is workable but easy to get wrong for a contributor who only sets `status = "approved"` and forgets the follow-up edit — worth a line in the register's own comment block or a PR template checklist item. 💡 Suggestion, not a bug (the job is informational for now anyway, so it can't block a merge, but it will show a permanently red workflow run).
- `owner = "joaquinbejar"` (accepted-breaks.toml:12) and `approval = { ref = "", date = "" }` (documented in the commented example at the bottom of the same file) are **not read anywhere in the code** — grepped both the script and the workflow, zero hits for `"owner"` or `"approval"`. There is also no `CODEOWNERS` file in the repo. So "only the repository owner may mark an entry approved" is currently enforced by nothing but prose and manual PR review; the script would accept `status = "approved"` from any commit regardless of who wrote the TOML change. 🟠 Worth flagging plainly since the PR description states this as a rule: the rule exists in the docstring and the register comment, not in code or branch protection.

## 5. Sequence fixture

Ran `bash tests/fixtures/semver-reports/sequence.sh /tmp/seq-run` fresh (own tmp dir, real `cargo semver-checks 0.50.0`, real git). It reproduces `sequence-results.txt` modulo SHAs exactly, item for item, exit code for exit code. Confirmed the recorded conclusions hold in my run:
- **Step 3 (normal PR, no trailer) is green on C2 and C3** (`rc=0`, `<none>`) while **C1 stays red** (`rc=100: function_missing function probe::old_fn`) — because C1 always compares against the fixed 0.21.3 baseline and doesn't know the break already landed.
- **Step 4 (break removes `added_fn`, which was only added in step 1, never in 0.21.3) is invisible to C1** — C1's output at step 4 is unchanged from step 3/2 (still only reports `old_fn`), confirming "C1 blind at step 4" — while C2 and C3 correctly report the new `function_missing function probe::added_fn`.
- The script's own `[wrong C2, inclusive]` line demonstrates concretely, on every "as push to main" state, what the wrong (self-inclusive) reference would have picked (the commit under test itself) versus the correct exclusive one — this is a nice, load-bearing regression check for the exact bug class C2 is designed to avoid.
- One portability note (🔵 nitpick, doesn't affect correctness of the result, only where the script can be run): `sequence.sh` uses BSD/macOS `sed -i ''` syntax (lines defining `rm_old`/`touch_body`/`rm_added`). This is not invoked by CI or by `make check-breaks` — it's a standalone reviewer/dev tool — but it will not run as-is on a Linux (GNU sed) box; GNU sed would treat the empty string as the sed script and the real script as the filename, and error out (loudly, not silently) rather than corrupt anything. Worth a one-line note in the README or a portable `perl -i -pe` swap if Linux contributors are expected to run it directly.

## 6. CI mechanics

- **Python version floor**: the workflow calls bare `python3` in every step (`accepted_breaks.yml:60`, and the three check steps) with no version assertion or `actions/setup-python`. `ubuntu-latest`'s current image ships Python ≥3.11 by default, so this works today, but it's inconsistent with the Makefile's own defensiveness (`Makefile:81`: `PYTHON311 ?= $(shell command -v python3.13 || command -v python3.12 || command -v python3.11 || echo python3)`) — the workflow trusts the runner image where the Makefile explicitly doesn't trust the developer's machine. If it ever fails, it fails closed (the script's own `ReportError` on `tomllib is None`, my local `python3` is 3.9.6 and reproduces exactly that), just not proactively. 💡 Suggestion: pin with `actions/setup-python@v5, python-version: '3.11'` for parity with the Makefile's intent.
- **fetch-depth**: `0` (full history) is required and present — correct, no complaint.
- **Tool install pinning**: `cargo-semver-checks@0.50.0` pinned via `SEMVER_CHECKS_VERSION` env var and matched against the register's `tool = "cargo-semver-checks 0.50.0"` — consistent, single source of truth for the version string (unlike the `v0.21.3` duplication in §2).
- 🟠 **Job cost, no caching.** Six-way surface matrix × three checks (C1/C2/C3), and each `cargo semver-checks` invocation does two full builds (baseline + current) — up to 36 full `cargo build`-equivalent compilations per push/PR, and the workflow has no `actions/cache` step for `~/.cargo/registry`, `~/.cargo/git`, or `target` (compare `build.yml`/`lint.yml` conventions in this repo). With `fetch-depth: 0` on top, this is a heavy, uncached job for something explicitly marked "informational" (`accepted_breaks.yml:16-18`). Recommend caching keyed on `Cargo.lock` before this becomes a required check.
- **`concurrency`**: present and correct (`group: ${{ github.workflow }}-${{ github.ref }}`, `cancel-in-progress: true`).
- **Can it pass by accident if `cargo-semver-checks` is missing?** No — verified directly: an unknown cargo subcommand exits `101` (`cargo totally-bogus-subcommand`), and `parse_report()` line 102 rejects any exit code outside `(0, 100)` before it ever looks for "zero findings." Confirmed fail-closed.

## Verdict

**Mergeable**, with the understanding stated in the PR itself that this workflow is informational, not a required check, until reviewed — which is exactly the review this is. Nothing I found is a security hole or a way to make an *unauthorised* break silently pass on the pinned tool version and the real report shapes I could produce; the parser held up under a real 69-item, 7-lint, multi-file report and under every fixture-covered malformation. The items worth addressing before this becomes a required, blocking gate, roughly in priority order:

1. 🟠 Delete the dead `reference_commit()` (scripts/check_accepted_breaks.py:146-159) or make the workflow call it, so the C2 fallback-tag logic has one source of truth instead of two (Python reads `initial_reference` from the register; bash hardcodes `v0.21.3`).
2. 🟠 Either implement the "misdeclared" verdict or remove the promise from the docstring (scripts/check_accepted_breaks.py:35-37) — right now it's aspirational.
3. 🟠 State plainly (in the register's own comment or a PR template) that `owner`/`approval` are enforced by human review only, not by any code path — worth being explicit about this since the PR description phrases it as "a rule."
4. 🟠 Add `actions/cache` before this graduates from informational to required — 36 uncached full builds per run is expensive for a matrix job that will presumably run on every PR forever.
5. 💡/🔵 Everything else above (empty-item on a hypothetical wrapped line, blank-line-inside-list drop, dead `default_features` key, redundant `plotly` in the `static_export` surface, `sed -i ''` portability in `sequence.sh`, bare `python3` vs. the Makefile's fallback chain) — none reachable with the pinned tool version and today's report shapes, but worth a follow-up pass since the whole point of this gate is defense in depth against future drift.

scripts/check_accepted_breaks.py
joaquinbejar: Confirmed and fixed in b5af4f84. `known_lints()` no longer derives anything from the report: it reads `tests/fixtures/semver-reports/lints-0.50.0.txt`, the 254 lints of the pinned version taken from `cargo semver-checks --list`, and every production call site uses it. The docstring says why it is not derived from the report. `unknown-lint.txt` now fails through the same call the CLI makes, not a list supplied from another fixture.

scripts/check_accepted_breaks.py
joaquinbejar: Confirmed and fixed in b5af4f84. Authorisation is now checked before an entry can populate `ready`: `authorised()` requires the status, an approval reference pointing at a comment on the register issue, a decision and a migration note. `--against <base>` compares the register with the base revision and refuses a change that moves one of its own entries to `approved` or that rewrites its approval reference in the same change that uses it; the workflow passes it on every pull request together with `--pr`, and an entry that is landing here must name that number in `landed.pr`. `--verify-approvals` reads each approval comment through `gh` and checks the author is the register owner and that the entry id appears in the body; CI runs it as its own step. What is still procedural, and now said plainly in the docstring rather than only in the PR body: one GitHub identity serves both roles, so none of this stops the owner account being used by someone else; `CODEOWNERS` with a second identity is the technical alternative.

scripts/check_accepted_breaks.py
joaquinbejar: Confirmed and fixed in b5af4f84; this was the most serious of the three. An entry is now expected by C1 always (the published baseline contains no landing) and by C2/C3 only while its landing commit is outside that comparison's baseline: `landing_commit()` resolves it from the `Accepted-Breaks:` trailer and `in_delta()` asks `git merge-base --is-ancestor`. An authorised break already inside the baseline goes to a new `tolerated` bucket: neither demanded nor reported as unforeseen if it does show up. And the sequence fixture now runs the register comparison itself, as you asked, not only `cargo semver-checks` plus awk: it writes a register into the probe repository (AB-01 at step 2, AB-01+AB-02 at step 4) and calls the script for C1, C2 and C3 in each state, in both the pull-request and push forms. 24 verdicts, all passing, step 3 included. Doing that exposed a parser defect no fixture had: cargo's progress lines are interleaved with the report and were being counted as items, so an item is now recognised by its exact two-space indentation, a progress line ends the list and anything else is an error.

joaquinbejar:
Review addressed in b5af4f84 (full gate list rerun, green). Beyond the three inline threads:

**Cache and job cost** — `Swatinem/rust-cache` keyed per surface, plus `actions/setup-python@v5` pinned to 3.12 so the job does not depend on the runner's default interpreter (the register needs `tomllib`).

**Single source of truth for C2** — `reference_commit()` is no longer dead code: the workflow passes `--before` and the script resolves the reference itself, reading the register's `initial_reference` instead of the hardcoded tag the workflow used to carry.

**Register cleanup** — `default_features` removed (never read by `surface_flags()`); `static_export` no longer lists the `plotly` it already implies, with a comment saying implied features are Cargo's job; a `repo` key added for `--verify-approvals`.

**Portability** — `sequence.sh` no longer uses BSD `sed -i ''`; it edits through a temporary file and runs on Linux.

**New fixtures** (18 files now): `lints-0.50.0.txt` (the pinned tool's inventory), `blank-line-in-list.txt` (a blank line inside a `Failed in:` list must not truncate it: all 69 items still parse) and `wrapped-item.txt` (a split item line is refused instead of producing a phantom empty item). Your two smaller observations were both real: the first would have dropped items silently, the second would have injected a phantom finding.

**What the sequence fixture proves now** — it writes a register into the probe repository and runs the register comparison, not only the tool:

| Step | C1 | C2 | C3 | register verdicts |
|---|---|---|---|---|
| 1 add `added_fn` | - | - | - | all OK (nothing expected) |
| 2 break removes `old_fn` | reports | reports | reports | all OK (AB-01 expected by all three) |
| 3 ordinary PR | reports AB-01 | - | - | all OK (AB-01 expected by C1, tolerated by C2/C3) |
| 4 break removes `added_fn` | blind | reports | reports | all OK (AB-02 expected by C2/C3 only) |

24 verdicts, zero failures, each state run twice (as a pull request on the synthetic merge commit, and as the push that integrates it).

scripts/check_accepted_breaks.py
joaquinbejar: Rechecked at `b5af4f84`: the original stale failure is fixed for a landing that is already an ancestor of the baseline. There is still a scope problem: with an approved entry whose `landed.pr` is 999, no matching landing commit, and the current PR 123, `in_delta` returns false and `expected` puts its finding in `tolerated`. A matching removal in PR 123 consequently passes both C2 and C3. I reproduced this with the policy functions; both verdict lists were empty.

Please distinguish an entry already present in the baseline from one that has not landed and does not belong to this PR. Only the former belongs in the tolerated set; the latter must not authorise a removal in this delta.

scripts/check_accepted_breaks.py
joaquinbejar: Rechecked at `b5af4f84`: missing approval references are now rejected. The new `self_approval` check still permits changing the approved scope: starting from an approved entry for `function probe::old_fn`, I replaced its findings with a different function while retaining its id, status and approval reference. `self_approval` returned no problems. The comment lookup also continues to accept the same id, so an existing approval can authorise a newly substituted removal.

Please compare the approved entry's substantive content with the base, allowing only the intended landing metadata and status transition. Changes to findings, surfaces or other approved scope need a fresh approval rather than inheriting the old comment.

scripts/check_accepted_breaks.py
joaquinbejar: Verified at `b5af4f84`: the production parser now reads the independent lint inventory and rejects `unknown-lint.txt` with `unknown lint 'brand_new_lint'`. `make check-breaks` also passes. This finding is fixed.

joaquinbejar (commented):
Reviewed the delta at b5af4f84402fce6eb16ede849dbc1b05b2cdbb03. Two blocking policy issues remain, detailed in the existing inline threads: unlanded entries can be tolerated in another PR, and approved finding scopes can be rewritten without a new approval. The original historical stale-entry failure and unknown-lint acceptance are fixed.

Validation: make check-breaks passed; focused Python policy probes reproduced the remaining issues and checked both fixes. These probes exercise policy functions with synthetic register data, not a full cargo comparison. I did not run the sequence fixture because it creates commits, which this review session is not authorised to do. No production code was changed. This COMMENT review is not a formal approval.

joaquinbejar (commented):
Reviewed the rebase at 3d7fdd049717118731ac5082cccd63be7b7872eb against main at 904b3f57. The policy script is byte-identical to b5af4f84, and the two existing blocking findings still reproduce: an approved but unlanded entry for another PR is tolerated in C2/C3, and an existing approved entry can substitute its findings without a new approval. The existing inline threads remain the actionable locations; no duplicate comments added.

make check-breaks passed. Focused Python policy probes on this head reproduced both issues, with a synthetic base-register response for the self-approval check. No full cargo comparison or workspace suite was run. This COMMENT is not formal approval.

joaquinbejar (commented):
Reviewed the delta at `8f27bc64012b79de0d4d52dbd6bcdfb16289d9bf` against `3d7fdd04`. One new blocking finding concerns the proposed AB-01 comparison scope. The two existing policy findings remain open: this delta does not change their implementation.

Validation: `make check-breaks` passed. Focused policy probes show that the proposed entry still produces misdeclared findings after simulated approval for C1/static_export and C2/none. The C2 reference resolver returns the v0.21.3 commit for the current base. I did not rerun the six-surface cargo-semver-checks matrix or wait for CI. This COMMENT is not approval of AB-01 or merge authorisation.

public-api/accepted-breaks.toml
```diff
@@ -0,0 +1,108 @@
...
+# the entries stay `proposed`, the accepted-breaks job reports them, and its
+# result does not block: see the `continue-on-error` note in
+# `.github/workflows/accepted_breaks.yml`.
+#
+# The alternative to approving is to make the generators available without
+# the feature again, which undoes part of #512.
+
+[[break]]
+id = "AB-01"
+surfaces = ["none", "plotly", "async"]
```
joaquinbejar, line 94: 🟠 **Warning**: Record the complete comparison scope before requesting approval for AB-01. `static_export` also omits `synthetic`: this surface uses `--only-explicit-features`, and `static_export` only implies `plotly` and `async`. The five generator removals therefore apply there too. C2 currently resolves to `v0.21.3` (there is no accepted-break reference commit in the base history), so the same removals also occur in C2, although all five entries here declare only C1.

Approving this entry as written will still leave those comparisons failing as `misdeclared break`. I reproduced that policy result for C1/static_export and C2/none with an approved copy of this entry in memory. Please include the affected surface and C2 findings for the initial reference, then check the entry against those reports as well as C1/none.

joaquinbejar (commented):
Reviewed the ANSI parsing delta at `1aca99383dc788f510c5d18a4b4a8057d64c3090` against `8f27bc64`. No new blocking finding in this delta. The three existing findings remain open; their policy and register code is unchanged.

`make check-breaks` passed. An additional focused probe preserves all 69 findings when the breaking-report fixture is ANSI-coloured, and the coloured unknown-lint fixture still fails closed. No cargo-semver-checks matrix rerun and no CI wait. This COMMENT is not merge approval.

joaquinbejar (commented):
Reviewed the workflow and parser delta at `b50806fc88ca0ca3602724b19e0b6ef08e7ba3b1` against `1aca9938`. No additional blocking finding in this delta. The three existing findings remain open and were reproduced against this revision.

`make check-breaks` passed, including rejection of a successful exit containing failure blocks. Focused policy probes still accept findings from an approved but unlanded entry outside its landing PR, permit a finding-scope rewrite with unchanged approval metadata, and classify the proposed AB-01 coverage as misdeclared for C1/static_export and C2/none. The scope-rewrite probe supplies an approved previous register through a mocked Git read; no commits or production edits were made. No full semver matrix rerun or CI wait. This COMMENT is not merge approval.

PR #603, opened 2026-09-20
Set the manifest version to 0.22.0 for the migration (#602)

### LINKED ISSUE #602: M8-00 — Set the manifest version to 0.22.0 for the migration
Milestone: 0.22 Milestone 8 — Package and release 0.22. Local planning material: `doc/M1-EXIT-PROPOSAL.md` (not in the repository).

## Problem

The migration needs incompatible changes, and `cargo semver-checks` compares against the published 0.21.3. While `Cargo.toml` says `0.21.3`, the tool reads "no version change, assume minor" and refuses every removal, so the batch cannot land and a break that already landed is reported as unauthorised.

The first run of the accepted-breaks gate (#592) showed one such case: the `synthetic` feature (#512, merged as #578) removed `chains::generator_positive`, `chains::generator_optionchain`, `series::generator_optionseries` and their prelude paths from every surface that does not enable it. For a 0.x crate the breaking bump is the minor digit, so declaring `0.22.0` is what authorises this and the rest of the batch.

## Required work

- `Cargo.toml`: `version = "0.22.0"`.
- The four version strings in the `src/lib.rs` documentation header, and `README.md` regenerated from it (`make readme`).
- Nothing else: **no tag, no GitHub release and no `cargo publish`**. Those happen when every issue of the migration is closed, in the rest of Milestone 8 (#557 to #563).
- Re-check the accepted-breaks gate afterwards: with the bump declared, C1 no longer reports the `synthetic` removals, so the register ships with no entry and its live role is C2 (controlled reference) and C3 (incremental), which keep comparing 0.22.0 against 0.22.0 and still catch an unannounced break.

## Acceptance criteria

- [ ] `Cargo.toml`, `src/lib.rs` and `README.md` all say 0.22.0 and agree with each other.
- [ ] `cargo semver-checks --all-features` passes without a required update.
- [ ] The accepted-breaks job is green on all six surfaces with an empty register.
- [ ] No tag, no release, nothing published.
- [ ] The `CHANGELOG.md` `[Unreleased]` section records the `synthetic` incompatibility as the first entry of the 0.22 migration notes.

## Verification commands

```sh
cargo semver-checks --all-features
make check-breaks
grep -rn '0\.22\.0' Cargo.toml src/lib.rs README.md | head
```

## Dependencies

- None. It unblocks the breaking batch of M1 and the removals scheduled for M6-04 (#545) and #511.

## Out of scope

- Tagging, releasing or publishing (#557 to #563).

joaquinbejar:
## Summary

Declares 0.22.0 in the manifest so the migration's incompatible changes can land. **No tag, no release, nothing published**: that is the rest of Milestone 8, once every issue of the migration is closed.

## Why now

`cargo semver-checks` compares against the published 0.21.3. With the manifest also at 0.21.3 it reads "no version change, assume minor" and refuses every removal, which blocks the breaking batch of M1 and reports an already-landed break as unauthorised. For a 0.x crate the breaking bump is the minor digit.

The first run of the accepted-breaks gate (#593) made the second point concrete: the `synthetic` feature (#512, landed as #578) removed `chains::generator_positive`, `chains::generator_optionchain`, `series::generator_optionseries` and their prelude paths from every surface that does not enable it. Declaring 0.22.0 is what authorises that.

## Changes

- `Cargo.toml`: `version = "0.22.0"`.
- The four version strings in the `src/lib.rs` doc header, and `README.md` regenerated with `make readme`.
- `CHANGELOG.md`: a `### Migration notes for 0.22` section, with the bump itself and the `synthetic` requirement written for consumers.

## Effect on the gates

- `cargo semver-checks --all-features`: `0 checks: 0 pass, 254 skip`, `no semver update required`.
- Accepted-breaks (#593, still open): C1 compares against the published 0.21.3 and, with the bump declared, stops reporting those removals, so the register ships with no entry. Its live role becomes C2 (controlled reference) and C3 (incremental), which compare 0.22.0 against 0.22.0 and still catch a break introduced during the migration. That PR will be restacked on this one and its temporary `continue-on-error` removed.

## Testing

- [x] Full local gate list green, including `make lint`, `make fmt-check`, `make public-api-check` and `cargo semver-checks --all-features`
- [x] `Cargo.toml`, `src/lib.rs` and `README.md` agree on 0.22.0

## Checklist

- [x] No tag, no release, no publish
- [x] No source change beyond the version strings and the changelog
- [x] One issue per PR

Closes #602.


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "optionstratlib"
-version = "0.21.3"
+version = "0.22.0"
```
joaquinbejar, line 3: 🟠 **Warning**: Keep the break-detection comparisons active before advancing this version. The proposed replacement in #593 does not yet compare 0.22.0 against 0.22.0 for C2: its controlled reference falls back to `v0.21.3` until an `Accepted-Breaks:` commit exists in the base history. Its `run_semver` also has no release-type override. Consequently this bump makes both C1 and the initial C2 treat every breaking change as covered by the version increase, not just the five generator removals; the first C3 comparison against the current 0.21.3 base has the same gap. The reported `0 checks ... 254 skip` is evidence that the compatibility checks were skipped, not that the breaks were individually authorised.

Please retain the migration version freeze until the replacement policy is ready, or land an explicitly approved policy change that keeps break detection active independently of the manifest version, for example by forcing the comparison release type. Merely restacking #593 and emptying the register leaves C1 blind and C2 blind until its reference advances, contrary to the per-break authorisation requirement in #592.

joaquinbejar (commented):
Reviewed `779baf0f362993e849250f49c43b23d14df2e02b` against main `904b3f57`, including #602 and the compatibility-policy requirements in #592. One blocking finding concerns disabling break detection before the replacement comparisons are ready. This version advance also conflicts with the review mandate to retain 0.21.3 throughout implementation; #602 requests a different policy, so that conflict needs an explicit owner decision.

Validation: inspected the complete diff, confirmed all four documentation version strings agree with the manifest in both source and README, and ran `git diff --check`. Inspected the current #593 reference resolver and semver invocation: C2 still resolves to v0.21.3 and no release-type override is passed. I did not rerun the full suite or semver matrix and did not wait for CI. No publishing automation was added in this diff. This COMMENT is not merge approval.

Cargo.toml
joaquinbejar: lgtm

PR #600, opened 2026-09-20
Move the cross-component dispatchers to facade-owned files (#598)

joaquinbejar:
## Summary

Third re-homing of the D2 set (roadmap M1). Two public items compose components that must not depend on each other; they move to a new `optionstratlib::facade` module instead of living inside one of the components. No public path changes.

**Stacked on #597** (`issue-595-simulate-to-backtesting`), which is itself stacked on #596; each will be retargeted as its parent merges.

## Why

- `PricingEngine::MonteCarlo { simulator }` stores a `Simulator`, so pricing named the concrete simulation engine. The component-level form, `GenericPricingEngine<M: MonteCarloPricer>` with `NoMonteCarlo` (#508), already sits beside it.
- `process_n_times_iter` returns the crate-level unified `Error`, so the core layer named a facade type. ADR-0001 D2 sends the helper to strategies with a `StrategyError` return, which changes its signature; hosting it in the facade keeps the 0.21 signature until that breaking change is made.

## Changes

- New `src/facade/` (`mod.rs`, `pricing_engine.rs`, `combinations.rs`), layer `facade` in the boundary checker, documented as the code that becomes the facade crate at extraction.
- `PricingEngine`, `price_option`, `Priceable`, `PricingEngine::as_generic` and the `From` impl move there, with the test module that compares the two dispatchers.
- `process_n_times_iter` and its tests move there.
- Re-exports keep every 0.21 path: `pricing::{PricingEngine, price_option, Priceable}`, `pricing::unified::{PricingEngine, price_option, Priceable}` (a public module path the examples import) and `utils::others::process_n_times_iter`, each marked `// facade-compat: facade`.
- `greeks::numerical` now prices with `price_option_with(&option, &ClosedFormEngine::ClosedFormBS)` instead of `Priceable::price(&PricingEngine::ClosedFormBS)`: the same Black-Scholes arm, without a facade type in the pricing layer.
- `scripts/check_module_boundaries.py`: `facade` added to `LAYER_OF`, the `("pricing", "simulation")` and `("utils", "error/unified")` entries removed, two self-test cases added, and a stale case that used the now-gone pricing/simulation pair replaced by a live one. Deferred edges: 28 → 26.

## Public API impact

None. Every 0.21 path resolves, including the `pricing::unified::*` ones; `cargo semver-checks --all-features` reports no required update. The snapshot diff is path rewrites only.

## Testing

- [x] `make check-graph`: `OK: no forbidden module edge (26 deferred edges tolerated; ...)`
[...]

PR #608, opened 2026-09-20
Retire the 0.21.3 compatibility gate and report API changes instead (#606)

### LINKED ISSUE #606: M1-22: Replace historical compatibility enforcement with 0.22 API review reports
## 0.22.0 implementation policy

- Develop the new 0.22.0 architecture and API. Compatibility with 0.21.3 or earlier releases is not an acceptance requirement. Necessary API removals, renames, signature and serialization changes do not require individual historical-compatibility approvals; explain their purpose and update consumers and documentation.
- Keep one defining owner for each type and capability. Remove wrappers, aliases and adapters maintained only for old paths. Retain explicit facade/prelude exports where they improve the new API without duplicating types or introducing reverse dependencies.
- Preserve useful capabilities and numerical correctness. A source-breaking change does not justify silently dropping behavior. Explain any capability change and provide an executable replacement workflow where applicable.
- Review API changes against the exact effective base of each PR across the six relevant feature surfaces: none, default, plotly, static_export, async and all. Do not use old-release compatibility gates, AB registers or per-break approvals. Tool/build/parser failures remain failures, even when API differences are informational.
- Use 0.22.0 for the new workspace packages. No tag, release or crate publication during implementation. Publication issues execute only in the final release phase, after implementation and release preparation are complete and the owner initiates release. Preserve existing history, tags and releases.

## Mandatory engineering acceptance

- [ ] Monetary values use `Decimal`/validated domain newtypes. Use checked Decimal and arithmetic counter operations; never saturating/wrapping financial arithmetic. Document rounding. Restrict `f64` to bounded numeric internals with finite-value validation and controlled boundary conversion.
- [ ] Public boundaries use domain types and validated constructors. Errors are concrete `thiserror` types owned by the failing layer; no `anyhow`, production `unwrap`/`expect`, unchecked indexing, or submitted placeholder implementations.
- [ ] Use structured `tracing`; no print/debug logging or library-installed global subscriber. Gate async/I/O explicitly, keep defaults synchronous, and never block async execution without `spawn_blocking`. No unsafe concurrency or secret logging.
- [ ] Apply `#[must_use]` to pure functions and builders, `#[inline]` to small hot helpers and `#[cold]` to error paths; follow derive order `Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default` where meaningful. Use one concern per file, named constants, minimal intentional preludes, validated private fields and appropriate serde contracts.
- [ ] Document every public item, units and fallible API errors; examples compile and use `?`. Cover successful and documented error paths. Preserve meaningful regression coverage; never weaken tolerances, ignore tests or suppress warnings to pass checks. Numerical changes need analytic/reference evidence, parity/Greek invariants where applicable and deterministic RNG tests.
- [ ] Substantiate runtime performance claims with Criterion benchmarks; record reproducible dependency/build measurements for architecture claims. Prefer monomorphized hot paths, buffer reuse and measured optimization. New dependencies require explicit approval; this policy does not authorize arbitrary additions.
- [ ] Run the required pre-submission gates: `make pre-push`, or explicit formatting, Clippy with `-D warnings`, workspace tests including separate plotly and static_export/plotly surfaces, and a warning-free release build. Also run applicable docs, API-report, architecture, consumer and packaging checks. Do not relax these requirements. Verify merged content on main before closing an implementation issue.

## Decision

The owner has changed the compatibility policy: **0.22.0 is under development and compatibility with the published 0.21.3 is no longer a requirement.** The machinery built in #592 to preserve it is therefore retired:

- the **C1 cumulative comparison** against the published 0.21.3;

- the **initial reference** to the `v0.21.3` tag;

- the **AB approval process**, that is the register of individually authorised breaks, the `Accepted-Breaks:` trailer, the `accepted-break` label and `--verify-approvals`.

**AB-01** (the `synthetic` feature becoming required for `chains::generator_positive`, `chains::generator_optionchain`, `series::generator_optionseries` and their prelude paths) no longer needs a decision: it is superseded by this policy and documented as a migration note instead.

What is **kept**, because none of it exists to protect 0.21.3:

- the **six-surface matrix** (`none`, `default`, `plotly`, `static_export`, `async`, `all`): a change can leave the default surface and stay behind a feature, which `--all-features` alone cannot see;

- the **API diff against each pull request's effective base**, reported for review rather than authorised item by item;

- the **fail-closed report parser** and its fixtures, so an unreadable report is never read as "no change";

- every other control: numerical correctness, behaviour, architecture (`make check-graph`), features, tests, packaging, the public-api snapshot and the existing `semver` job's usefulness as a report.

History, tags and past releases are untouched.

## Required work

- Delete `public-api/accepted-breaks.toml` and the approval code paths (`--verify-approvals`, `self_approval`, `validate_register`, the `landed`/`approval` schema, the trailer lookup).

- Reduce the workflow to one comparison per surface: the pull request's effective base (the first parent of the merge commit CI checks out, asserted against `base.sha`/`head.sha` as today) versus the checked-out tree, with `--release-type patch` so the verdict never depends on the manifest version.

- Print the findings as the job summary and **do not fail on them**; keep failing on an unparseable or impossible report.

- Rename the script and the job to say what they now do (`scripts/report_api_changes.py`, job `api-changes`), keeping the fixtures and the sequence test that still apply.

- Record the policy change and the superseded AB-01 in `CHANGELOG.md` and in `doc/` where the gate is described.

## Acceptance criteria

- [ ] No register file, no approval flow, no reference to the `v0.21.3` tag in the gate.

- [ ] The job reports API changes per surface against the pull request's base and does not block on them.

- [ ] A malformed, truncated or unknown-lint report still fails the job.

- [ ] The six surfaces still run on every pull request.

- [ ] `make check-breaks` (renamed) still runs the parser self-test over the checked-in fixtures.

- [ ] `CHANGELOG.md` states the policy and that AB-01 is superseded.

## Dependencies

- #592 / #593 (merged): the machinery exists.

- #602 (merged): 0.22.0 is declared.

joaquinbejar:
## Summary

Implements the owner's policy change: 0.22.0 is under development and compatibility with the published 0.21.3 is not a requirement. Everything that existed to preserve it is removed; everything that exists for other reasons stays.

## Removed

- the cumulative comparison against the published 0.21.3 and the `v0.21.3` initial reference;
- `public-api/accepted-breaks.toml` and the whole approval flow: register schema, `Accepted-Breaks:` trailer lookup, `--verify-approvals`, the self-approval refusal and the per-entry verdicts;
- the sequence fixture that exercised that policy.

## Kept, because none of it protected 0.21.3

- the **six surfaces** (`none`, `default`, `plotly`, `static_export`, `async`, `all`): an item can leave the default surface and stay behind a feature, which a single `--all-features` run cannot see;
- `--release-type patch`, so the report never depends on the manifest version;
- the **fail-closed parser** and its fixtures: a non-0/100 exit, a missing summary, a block count that disagrees with it, a truncation marker, an unknown lint, an item line with no item or an unpinned tool version is still an error, because an unreadable report must never be read as "this pull request changes no API";
- the synthetic-merge pinning (`HEAD^1`/`HEAD^2` asserted against the event's `base.sha`/`head.sha`), so a re-run cannot follow a moved ref.

## What it does now

`scripts/report_api_changes.py --surface <s> --baseline <rev>` prints the items that differ, and the `API CHANGES` workflow runs it once per surface against the pull request's own base. Its output is review material; it does not ask for authorisation. `make check-api-report` runs the parser self-test, as `make check-breaks` did.

AB-01, the `synthetic` requirement for `chains::generator_positive`, `chains::generator_optionchain`, `series::generator_optionseries` and their prelude paths, is superseded by the policy and stays documented as a 0.22 migration note. History, tags and past releases are untouched.

## Testing

- [x] `python3 scripts/report_api_changes.py --self-test`: 21 cases over the checked-in fixtures
- [x] `make check-api-report`
- [x] Full local gate list green, including `make lint`, `make fmt-check`, `make check-graph`, `make public-api-check` and `cargo semver-checks --all-features`

Closes #606.


scripts/report_api_changes.py
```diff
@@ -0,0 +1,295 @@
...
+    if surface.get("all_features"):
+        return ["--all-features"]
+    flags = ["--only-explicit-features"]
+    for feature in surface.get("features", []):
+        flags += ["--features", feature]
+    return flags
+
+
+def run_semver(baseline: str, surface: dict, published: str | None, cwd: Path = ROOT) -> tuple[str, int]:
+    command = ["cargo", "semver-checks", *surface_flags(surface)]
```
joaquinbejar: 🟠 **Warning**: Actually pass `--release-type patch` to the tool. The script documentation and #606 require this, but the command built here omits it, so the manifest version still decides which checks run. With cargo-semver-checks 0.50.0, a temporary 0.21.3 baseline containing `pub fn removed()` and a 0.22.0 current crate without it produced `0 checks: 0 pass, 254 skip`; this script then printed `unchanged` and returned 0. Adding the promised override reported `function_missing` for that same change. Include the override in the executed command and cover the invocation with a regression test.

scripts/report_api_changes.py
```diff
@@ -0,0 +1,295 @@
...
+        text, code = run_semver(args.baseline, SURFACES[args.surface], None, cwd=args.root)
+        found = parse_report(text, code, TOOL, known_lints())
+    except ReportError as error:
+        # A report we cannot read is a failure: it must never be mistaken for
+        # "this pull request changes no API".
+        print(f"api-changes: {error}", file=sys.stderr)
+        return 2
+    if not found:
+        print(f"api-changes: {args.surface} unchanged against {args.baseline[:12]}")
+        return 0
```
joaquinbejar, line 322: 🟠 **Warning**: Do not equate an empty semver-lint report with an unchanged public API. Even after adding `--release-type patch`, cargo-semver-checks does not report ordinary compatible additions. I added `pub fn newly_added()` to an otherwise identical temporary crate: all 223 enabled checks passed, and feeding the real report through this script printed `api-changes: none unchanged`. The promised report of API additions, removals and reshaping therefore misses new capabilities that reviewers need to inspect.

Generate an actual public-API diff against the effective base for each surface, with semver findings as additional context, or explicitly narrow the output and contract to semver findings without claiming the API is unchanged. For #606's API-diff deliverable, the former also needs an additive-change regression case.

.github/workflows/api_changes.yml
```diff
@@ -0,0 +1,63 @@
...
+#
+# Six surfaces, because an item can leave the default surface and stay behind
+# a feature, which a single `--all-features` run cannot see.
+name: API CHANGES
+
+on:
+  pull_request:
+    branches:
+      - main
+      - 'release/**'
```
joaquinbejar: 🟠 **Warning**: Run the report for PRs targeting stack branches too. This `branches` filter matches the PR base, so a child targeting a branch such as `issue-595-simulate-to-backtesting` receives none of the six reports. That is the topology used by #600 in this migration, and #606 requires the six-surface report on every pull request against its own effective base. Remove this base-branch restriction or include the migration stack branches; the existing parent-SHA assertions can still pin the comparison.

joaquinbejar (commented):
Reviewed `561a9ecb83f76e853b753324bb2f5474b06a187b` against main `0f88bdc5` and the policy change in #606. Three blocking findings concern the replacement report itself: the missing release-type override, compatible API additions incorrectly described as unchanged, and stack branches excluded from the workflow. These findings do not ask to restore historical compatibility enforcement or AB approvals.

Validation: `make check-api-report` and `git diff --check` pass. Isolated temporary crates using the pinned cargo-semver-checks 0.50.0 reproduce both false unchanged cases; forcing patch reveals the removed function but still does not report an ordinary new public function. The captured reports were passed through this revision's reporting path. No production changes or Git commits were created for the probes. I did not run the full six-surface crate comparison or wait for CI. This COMMENT is not formal approval or merge authorisation.

scripts/report_api_changes.py
joaquinbejar: Confirmed and fixed in f16ad457. The rewrite dropped the override, so the command is now built by `semver_command`, which always appends `--release-type patch`, and a self-test case asserts exactly that on the built argument list. Your probe is the reason it matters: against a 0.21.3 baseline a 0.22.0 tree skipped all 254 lints and the script happily printed nothing.

scripts/report_api_changes.py
joaquinbejar: You are right, and rather than promise a diff `cargo-semver-checks` cannot produce, the contract is narrowed and stated where it can be read: the module docstring, the workflow header and the output line all say this reports removals and incompatible reshaping, and that an empty report means "nothing incompatible", never "the public API is unchanged". Additions reach the reviewer through the checked-in snapshot, which `make public-api-check` refuses until the author regenerates it, so a new public item shows up as a diff in the pull request itself.

.github/workflows/api_changes.yml
joaquinbejar: Confirmed and fixed in f16ad457. The base-branch filter is gone, so the workflow runs on every pull request whatever it targets, which is what a stacked pull request needs; the parent-SHA assertions still pin the comparison to the event base and head.

PR #612, opened 2026-09-20
Stop carrying a higher layer's error in a lower layer's enum (#511)

### LINKED ISSUE #511: M1-14: Partition errors by target crate

joaquinbejar:
## Summary

Eleven error variants existed only to wrap the error of a layer above, which is the reverse dependency the target architecture forbids. They are removed with their `From` impls, and the layer that actually composes both maps the failure into its own error, keeping the reason a caller reads.

Stacked on #611 (`stack-599-rehome-helpers`); rebased onto `main` before merging.

## Removed

`OptionsError::Greeks`, `CurveError::{Greeks, Graph}`, `SurfaceError::{Greeks, Graph}`, `VolatilityError::Chain`, `SimulationError::{Strategy, Chain, GraphError}`, `StrategyError::Simulation`, and the conversions that fed them.

## Where the layers meet

| Site | Before | Now |
| --- | --- | --- |
| `analytics::projections::{curves, surfaces}` | `?` into `CurveError::Greeks` | `CurveError::MetricsError` / `SurfaceError::AnalysisError`, naming the Greek that failed |
| `chains::options::deltas` | `?` into `OptionsError::Greeks` | `OptionsError::greeks_error`, naming the leg |
| `backtesting::strategy_simulation` | `?` into `SimulationError::Strategy` | `SimulationError::InvalidParameters` with the fee reason |
| `chains::chain::atm_iv` | wrapped a `ChainError` | `VolatilityError::NumericalFailure` with the reason |
| `simulation::traits` (volatility) | routed through `SimulationError::Chain` | `SimulationError::InvalidParameters` |
| `simulation::generator_positive` | returned the market `ChainError` | returns `SimulationError` |

In every case the message the caller sees still contains the original reason; what disappears is the type coupling.

## Tests

The tests that asserted the removed conversions now assert the boundary: that the reason survives the mapping (`test_curve_error_does_not_absorb_a_greeks_error`, `test_surface_error_does_not_absorb_a_greeks_error`, `test_a_greeks_failure_is_reported_as_a_numerical_failure`), and that `GraphError` keeps wrapping the math errors rather than the reverse (`test_graph_error_wraps_the_curve_error`, `test_graph_error_keeps_its_own_direction`). No assertion was weakened and no test was removed without a replacement that covers the same path.

## Examples

Example binaries whose `main` mixed a math error with a rendering error returned `Result<(), SurfaceError>` and relied on `From<GraphError>`. They return `Box<dyn std::error::Error>`, which is what mixing two layers' errors in one entry point means.

## Verification

- [x] `make check-graph`: **9 deferred edges** (from 20 on the base), no forbidden edge; the ten error edges are gone
- [x] Full local gate list green: workspace tests, default-feature tests, `make pre-push`, `make lint`, `make fmt-check`, release build, `cargo doc --all-features -D warnings`, `make scan-banned`, `make public-api-check`, `cargo semver-checks --all-features`
- [x] `rules/global_rules.md`: typed errors per layer with `#[cold]`/`#[inline(never)]` on the new mapping helpers, no `unwrap`/`expect`/indexing added, no new dependency, no numerical change

Closes #511.


src/error/simulation.rs
```diff
@@ -189,10 +176,12 @@ impl SimulationError {
...
     fn from(err: crate::error::VolatilityError) -> Self {
-        // Route through the existing lossless Chain wrapper: ChainError
-        // already models volatility failures and SimulationError is not
-        // extensible without a breaking change.
-        SimulationError::Chain(Box::new(ChainError::from(err)))
+        // Volatility is a pricing capability simulation depends on, so this
+        // direction is downward and the reason is carried as text: the chain
+        // wrapper it used to route through belonged to another layer.
+        SimulationError::InvalidParameters {
+            reason: format!("volatility: {err}"),
```
joaquinbejar: 🟠 **Warning**: Preserve typed causes when removing the reverse error edges. This blanket conversion turns every `VolatilityError`, including `DecimalError` and `NonFinite`, into `SimulationError::InvalidParameters` with only a string. For example, `walk_volatility` propagates `constant_volatility` failures through this conversion, so a numerical failure loses its structured payload and becomes indistinguishable from invalid parameters without parsing text. The new Greek adapters in `analytics/projections/curves.rs:100` and `surfaces.rs:32` similarly discard the `GreeksError` variant into math-owned string errors. #511 explicitly excludes flattening domain errors into strings; preserving the displayed reason does not satisfy that contract.

Simulation already depends on pricing, so it can retain a typed `VolatilityError` directly without restoring the market dependency. For the projections, return an analytics-owned error that wraps the pricing and math causes rather than making math own Greek failures. Add tests through the actual conversion/projection paths that verify the retained variant and payload; manually constructing a formatted error and checking its message cannot detect this loss.

joaquinbejar (commented):
Reviewed 9d4d3bf82983e29ea554fd8ef8fc7763e8f8d6d2 against logical parent 2e02e65bca0fa56bd01a3d62511ec383e4132440, excluding the helper relocation from #611. One blocking finding for #511: removing reverse dependencies by flattening typed failures into strings loses error classification. This is a 0.22.0 error-contract concern, not a request to preserve historical API paths.

Validation: inspected the incremental diff and affected callers; make check-graph passed, including its self-tests, with 9 deferred edges and no forbidden edges; git diff --check passed. No Rust builds or runtime tests were executed in this independent review. Published as COMMENT, not formal approval or merge authorization.

src/error/simulation.rs
joaquinbejar: You were right, the string conversion was losing the payload, so I redid all of it to keep the cause typed. Simulation now has `SimulationError::Volatility(Box<VolatilityError>)` and the `From` impl keeps the variant intact, so a `NonFinite { context, value }` coming out of `constant_volatility` through `walk_volatility` arrives with its fields, and `tests_typed_causes` asserts exactly that through the real path.

For the projections I added an analytics-owned `ProjectionError` that wraps the pricing and math causes instead of making math own Greek failures, with `Greek { greek, source: Box<GreeksError> }`, `Curve`, `Surface`, `Chain` and `Options` variants. `tests_typed_projection_errors` drives `get_curve_strike_versus` with a zero-strike option and matches on `ProjectionError::Greek { greek: "delta", source: GreeksError::InputError(_) }`, so a regression back to a string would fail the test rather than pass it.

While redoing it I found a defect of my own that your comment surfaced. My first pass mapped every construction failure into a single axis error, which swallowed "no valid points generated" and "failed to retrieve option data". Those are now `NoPoints { kind }` and `MissingOptionData`, and the tests assert the variant fields, not the message. Chain keeps `ChainError` since market may own `GreeksError`, and backtesting got its own `BacktestError` for the same reason.

